PR #1, opened 2021-07-29
Add Cocoapods support

jkmassel:
Adds CocoaPods/RubyGems high-level functions.

Can be tested as part of https://github.com/wordpress-mobile/WordPress-iOS/pull/16935

### DIFF
--- a/bin/cache_cocoapods
+++ b/bin/cache_cocoapods
@@ -1,0 +1,8 @@
+#!/bin/bash
+
+# Follow the typical pod installation process
+install_cocoapods
+
+# Then overwrite the cache for future jobs to use
+save_cache ~/.cocoapods "$BUILDKITE_PIPELINE_SLUG-specs-repos" --force
+save_cache ~/Library/Caches/CocoaPods/ "$BUILDKITE_PIPELINE_SLUG-global-pod-cache" --force
--- a/bin/cache_cocoapods
+++ b/bin/cache_cocoapods
@@ -2,9 +2,9 @@
 
 set -e
 
-# Follow the typical pod installation process
+# Follow the typical pod installation process...
 install_cocoapods
 
-# Then overwrite the cache for future jobs to use
+# ...then overwrite the cache for future jobs to use
 save_cache ~/.cocoapods "$BUILDKITE_PIPELINE_SLUG-specs-repos" --force
 save_cache ~/Library/Caches/CocoaPods/ "$BUILDKITE_PIPELINE_SLUG-global-pod-cache" --force
--- a/bin/cache_cocoapods
+++ b/bin/cache_cocoapods
@@ -1,5 +1,7 @@
 #!/bin/bash
 
+set -e
+
 # Follow the typical pod installation process
 install_cocoapods
 
--- a/bin/cache_repo
+++ b/bin/cache_repo
@@ -12,7 +12,8 @@ mkdir -p "/tmp/$DATE"
 git clone --mirror $BUILDKITE_REPO "/tmp/$DATE/$BUILDKITE_PIPELINE_SLUG.git"
 
 # Create the tarball
-tar -C "/tmp/$DATE" -cvf $DATE.git.tar $BUILDKITE_PIPELINE_SLUG.git
+TAR_NAME=$DATE.git.tar
+tar -C "/tmp/$DATE" -cvf $TAR_NAME $BUILDKITE_PIPELINE_SLUG.git
 
 # Copy the file to S3
-aws s3 cp $DATE.git.tar s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$DATE.git.tar
+aws s3 cp $TAR_NAME s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$TAR_NAME
--- a/bin/cache_repo
+++ b/bin/cache_repo
@@ -6,7 +6,7 @@ AWS_BUCKET=$1
 
 DATE=$(date +"%Y-%m-%d")
 
-mkdir -p /tmp/$DATE
+mkdir -p "/tmp/$DATE"
 
 # Download the repo
 git clone --mirror $BUILDKITE_REPO "/tmp/$DATE/$BUILDKITE_PIPELINE_SLUG.git"
--- a/bin/cache_repo
+++ b/bin/cache_repo
@@ -1,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+AWS_BUCKET=$1
+
+DATE=$(date +"%Y-%m-%d")
+
+mkdir -p /tmp/$DATE
+
+# Download the repo
+git clone --mirror $BUILDKITE_REPO "/tmp/$DATE/$BUILDKITE_PIPELINE_SLUG.git"
+
+# Create the tarball
+tar -C "/tmp/$DATE" -cvf $DATE.git.tar $BUILDKITE_PIPELINE_SLUG.git
+
+# Copy the file to S3
+aws s3 cp $DATE.git.tar s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$DATE.git.tar
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -14,7 +14,11 @@ restore_cache "$(hash_file Podfile.lock)"
 # If the `pod check` plugin is installed, use it to determine whether or not to install Pods at all
 # If it's not installed (or if it fails), we'll try to install Pods. 
 # If that fails, it may be due to an out-of-date repo. We can use `--repo-update` to try to resolve this automatically.
-bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
+if [ $(bundle exec pod plugins installed | grep check) ]; then
+	bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
+else
+	bundle exec pod install || bundle exec pod install --repo-update --verbose
+fi
 
 # If this is the first time we've seen this particular hash of `Podfile.lock`, create a cache entry for future use
 save_cache Pods "$(hash_file Podfile.lock)"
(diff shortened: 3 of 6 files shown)


bin/cache_cocoapods
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
+
+set -e
+
+# Follow the typical pod installation process
+install_cocoapods
+
+# Then overwrite the cache for future jobs to use
```
mokagio: Nitpick: what do you think of this?

```suggestion
# Follow the typical pod installation process...
install_cocoapods

# ...then overwrite the cache for future jobs to use
```

Just an idea. 100% irrelevant to the code's behavior, so 🤷‍♂️

bin/cache_repo
```diff
@@ -0,0 +1,18 @@
...
+mkdir -p /tmp/$DATE
+
+# Download the repo
+git clone --mirror $BUILDKITE_REPO "/tmp/$DATE/$BUILDKITE_PIPELINE_SLUG.git"
+
+# Create the tarball
+tar -C "/tmp/$DATE" -cvf $DATE.git.tar $BUILDKITE_PIPELINE_SLUG.git
+
+# Copy the file to S3
+aws s3 cp $DATE.git.tar s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$DATE.git.tar
```
mokagio: Nitpick

```suggestion
# Create the tarball
TAR_NAME=$DATE.git.tar
tar -C "/tmp/$DATE" -cvf $TAR_NAME $BUILDKITE_PIPELINE_SLUG.git

# Copy the file to S3
aws s3 cp $TAR_NAME s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$TAR_NAME
```

mokagio: I _think_ this would work without the file name for the S3 destination, too:

```
aws s3 cp $TAR_NAME $S3_BUCKET_PATH
```

I'm saying this because I upload the assets for [mokacoding](https://mokacoding.com) with a command like

```
for f in $(ls ~/Desktop/2021-06-07*.jpg); do aws s3 cp $f s3://mokacoding/ --acl public-read; done
```

But, it doesn't hurt to leave it. 🤷‍♂️ 👍

bin/install_cocoapods
```diff
@@ -0,0 +1,24 @@
...
+restore_cache "$(hash_file Podfile.lock)"
+
+# If the `pod check` plugin is installed, use it to determine whether or not to install Pods at all
+# If it's not installed (or if it fails), we'll try to install Pods. 
+# If that fails, it may be due to an out-of-date repo. We can use `--repo-update` to try to resolve this automatically.
+if [ $(bundle exec pod plugins installed | grep check) ]; then
+	bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
+else
+	bundle exec pod install || bundle exec pod install --repo-update --verbose
+fi
```
mokagio, line 21: Code golf. I wonder if something like would work

```
bundle exec pod plugins installed | grep check \
	&& bundle exec pod check
	|| bundle exec pod install \
    || bundle exec  pod install --repo-update --verbose
```

As I said, code golf 🤷‍♂️ It might be DRY but not necessarily cleaner

bin/save_cache
```diff
@@ -37,6 +37,12 @@ if [ -z "$CACHE_BUCKET_NAME" ]; then
 	fi
 fi
 
+# Use with caution – in general it's not a good idea to overwrite a cache entry
+if [ $3 == '--force' ]; then
+	echo "Deleting the existing cache key"
+	aws s3 rm "s3://a8c-ci-cache/$CACHE_KEY"
+fi
```
mokagio: I'm guessing this is here to help devs with debugging or to remove compromised caches?

bin/install_cocoapods
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
+
+set -e
+
+# Start by restoring specs repo and pod cache:
+# - The specs repo cache holds all of the Podspec files. This avoids having download them all from the CDN.
+# - The pod cache holds the downloaded Pod source files. This avoids having to check them out again.
+restore_cache "$BUILDKITE_PIPELINE_SLUG-specs-repos"
+restore_cache "$BUILDKITE_PIPELINE_SLUG-global-pod-cache"
```
mokagio, line 9: I wonder if extracting these cache keys definitions in a shared file to `source` here and in the `cache_cocoapods` script would be useful or make the code harder to follow 🤔

bin/save_cache
```diff
@@ -37,6 +37,12 @@ if [ -z "$CACHE_BUCKET_NAME" ]; then
 	fi
 fi
 
+# Use with caution – in general it's not a good idea to overwrite a cache entry
+if [ $3 == '--force' ]; then
```
mokagio: Buildkite (or the machine on which the agent run, I guess?) threw this error/warning running this code in [this build](https://buildkite.com/wordpress-mobile/wordpress-ios/builds/1681#f18e48c4-ffdd-4b40-bb56-58ce333fec81/323-347)

```
path/to/installed/pluging/bin/save_cache: line 41: [: ==: unary operator expected
```

My guess is this can be fixed by using `[[`:

```suggestion
if [[ $3 == '--force' ]]; then
```

bin/cache_repo
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+AWS_BUCKET=$1
+
+DATE=$(date +"%Y-%m-%d")
+
+mkdir -p /tmp/$DATE
```
mokagio:
```suggestion
mkdir -p "/tmp/$DATE"
```

For consistency with the quoted usage below. Or, vice versa we could remove the quotes for the usage below.

IMHO, better to err on the side of caution and quote everything.

bin/cache_repo
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
+
+set -e
+
+AWS_BUCKET=$1
+
+DATE=$(date +"%Y-%m-%d")
```
mokagio, line 7: Apologies if I missed where this was already discussed. But could you elaborate on the rationale of caching the the repo on a daily basis?

I noticed that the scheduled pipeline is [part of the WordPress iOS repo](https://github.com/wordpress-mobile/WordPress-iOS/pull/16935/files#diff-45d463ed168fd7f9dae18185bf9327f3b489aa63a91c0798a5bbeb5837f9d15f). Have you considered moving it to its own repository to centralize the schedule configuration? Or is the idea behind it that every repo has different caching needs, so it makes sense for each to setup its own caching "cron" job?

mokagio (commented):
Left a few questions

bin/cache_repo
jkmassel: > Have you considered moving it to its own repository to centralize the schedule configuration? 

Good question! The main reason for this is the fact that each project has a different SSH key, so having one project that does all of the cloning would be kinda complex, whereas we can set-it-and-forget it this way.

> could you elaborate on the rationale of caching the the repo on a daily basis

For sure – we could do this on basically any basis, doesn't really matter – the main idea is that for large repos we don't need to fetch the whole thing from GitHub every time – instead we can use the cached repo as a [reference repo](https://git-scm.com/docs/git-clone#Documentation/git-clone.txt---reference-if-ableltrepositorygt) when cloning, which is significantly faster – the only thing downloaded from GitHub is differences since the last cache.

bin/cache_repo
jkmassel: > better to err on the side of caution and quote everything

Yeah, it _should_ be impossible for this to break, but better safe than sorry 🙂

bin/install_cocoapods
jkmassel: Yeah, I thought about DRY-ing this, but given the fact that it should never change, I feel like it's ok – the worst case if we break it is that builds take longer, which isn't the end of the world IMHO? And I think we'd notice it before long, so I went for the less complex route 🤷‍♂️

bin/save_cache
jkmassel: Yep – also the nightly job will regenerate the CocoaPods base caches too, so we use the `--force` flag for that process.

bin/install_cocoapods
jkmassel: Yeah, I like the DRY of it, but I feel like it'd be a bit harder to maintain?

bin/cache_repo
jkmassel: This was another case of "better safe than sorry" 😅

bin/cache_cocoapods
jkmassel: Yeah, this flows nicer!

bin/cache_repo
jkmassel: Addressed in 952ca4d91dad3f67d58d5cd2fe920990044fedab because I couldn't use the "automagically commit" button for some reason

bin/install_cocoapods
mokagio: Yeah. Fine by me. If we discover this needs to change (which I can't foresee ATM) we can always refine 👍

bin/cache_repo
mokagio: > we could do this on basically any basis, doesn't really matter – the main idea is that for large repos we don't need to fetch the whole thing from GitHub every time – instead we can use the cached repo as a reference repo when cloning, which is significantly faster – the only thing downloaded from GitHub is differences since the last cache.

This is super neat! 🤯 

Are the two rationales, for why using a local schedule pipeline and for the repo caching necessity, documented somewhere?

Finally, regarding the centralization, I guess by having the logic live in the plugin here, we centralize the core of the work: defining the logic. Each project can then opt-in and configure the caching schedule as it prefers. _**If**_ we'll need to, we could make it more flexible by allowing consumers to set their own key for the caching.

jkmassel: > we could make it more flexible by allowing consumers to set their own key for the caching

Certainly, but I figured we'd go for convention over configuration here – if both sides use the Buildkite pipeline slug then it "just works" – hopefully nobody ever has reason to change it 😅

jkmassel:
I addressed feedback so I'm just going to land this for now to unblock https://github.com/wordpress-mobile/WordPress-iOS/pull/16935!

Happy to adjust further later

mokagio:
@jkmassel 👍 

I started a review but never submitted it 🤦‍♂️ Not sure why, sorry you had to merge in that way

image

PR #9, opened 2021-11-15
Add SSH Key to Agent Function

skoonin:
Creates a small function to add a ssh key from an environment variable to the running ssh-agent. 

- This also creates a key file, this can be avoided if desired with something like `echo $NEW_SSH_KEY | ssh-add -`
- It does not clear existing keys
- [The reason it was created did not work due to an issue with how git clone works with authentication and deploy keys.](https://www.webfactory.de/blog/using-multiple-ssh-deploy-keys-with-github)

## Usage: 
`add_ssh_key_to_agent "${exported_SSH_KEY}" name-of-exported-ssh-key-file`

### DIFF
--- a/bin/add_ssh_key_to_agent
+++ b/bin/add_ssh_key_to_agent
@@ -1,0 +1,12 @@
+#!/bin/bash -eu
+
+NEW_SSH_KEY=$1
+NEW_SSH_KEY_NAME=$2
+
+# Create key in ~/.ssh dir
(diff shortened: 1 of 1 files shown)


skoonin:
@jkmassel just bumping this PR

jkmassel (approved):

skoonin:
@jkmassel in the end, it turned out this didn't work with github.com due to how they authenticate ssh connections.  This function seems useful, but not sure if it will ever get used since it didn't meet the use-case?

PR #12, opened 2021-11-28
Fix missing quotes in shell scripts

AliSoftware:
I've run [`shellcheck`](https://github.com/koalaman/shellcheck) on all our shell scripts in `bin` and with the default options it already found some issues:

 - [`SC2086`](https://www.shellcheck.net/wiki/SC2046) – aka missing quotes around variable usage
 - [`SC2034`](https://www.shellcheck.net/wiki/SC2034) – aka unused variable
 - [`SC2143`](https://www.shellcheck.net/wiki/SC2143) – aka misuse of `[ … grep ]`

### Violations being fixed in this PR

For now I've only auto-fixed the ones about missing quotes (`SC2086`) using this command to generate a git diff to get the suggested autocorrection, then apply it:
```
find bin -type f -print0 | xargs -0 shellcheck -i SC2086 -f diff | git apply
```

### Other violations to fix

For now this PR only contains the result of fixing that SC2086 issue in existing scripts, as I didn't want to spend too much time on this today.

But we should also consider fixing the remaining ones manually, especially since most of them should be easily fixable

 - Fixing the other default violations (manually, since they don't have auto-correct), especially `SC2034` and `SC2143`
 - Run `shellcheck` with all optional checks turned on, and consider fixing those violations too

For information here are the links to the other kind of violations that `find bin -type f -print0 | xargs shellcheck -o all -W20` prints:

 - https://www.shellcheck.net/wiki/SC2034 – unused variable
 - https://www.shellcheck.net/wiki/SC2046 – quote around variables (†)
 - https://www.shellcheck.net/wiki/SC2154 – uninitialized variable (without any fallback mechanism in place)
 - https://www.shellcheck.net/wiki/SC2312 – command exit code ignored
 - https://www.shellcheck.net/wiki/SC2143 – Misuse of `grep` in `[ … ]`  test condition, use `grep -q` instead
 - https://www.shellcheck.net/wiki/SC2236 – Use `-n` instead of `! -z`
 - https://www.shellcheck.net/wiki/SC2243 – Prefer explicit use of `-n`
 - https://www.shellcheck.net/wiki/SC2250 – Prefer putting braces around variable references
 - https://www.shellcheck.net/wiki/SC2292 – Prefer `[[ ]]` over `[ ]`

I didn't get to them myself in that PR, but feel free to either add commits to this PR to fix them manually, or to fix them in a separate PR once this one is merged.

_(†) there's still one unfixed by this PR (in `bin/install_cocoapods:18`) because there is a conflicting `SC2143` violation on that same line so it wasn't able to auto-fix it until `SC2143` is addressed._

### Automate this

In parallel to this one-time fix, we should also configure our CI to run `shellcheck` on this repo on each PR.
I've opened a platform request (ref: pdnsEh-4c-p2) to handle this at some point.
Should be an easy win as well, I might tackle if during HACK week if you don't get to it beforehand, just didn't plan to spend my Sunday on it 😅


bin/build_and_test_pod
```diff
@@ -24,4 +24,4 @@ echo "--- :test_tube: Building and Running Tests"
 # https://github.com/Automattic/buildkite-ci/issues/7
 xcrun simctl list >> /dev/null
 
-bundle exec fastlane $LANE_NAME
+bundle exec fastlane "$LANE_NAME"
```
AliSoftware: ⚠️  We should extra-check that none of our existing pipelines rely on that glitch/bug btw.

Like if one calls `build_and_test_pod "test param1:value1"` in an attempt to pass custom params to the test lane (which seems like a reasonable need to have), this fix will now break that behavior. Imho that's still a good thing as the proper solution instead would be to make `build_and_test_pod` accept optional parameters (in addition to `$1` aka `$LANE_NAME`) explicitly, to make it clear that this is supported, instead of making caller rely on the bug and implementation details.

bin/cache_repo
```diff
@@ -9,11 +9,11 @@ rm -rf "/tmp/$DATE"
...
+git clone --mirror "$BUILDKITE_REPO" "/tmp/$DATE/$BUILDKITE_PIPELINE_SLUG.git"
 
 # Create the tarball
 TAR_NAME=$DATE.git.tar
-tar -C "/tmp/$DATE" -cvf $TAR_NAME $BUILDKITE_PIPELINE_SLUG.git
+tar -C "/tmp/$DATE" -cvf "$TAR_NAME" "$BUILDKITE_PIPELINE_SLUG".git
 
 # Copy the file to S3
-aws s3 cp $TAR_NAME s3://$AWS_BUCKET/$BUILDKITE_ORGANIZATION_SLUG/$BUILDKITE_PIPELINE_SLUG/$TAR_NAME
+aws s3 cp "$TAR_NAME" s3://"$AWS_BUCKET"/"$BUILDKITE_ORGANIZATION_SLUG"/"$BUILDKITE_PIPELINE_SLUG"/"$TAR_NAME"
```
AliSoftware: This is what the `shellcheck -f diff` suggested, but I think quoting the whole URL at once, in one big quotation, would work as well? 🤷

bin/build_and_test_pod
AliSoftware: Actually there's another use case where this would break, which is when passing an explicit platform to run either ios or macos tests.

🚨  This is actually what's used [in MobilePay-Apple here](https://github.com/Automattic/MobilePay-Apple/blob/develop/.buildkite/pipeline.yml#L15) – which this fix would break. 🚨 

We should thus fix this in another way instead, making the whole `build_and_test_pod` script allow arbitrary parameters other than `$LANE_NAME`, but instead likely use `"$@"` (which auto-expands to `"$1" "$2" "$3" …`, quoting each param in turn), and make it clear and explicit that _every_ parameter (`"%@"`) passed to that script will be passed along to `fastlane` unchanged, allowing you to pass arbitrary platform name before lane name and arbitrary parameters (like `foo:bar` but also flags like `--verbose`) to fastlane if needed.

Then we'd fix the call sites (like [in MobilePay-Apple](https://github.com/Automattic/MobilePay-Apple/blob/develop/.buildkite/pipeline.yml#L15)) to remove the single quotes around `'ios test'` (which are making it a single parameter with space, exploiting the bug) so that they become 2 parameters properly processed via `$@` in that `build_and_test_pod` command.

bin/add_host_to_ssh_known_hosts
```diff
@@ -7,4 +7,4 @@ URL="${1:?You need to provide an URL as first parameter}"
 [[ $URL =~ ^(https?://|git@)?([^/:]+) ]] && HOST=${BASH_REMATCH[2]}
 
 echo "Adding ${HOST} to '~/.ssh/known_hosts'..."
-for ip in $(dig @8.8.8.8 "${HOST}" +short); do ssh-keyscan "${HOST}",$ip; ssh-keyscan $ip; done 2>/dev/null >> ~/.ssh/known_hosts || true
+for ip in $(dig @8.8.8.8 "${HOST}" +short); do ssh-keyscan "${HOST}","$ip"; ssh-keyscan "$ip"; done 2>/dev/null >> ~/.ssh/known_hosts || true
```
jkmassel, line 10: Hmm – assuming a compatible `dig` it shouldn't be possible for the IP address to have spaces 🤔 

That said, I can see doing this for style reasons, but it's _technically_ not necessary...

bin/cache_repo
jkmassel: Yeah, IINM this would actually break the URL since it would interpret the `"` as components of the URL, no?

bin/build_and_test_pod
jkmassel: It should be possible to step this out – apply this fix and ship a new version of the plugin, then update that repo. Assuming the projects are well-behaved, they should always be pulling based on a tag.

I like the idea of the "splat operator" approach where we could "pass through" args.

jkmassel (changes requested):

bin/cache_repo
AliSoftware: Nope the way this diff is done won't break, it's totally legal in shell to have quotes in the middle of a string and it will still eat those quotes, not keep them verbatim. So this will have the intended behavior and the same behavior as is we quoted the whole strings as once.

In other words, in bash all these are equivalent and won't eg any literal quote:

```
echo "Hello foo/bar/baz"
echo "Hello "foo/"bar"/"baz"
echo Hello\ foo/bar/baz
echo He"llo fo"o/b"ar/baz"
```

The only case when a double quote would be kept as a literal even in the middle of a string is if that double quote is either escaped with a backslash, or if it is itself inside a single-quoted string:

```
echo '"Hello", she said.'
echo \"Hello\",\ she\ said.
```

---

So TL;DR:
 - the change made by shellcheck here is totally valid and won't introduce literal quotes
 - quoting the whole URL at once will produce the exact same result. Which since it's arguably more readable for humans, is likely what we should do instead. But if we don't and keep that suggested diff instead that will still work.

bin/add_host_to_ssh_known_hosts
AliSoftware: Yep I agree. But I'd still do it for style reasons … and for making shellcheck happy, especially if we plan to make it run regularly via CI at some point

bin/cache_repo
AliSoftware: Addressed in af3e002

bin/build_and_test_pod
AliSoftware: Reverted in d08a95f (+ added comment).

I'll make a separate PR on top of this to include that fix separately in preparation for a future breaking version bump.

AliSoftware: Separate PR to prepare the major version bump now ready here: https://github.com/Automattic/bash-cache-buildkite-plugin/pull/13

jkmassel (approved):
Approved pending merge of https://github.com/Automattic/bash-cache-buildkite-plugin/pull/13

AliSoftware:
> Approved pending merge of #13

Well up to you to decide the order in which you want to merge this, but the whole point for me to separate #12 from #13 was, per your request, for you to be able to do a non-breaking version with #12 first, and a breaking release with #13 next, wasn't it? In other words, the plan was to merge #12, do a minor tag, and only then merge #13 and do a major tag.

(I don't mind if you go straight to {merge #12 + #13 both and do a single, major version tag} directly, but then I'm not sure what the point was to make me split them in two 😅 )

PR #11, opened 2021-11-25
Update save_cache

jkmassel:

PR #15, opened 2021-12-07
Enable shellcheck on files in bin/* too

AliSoftware:
Implements the Platform Request pdnsEh-4c-p2 to lint all our shell scripts in `bin/`.

This PR is cut from the `shellcheck` branch to build on top of the violations I already fixed in #12.

### Fixing Current Violations

It is ready for review, even though the CI to currently red. This is because we _do_ have some violations in our current shell scripts that we'll need to fix first

Having the CI be currently red is a good way to show that when we have shellcheck violations they are properly reported.

But now that we got that [initial `shellcheck` report on CI](https://buildkite.com/wordpress-mobile/bash-cache-buildkite-plugin/builds/136#9167f396-760c-4440-833a-5553fb5e1096), we should fix the existing violations first via the other PRs before rebasing this PR on top of the fixes and merging it to `trunk`. See the [comment below](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/15#issuecomment-988058530) for a suggestion on the plan around the order in which we should merge the various currently-open PRs, to make everything work nicely and be fixed in the right order.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -2,7 +2,7 @@ steps:
   - label: ":shell: Lint (Shellcheck)"
     plugins:
       shellcheck#v1.1.2:
-        files: hooks/**
+        files: ["hooks/**", "bin/**"]
 
   - label: ":sparkles: Lint (Buildkite Plugin Linter)"
     plugins:
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -15,7 +15,7 @@ restore_cache "$LOCAL_CACHE_KEY"
 # If the `pod check` plugin is installed, use it to determine whether or not to install Pods at all
 # If it's not installed (or if it fails), we'll try to install Pods. 
 # If that fails, it may be due to an out-of-date repo. We can use `--repo-update` to try to resolve this automatically.
-if [ $(bundle exec pod plugins installed | grep check) ]; then
+if bundle exec pod plugins installed | grep -q check; then
 	bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
 else
 	bundle exec pod install || bundle exec pod install --repo-update --verbose
--- a/bin/publish_pod
+++ b/bin/publish_pod
@@ -2,7 +2,7 @@
 
 PODSPEC_PATH=$1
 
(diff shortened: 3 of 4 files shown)


AliSoftware:
[Here's a link to the initial CI run which failed as expected, reporting our currently remaining violations in our `bin/**` bash scripts](https://buildkite.com/wordpress-mobile/bash-cache-buildkite-plugin/builds/136#9167f396-760c-4440-833a-5553fb5e1096).

---

As mentioned in the PR description, we should probably fix the violations reported before rebasing and merging.

Note that:
 - For the violation in `build_and_test_pod`, the proper fix will require a breaking change in behavior of the script, but #13 already addresses this.
 - #11, which was cut from `trunk` after the commit this branch train was cut from, includes one more violation (see https://github.com/Automattic/bash-cache-buildkite-plugin/pull/11#pullrequestreview-822908275). Which means that if we merge #11 without fixing this violation first (which would then not be reported by shellcheck until both #11 and this PR are merged, since they are currently on divergent branch paths), this would make `trunk` go red; so probably best to fix it beforehand.

Therefore, we should just agree on the order in which we want to merge those PRs. I'm suggesting the following, but open to other plans:
 - merge #12 first
 - then fix the additional violations mentioned in #11 (not present in the branch from #12), and merge it
 - then publish a minor version (1.8.0)
 - then merge #13 – which is a breaking change
 - then rebase this #15 on top – to ensure that once rebased on top of all the violation fixes from previous PRs, the CI finally goes green on this one
 - then do a major version (given the breaking change from #13), and fix all the call sites that relied on the bug fixed in #13 (at least the [MobilePay-Apple](https://github.com/Automattic/MobilePay-Apple/blob/develop/.buildkite/pipeline.yml#L15) pipeline, maybe others?)

AliSoftware:
For the record, the changes you made with the 4 additional commits you pushed on this PR all make sense to me.
(GitHub does not allow me to mark my own PR as approved, but take this as a ✅ )

I forgot those would be necessary when I wrote my previous comment, but it was indeed expected that those violations would remain to be fixed (what you did in those 4 commits), since #12  only fixed one type of violation (`SC2086`) but didn't fix the other ones (especially `SC2034` and `SC2143` which I mention in #12 as being reported byt not being fixed by the PR). So that all make sense to me.

jkmassel (approved):

PR #19, opened 2022-03-25
Add check to ensure `pod install` doesn't change `Podfile.lock`

mokagio:
This PR:

- Adds a check to ensure that the tracked `Podfile.lock` matches what the `Podfile` would generate on a clean install. This should prevent issues such as https://github.com/wordpress-mobile/WordPress-iOS/pull/18145 and https://github.com/wordpress-mobile/WordPress-iOS/pull/18211
- Updates the hashing logic to work on both macOS and Linux after a recent change

~I created https://github.com/wordpress-mobile/WordPress-iOS/pull/18212 to test this change, but I noticed [something quite odd](https://buildkite.com/automattic/wordpress-ios/builds/6247#d5d69166-6260-4fd3-8a01-0dec82e3d316):~

~`/usr/local/var/buildkite-agent/plugins/github-com-automattic-bash-cache-buildkite-plugin-5015983/bin/hash_file:
line 8: sha256sum: command not found`~

~~For reference, the `sha256sum` was introduced in https://github.com/Automattic/bash-cache-buildkite-plugin/pull/18.~~

See [this build](https://buildkite.com/automattic/wordpress-ios/builds/6347) for an example of how the logic catches issues and [this other build](https://buildkite.com/automattic/wordpress-ios/builds/6346#2918d078-cbd3-4dd9-9bc9-dbd9060784c7) to verify that, [once the issue is addressed](https://github.com/wordpress-mobile/WordPress-iOS/commit/323b7818e07162f02430214195a7a2a657a0edd4), everything is green.

image

### DIFF
--- a/bin/hash_file
+++ b/bin/hash_file
@@ -9,4 +9,13 @@ fi
 # printing the error message if `$1` was unbound.
 set -u
 
-sha256sum "$1" | cut -f1 -d " "
+# `shasum` is available on only macOS
+if command -v shasum &> /dev/null; then
+  sha_command='shasum -a 256'
+else
+  sha_command='sha256sum'
+fi
+
+# Both `shasum` and `sha256sum` will print the hash and the file name (`$1`).
+# We only care about the hash, so we use `cut` to extract it.
+$sha_command $1 | cut -f1 -d " "
--- a/bin/hash_file
+++ b/bin/hash_file
@@ -5,8 +5,8 @@ if [ -z "$1" ]; then
 	exit 1
 fi
 
-# Calling this here otherwise the `-z "$1"` above would have failed without
-# printing the error message if `$1` was unbound.
+# Calling `set -u` here instead of the shebang otherwise the `-z "$1"` above
+# would fail without printing the error message if `$1` was unbound.
 set -u
 
 # `shasum` is available on only macOS
--- a/bin/hash_file
+++ b/bin/hash_file
@@ -18,4 +18,4 @@ fi
 
 # Both `shasum` and `sha256sum` will print the hash and the file name (`$1`).
 # We only care about the hash, so we use `cut` to extract it.
-$sha_command $1 | cut -f1 -d " "
+$sha_command "$1" | cut -f1 -d " "
--- a/bin/hash_file
+++ b/bin/hash_file
@@ -1,8 +1,12 @@
-#!/bin/bash -eu
+#!/bin/bash -e
 
 if [ -z "$1" ]; then
 	echo "You must pass a filename to hash"
 	exit 1
 fi
+
+# Calling this here otherwise the `-z "$1"` above would have failed without
+# printing the error message if `$1` was unbound.
+set -u
 
 sha256sum "$1" | cut -f1 -d " "
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -13,7 +13,7 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
 restore_cache "$LOCAL_CACHE_KEY"
 
 # If the `pod check` plugin is installed, use it to determine whether or not to install Pods at all
-# If it's not installed (or if it fails), we'll try to install Pods. 
+# If it's not installed (or if it fails), we'll try to install Pods.
 # If that fails, it may be due to an out-of-date repo. We can use `--repo-update` to try to resolve this automatically.
 if bundle exec pod plugins installed | grep -q check; then
 	bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -32,7 +32,9 @@ else
 fi
 
 function lockfile_error () {
-  echo "Podfile.lock was changed by 'pod install'. Please run 'pod install' and try again."
+  message='`Podfile.lock` was changed by `bundle exec pod install`. Please run `bundle exec pod install` locally with a clean cache and commit the result.'
+  echo $message
+  buildkite-agent annotate "$message" --style 'error' --context 'ctx-error'
 }
 trap lockfile_error ERR
 
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -37,7 +37,8 @@ function lockfile_error () {
 trap lockfile_error ERR
 
 echo "Checking that Podfile.lock was not modified by 'pod install'"
-echo "${PODFILE_HASH} Podfile.lock" | $sha_command --check --status
+# Notice the two spaces as per shasum/sha256sum output
+echo "${PODFILE_HASH}  Podfile.lock" | $sha_command --check --status
 
 # Remove trap for the lockfile error now that we've done the check.
 trap - ERR
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -23,13 +23,21 @@ fi
 
 # Check that `Podfile.lock` was unchanged by `pod install`. If it was, it means
 # the lockfile might have been inadvertently changed.
+
+# `shasum` is available only on macOS
+if command -v shasum &> /dev/null; then
+  sha_command='shasum -a 256'
+else
+  sha_command='sha256sum'
+fi
+
 function lockfile_error () {
   echo "Podfile.lock was changed by 'pod install'. Please run 'pod install' and try again."
 }
 trap lockfile_error ERR
 
 echo "Checking that Podfile.lock was not modified by 'pod install'"
-echo "${PODFILE_HASH} Podfile.lock" | sha256sum --check --status
+echo "${PODFILE_HASH} Podfile.lock" | $sha_command --check --status
 
 # Remove trap for the lockfile error now that we've done the check.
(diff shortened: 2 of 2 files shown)


bin/install_cocoapods
```diff
@@ -13,13 +13,26 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
...
 
+# Check that `Podfile.lock` was unchanged by `pod install`. If it was, it means
+# the lockfile might have been inadvertently changed.
+function lockfile_error () {
+  echo "Podfile.lock was changed by 'pod install'. Please run 'pod install' and try again."
+}
+trap lockfile_error ERR
+
+echo "Checking that Podfile.lock was not modified by 'pod install'"
+echo "${PODFILE_HASH} Podfile.lock" | sha256sum --check --status
```
mokagio: `PODFILE_HASH` is defined at line 9.

bin/install_cocoapods
```diff
@@ -13,13 +13,26 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
...
 if bundle exec pod plugins installed | grep -q check; then
 	bundle exec pod check || bundle exec pod install || bundle exec pod install --repo-update --verbose
 else
 	bundle exec pod install || bundle exec pod install --repo-update --verbose
 fi
 
+# Check that `Podfile.lock` was unchanged by `pod install`. If it was, it means
+# the lockfile might have been inadvertently changed.
+function lockfile_error () {
+  echo "Podfile.lock was changed by 'pod install'. Please run 'pod install' and try again."
```
mokagio: My test build is giving us a false positive. It fails ecause of `sha256sum`, so it does not verify the actual `Pofile.lock` check logic. Still, the error printing behavior is correct.

image

Looking at it in the builds logs, I think we ought to make it more visible to developers. Ideas:

1. Colorize it
2. ~~Annotate the build~~

bin/install_cocoapods
```diff
@@ -13,13 +13,35 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
...
 
+# Check that `Podfile.lock` was unchanged by `pod install`. If it was, it means
+# the lockfile might have been inadvertently changed.
+
+# `shasum` is available only on macOS
+if command -v shasum &> /dev/null; then
+  sha_command='shasum -a 256'
+else
+  sha_command='sha256sum'
+fi
```
mokagio, line 32: You'll notice that this block of code is duplicated from the `bin/hash_file` file above.

I started thinking about extracting this and updating all the `shasum`/`sha256sum` call sites that are still around, but decided to stop and submit this PR as focused only on:

1. Fixing the hashing behavior (currently a problem for PocketCasts iOS)
2. Adding the `Podfile.lock` check

I'd love to keep working on this, but I think it's better to ship this fixes first.

mokagio:
@jkmassel this PR is now ready for review. See description for how I verified my changes.

mokagio:
Asked @AliSoftware for a review, too, in case Jeremy doesn't get to this soon.

bin/install_cocoapods
AliSoftware: Worth opening a GitHub issue to not forget about it?

bin/hash_file
```diff
@@ -1,8 +1,21 @@
...
 if [ -z "$1" ]; then
 	echo "You must pass a filename to hash"
 	exit 1
 fi
 
-sha256sum "$1" | cut -f1 -d " "
+# Calling `set -u` here instead of the shebang otherwise the `-z "$1"` above
+# would fail without printing the error message if `$1` was unbound.
+set -u
+
```
AliSoftware, line 11: Some other options for this if you're curious:
 - Use `${1:-}` instead of `$1`, which basically means "`$1` if it is defined, but defaults to empty string if undefined", which will thus avoid triggering the `-u` "unset variable" error while still correctly work for the `-z` test
 - Use `${1:?You must pass a filename to hash}` which is a built-in construct which takes care of printing the error message and exiting if `$1` is unset

Source: https://devhints.io/bash

Personally I feel like the 2nd option, while nice to know, might be a bit too obscure (and little known syntax), but I do like the 1st one.

_(Tried to post this as a GitHub suggestion but apparently I can't post a GH code suggestion to modify line 3, as it was not a line changed by the PR in the first place 🤷  so if you agree with the idea I'll let you do the change manually by reverting line 1, changing line 3 and deleting those lines 8-11.)_

bin/install_cocoapods
```diff
@@ -13,13 +13,37 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
 restore_cache "$LOCAL_CACHE_KEY"
 
 # If the `pod check` plugin is installed, use it to determine whether or not to install Pods at all
-# If it's not installed (or if it fails), we'll try to install Pods. 
+# If it's not installed (or if it fails), we'll try to install Pods.
 # If that fails, it may be due to an out-of-date repo. We can use `--repo-update` to try to resolve this automatically.
```
AliSoftware, line 17: Nit, but… is that logic of checking if `pod check` is installed really necessary?

I mean, if `pod check` is not installed, then `bundle exec pod check` will just exit with a non-zero status code (not because it would detect `Podfile.lock` to be outdated but because it would not recognize the command), and the `||` would then take care to falling back to `bundle exec pod install` in such case anyway and do the same thing as what the `else` is doing…

Or is this test there mostly to avoid the risk of having a misleading error message in the logs about `pod check` not being installed? (though if that's it I'm sure a `2>/dev/null` would be enough to solve it?)

🤷

bin/install_cocoapods
AliSoftware: Also, I have no problem keeping the larger "update all the call sites" for a later. But given that you already fixed the `hash_file` in this very PR, why not call `hash_file` here in `install_cocoapods` instead of duplicating its code? 🤷 
 (while still only keep the work to update _other_ call sites for later if you want)

bin/install_cocoapods
```diff
@@ -13,13 +13,37 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
...
+else
+  sha_command='sha256sum'
+fi
+
+function lockfile_error () {
+  message="\`Podfile.lock\` was changed by \`bundle exec pod install\`. Please run \`bundle exec pod install\` locally with a clean cache and commit the result."
+  echo "$message"
+  buildkite-agent annotate "$message" --style 'error' --context 'ctx-error'
+}
+trap lockfile_error ERR
```
AliSoftware, line 39: Nice use of `trap` here! 👍 👌

bin/install_cocoapods
```diff
@@ -13,13 +13,37 @@ LOCAL_CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-local-pod-cache-$PODFILE_HASH"
...
+if command -v shasum &> /dev/null; then
+  sha_command='shasum -a 256'
+else
+  sha_command='sha256sum'
+fi
+
+function lockfile_error () {
+  message="\`Podfile.lock\` was changed by \`bundle exec pod install\`. Please run \`bundle exec pod install\` locally with a clean cache and commit the result."
+  echo "$message"
+  buildkite-agent annotate "$message" --style 'error' --context 'ctx-error'
```
AliSoftware, line 37: Very nice touch to use the buildkite annotation feature here! ❤️ 

(We should think of using this feature more often, I'm sure we have more use cases in other actions and pipelines for which this could be a nice thing to use 🤔 💭 )

AliSoftware (approved):
Approving to unblock, though left a couple of small nits that would be nice to address before landing if you can 🙂

bin/install_cocoapods
mokagio: > Worth opening a GitHub issue to not forget about it?

Good idea, thanks. 👉 Created #21.

> why not call `hash_file` here...

Because in `hash_file` we use `sha*` to get the hash, but here we use it to check the hash matches (`$sha_command --check --status`).

I guess we could update `hash_file` to use more than just `$1` and so pass the options through, but that seems more in scope with #21.

AliSoftware: Ah, gotcha. Fair enough 👍

bin/install_cocoapods
mokagio: Thanks 😄 I copy pasted it [from the CircleCI orb](https://github.com/wordpress-mobile/circleci-orbs/blame/25eee7663ad3d9fc908c7602ae3262379ea33c61/src/ios/orb.yml#L501-L504)... Kudos to @jtreanor .

bin/install_cocoapods
mokagio: I'm not sure. Plus, I had a quick look at the time and couldn't find that plugin installed in our projects. Might be best to subtract this and simplify.

jkmassel (approved):

bin/hash_file
mokagio: I'm going to merge this PR and followup with a dedicated one for this tweak.

PR #22, opened 2022-04-06
Simplify `hash_file` by defaulting `$1` to empty string in `if` check

mokagio:
Hat tip @AliSoftware. See https://github.com/Automattic/bash-cache-buildkite-plugin/pull/19#discussion_r839393747

To test, run `bin/hash_file` with and without a file as argument.

image


AliSoftware (approved):
👌

PR #27, opened 2022-06-29
Add command to annotate Unit/UI Test failures on the Buildkite page

AliSoftware:
The PR provides a command to annotate Buildkite builds with annotations showing test failures at the top of the Buildkite job page.

## Context

This work is a follow-up of what I've been playing with in https://github.com/wordpress-mobile/WordPress-iOS/pull/18952 — this mainly moves the feature from the client repo into this plugin so that more projects can benefit from this.

> **Note** I've reimplemented the logic from using an XSLT (what I used in the WPiOS PR originally) to using a ruby script, because while XSLTs are nice and powerful, and `xsltproc` is installed by default on every macOS standard installation and thus each `mac` agent and VM we use in Buildkite, this might not be available in other client repos using other agents like `android`. While having ruby available in the build agents is more common (e.g. both iOS and Android clients use `fastlane` and thus have `ruby` installed in some capacity)

## To Test

This new command is used in https://github.com/wordpress-mobile/WordPress-iOS/pull/18961, so if that WPiOS PR properly reports test failures as Buildkite annotations, we can consider this working.

### DIFF
--- a/bin/annotate_test_failures
+++ b/bin/annotate_test_failures
@@ -5,7 +5,7 @@
 #
 require 'rexml/document'
 
-junit_path = ARGV.first || File.join('build/results/report.junit')
+junit_path = ARGV.first || File.join('build', 'results', 'report.junit')
 title = ENV.fetch('BUILDKITE_LABEL', 'Tests')
 
 file = File.open(junit_path)
@@ -13,17 +13,17 @@ xmldoc = REXML::Document.new(file)
 
 failures = []
 REXML::XPath.each(xmldoc, '//testcase[failure]') do |testcase|
-    failure = testcase.elements['failure']
-    failures.append <<~ENTRY
-    <details><summary><tt>#{testcase['name']}</tt> in <tt>#{testcase['classname']}</tt></summary>
-    
-    #{failure['message']}
+  failure = testcase.elements['failure']
+  failures.append <<~ENTRY
+  <details><summary><tt>#{testcase['name']}</tt> in <tt>#{testcase['classname']}</tt></summary>
+  
+  #{failure['message']}
 
-    ```
-    #{failure.text}
-    ```
-    </details>
-    ENTRY
+  ```
+  #{failure.text}
+  ```
+  </details>
+  ENTRY
 end
 
 markdown = "\#\#\#\# #{title}: #{failures.length} failure(s)\n\n" + failures.join("\n")
--- a/bin/annotate_test_failures
+++ b/bin/annotate_test_failures
@@ -8,6 +8,11 @@ require 'rexml/document'
 junit_path = ARGV.first || File.join('build', 'results', 'report.junit')
 title = ENV.fetch('BUILDKITE_LABEL', 'Tests')
 
+unless File.exist?(junit_path)
+  puts "JUnit file not found: #{junit_path}."
+  exit 0
+end
+
 file = File.open(junit_path)
 xmldoc = REXML::Document.new(file)
 
--- a/bin/annotate_test_failures
+++ b/bin/annotate_test_failures
@@ -1,0 +1,30 @@
+#!/usr/bin/env ruby
+#
+# Usage:
+#   annotate-test-failures [junit-file-path]
+#
+require 'rexml/document'
+
+junit_path = ARGV.first || File.join('build/results/report.junit')
+title = ENV.fetch('BUILDKITE_LABEL', 'Tests')
+
+file = File.open(junit_path)
+xmldoc = REXML::Document.new(file)
+
+failures = []
+REXML::XPath.each(xmldoc, '//testcase[failure]') do |testcase|
+    failure = testcase.elements['failure']
+    failures.append <<~ENTRY
+    <details><summary><tt>#{testcase['name']}</tt> in <tt>#{testcase['classname']}</tt></summary>
+    
+    #{failure['message']}
+
+    ```
+    #{failure.text}
+    ```
+    </details>
+    ENTRY
+end
+
+markdown = "\#\#\#\# #{title}: #{failures.length} failure(s)\n\n" + failures.join("\n")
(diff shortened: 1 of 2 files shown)


bin/annotate_test_failures
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env ruby
+#
+# Usage:
+#   annotate_test_failures [junit-file-path]
+#
+require 'rexml/document'
```
AliSoftware, line 6: > **Note** this is part of standard ruby installations (no need to `gem install` it or whatnot, it's bundled in ruby, even if we still need to `require` it to use it) so no risk of this script depending on a `Gemfile` or installed gems or whatever, it's still fully self-contained.

bin/annotate_test_failures
```diff
@@ -0,0 +1,43 @@
...
+
+junit_path = ARGV.first || File.join('build', 'results', 'report.junit')
+title = ENV.fetch('BUILDKITE_LABEL', 'Tests')
+
+unless File.exist?(junit_path)
+  puts "JUnit file not found: #{junit_path}."
+  exit 0
```
AliSoftware, line 14: > **Note** Exiting with status 0 (no error) is intentional here. We don't want the `annotate_test_failures` command to fail for builds where the "test" step failed for reasons other than a test failure (e.g. a crash of the test runner, or a failure to even launch the app, or a compilation issue…) and which would thus not have created the `.junit` file, because failing on this annotation step in those cases (instead of reporting the earlier failure at a higher level in the caller script) could otherwise be misleading and confusing.

bin/annotate_test_failures
```diff
@@ -0,0 +1,43 @@
...
+  ```
+  #{failure.text}
+  ```
+  </details>
+  ENTRY
+end
+
+if failures.empty?
+  puts "No test failure found. Removing any previous `#{title}` Buildkite annotation."
+  system('buildkite-agent', 'annotation', 'remove', '--context', title)
```
AliSoftware, line 37: Ok I kinda hoped that `remove` would fail silently if there was no annotation with that context — so that if there's one it deletes it but if there isn't it would be a no-op — but apparently that's not the case, as I just got this:

```
  | 2022-06-29 18:34:33 DEBUG  DELETE https://agent.buildkite.com/v3/jobs/0181b09e-4dca-44e4-8610-8b83fdd87da9/annotations/%F0%9F%94%AC%20Unit%20Tests
  | 2022-06-29 18:34:33 DEBUG  ↳ DELETE https://agent.buildkite.com/v3/jobs/0181b09e-4dca-44e4-8610-8b83fdd87da9/annotations/%F0%9F%94%AC%20Unit%20Tests proto=HTTP/1.1 status=410 Δ=181.218772ms
  | 2022-06-29 18:34:33 FATAL  Failed to remove annotation: DELETE https://agent.buildkite.com/v3/jobs/0181b09e-4dca-44e4-8610-8b83fdd87da9/annotations/%F0%9F%94%AC%20Unit%20Tests: 410 Annotation no longer exists
```

AliSoftware:
Buildkite job that had some failures — one of them that I created intentionally by connecting on the VM via ssh while the UI Tests were running, then killing `wiremock` to make the test fail.:
image

Notice that:
 - There were 2 annotations ("Unit Tests" and "UI Tests (iPhone)")
 - But only the "Unit Tests" job is marked in red and as failed in the summary
 - This is likely because, for the "UI Tests (iPhone)" step, it seems that there was some internal retry mechanism that got kicked… yet the `report.junit` still kept the 2 failed tests marked as failures, while the logs suggests that there were failures but ends with "Tests succeeded"…

```
  | [20:10:54]: ▸ WordPressUITests.EditorGutenbergTests
  | [20:10:54]: ▸   testAddGalleryBlock, failed: caught error: "timedOut"
  | [20:10:54]: ▸   <unknown>:0
  | [20:10:54]: ▸   testAddGalleryBlock, Failed to get automation session for org.wordpress:5810: Timed out while requesting automation session for org.wordpress (pid:5810).
  | [20:10:54]: ▸   <unknown>:0
  | [20:10:54]: ▸ 	 Executed 16 tests, with 1 test skipped and 2 failures (1 unexpected) in 1997.767 (1997.789) seconds
  | [20:10:54]: ▸
  | [20:11:13]: ▸ 2022-06-29 20:11:13.373 xcodebuild[5355:14972] [MT] IDETestOperationsObserverDebug: 2098.462 elapsed -- Testing started completed.
  | [20:11:13]: ▸ 2022-06-29 20:11:13.373 xcodebuild[5355:14972] [MT] IDETestOperationsObserverDebug: 0.000 sec, +0.000 sec -- start
  | [20:11:13]: ▸ 2022-06-29 20:11:13.373 xcodebuild[5355:14972] [MT] IDETestOperationsObserverDebug: 2098.462 sec, +2098.462 sec -- end
  | [20:11:13]: ▸ Test execute Succeeded
  | +--------------------+--------------------+
  | |              Test Results               |
  | +--------------------+--------------------+
  | | Number of tests    | 14 (and 2 retries) |
  | | Number of failures | 0                  |
  | +--------------------+--------------------+
```

bin/annotate_test_failures
AliSoftware: Addressed in [9c9b755](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/27/commits/9c9b755d001d21bd71a7229eeb14222336b235de).

bin/annotate_test_failures
mokagio: Have you considered adding this information to the code?

```suggestion
  # Don't fail if `.junit` file not found to avoid "hiding" the failure in the build that resulted in the that file not being created
  exit 0
```

mokagio (approved):
Great work. Thank you for moving this in the plugin so we can use it in other projects.

I had a look at https://github.com/wordpress-mobile/WordPress-iOS/pull/18961, which reported the failures as expected 👍

PR #28, opened 2022-07-13
Ensure we're using the latest version of `bundler` before running `bundle install`

### LINKED ISSUE #16: Use latest Bundler or version from `Gemfile.lock`?
## Background

While working on [adding a pipeline to rebuild the pods and Git cache on a schedule for Pocket Casts iOS](https://github.com/shiftyjelly/pocketcasts-ios/pull/3667/) I run into what looks like a Bundler related issue:

```
bundler: failed to load command: pod (/usr/local/var/buildkite-agent/builds/builder/wordpress-mobile/pocket-casts-ios/vendor/bundle/ruby/2.7.0/bin/pod)

Gem::LoadError: You have already activated bundler 2.2.28, but your Gemfile requires bundler 2.1.4. Since bundler is a default gem, you can either remove your dependency on it or try updating to a newer version of bundler that supports bundler as a default gem.
```

See:

- [The first time I got the error](https://buildkite.com/wordpress-mobile/pocket-casts-ios/builds/7434#2577bfdc-187a-47e0-a415-90dd8415c960/711-712)
- [The build that gets the error again after I reverted the fix as a test](https://buildkite.com/wordpress-mobile/pocket-casts-ios/builds/7444#4864878c-7ae6-4826-8ec6-a7048e3a5593/536-537)

The only way I managed to work around it was by calling `gem install bundler:2.2.32` (the latest version at the time of writing) before calling `install_gems`. Green build [here](https://buildkite.com/wordpress-mobile/pocket-casts-ios/builds/7441).

The issue, as far as I can place it, is that the machine is running on a 2.1.x version of Bundler.

## Suggestion / Question

I wonder if it would be safe, or useful, to parse the `Gemfile.lock` for the `BUNDLE WITH` value and install that version of Bundler before running other commands in those actions that depend of `bundle exec` 🤔 

Alternatively, if we can prove the issue is just a matter of `2.1.x` vs `2.2.x` quirks, we might want to ensure we run on the latest `2.2.x` as part as our VMs boot process (which is of course out of scope for this repo).

AliSoftware:
Fixes https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16

Installs the latest version of `bundler` to avoid `Gem::LoadError` crashes when using mismatching versions.

### DIFF
--- a/bin/install_gems
+++ b/bin/install_gems
@@ -7,7 +7,9 @@ CACHEKEY="$BUILDKITE_PIPELINE_SLUG-$ARCHITECTURE-ruby$RUBY_VERSION-$GEMFILE_HASH
 
 restore_cache "$CACHEKEY"
 
-# Ensure we're using the latest version of Bundler before running bundle install
+# Ensure we're using the latest version of Bundler before running bundle install.
+#
+# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16  for more details on why this is recommended.
 gem install bundler
 bundle install
 
--- a/bin/install_gems
+++ b/bin/install_gems
@@ -6,6 +6,9 @@ GEMFILE_HASH=$(hash_file Gemfile.lock)
 CACHEKEY="$BUILDKITE_PIPELINE_SLUG-$ARCHITECTURE-ruby$RUBY_VERSION-$GEMFILE_HASH"
 
 restore_cache "$CACHEKEY"
+
+# Ensure we're using the latest version of Bundler before running bundle install
+gem install bundler
 bundle install
 
 # If this is the first time we've seen this particular cache key, save it for the future


oguzkocer (approved):
This PR review seems to have fallen through the cracks. Sorry about that @AliSoftware!

---

I am approving this PR because I don't think updating the `bundler` would have any negative impact. However, with the latest `Bundler` versions, [the Bundler version itself is locked](https://bundler.io/blog/2022/01/23/bundler-v2-3.html) in the `Gemfile.lock`, so I am curious about how that impacts our approach. Particularly, I wonder if we need to be installing the locked version of the Bundler, and if so, if there is a direct way to do that without parsing the `Gemfile.lock`.

> Automatic update of Bundler when running bundle install without a lockfile.

This is shared as a next step [in the blog post I shared above](https://bundler.io/blog/2022/01/23/bundler-v2-3.html), so I guess running `bundle install` will be enough in the future versions.

_P.S:_ I disabled auto-merge to give @AliSoftware a chance to read my comment.

AliSoftware:
Good point, and if I remember correctly, parsing the version from the `Gemfile.lock` was another option that was considered in the description of https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16 that this PR fixes.

I went with the simpler version of installing the latest (`gem install bundler`) because that's what we have been doing in practice in all places in our pipelines where we applied the workaround locally (e.g. [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/.buildkite/commands/run-ui-tests.sh#L25-L27) or [here](https://github.com/wordpress-mobile/WordPress-iOS/blob/trunk/.buildkite/commands/installable-build-wordpress.sh#L3-L5)).

So the aim of that PR was simply moved applying that workaround from each individual pipeline to move it to the `install_gems` command directly, as [was initially suggested in the original issue](https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16#issuecomment-1164105306)

_I should have included all that in my PR description, sorry for not providing all the pointers in the first place._

AliSoftware:
PS: @rynaardb also mentioned in https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16#issuecomment-1184639099 that they should have now fixed this for new VMs that will be provisioned, via 77-gh-Automattic/buildkite-ci, by disabling `bundler` as a default gem when creating new VMs.

That being said, as @mokagio pointed out [in the subsequent comment](https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16#issuecomment-1188216245), it will probably take some time before we start using new VMs  provisioned with the fix from 77-gh-Automattic/buildkite-ci.
Sure @crazytonyli is currently provisioning new VMs for Xcode 14 betas, but we won't use Xcode 14 until it's final (likely end of September / start of October), if not later (as we might need to fix potential compilation issues on the projects adopting Xcode 14 before moving their CI to those new VMs). So in the meantime it's nice to have this workaround be applied and work even for older VMs.

wdyt @jkmassel  @rynaardb ? Given the above:
 - Should we ship that PR as is (i.e. same workaround that we use in client apps) to cover existing VMs?
 - Should we change it to parse the bundler version from the `Gemfile.lock` to be more accurate?
 - Should we consider that we already fixed it locally everywhere, and new VMs will be deployed and used by all repos soon enough, for this workaround not to be worth it anymore?

bin/install_gems
```diff
@@ -6,6 +6,9 @@ GEMFILE_HASH=$(hash_file Gemfile.lock)
 CACHEKEY="$BUILDKITE_PIPELINE_SLUG-$ARCHITECTURE-ruby$RUBY_VERSION-$GEMFILE_HASH"
 
 restore_cache "$CACHEKEY"
+
+# Ensure we're using the latest version of Bundler before running bundle install
```
mokagio: Might be worth adding a note about why it is necessary to do this?

A link to https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16 might be enough.

```suggestion
# Ensure we're using the latest version of Bundler before running bundle install.
#
# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16  for more details on why this is recommended.
```

mokagio (approved):
Explicit approval* to followup on the conversation from https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16#issuecomment-1248889569

---

* – My approval technically is unnecessary because `@oguzkocer` already did approve this.

PR #29, opened 2022-07-26
Add Test Analytics Upload Action

jkmassel:
Adds a Test Analytics Action (used by https://github.com/wordpress-mobile/WordPress-FluxC-Android/pull/2472) that makes uploading JUnit XML files a one-liner ([without any bash-pipe-curl stuff](https://github.com/buildkite/test-collector-junit)).

### DIFF
--- a/bin/upload_buildkite_test_analytics_junit
+++ b/bin/upload_buildkite_test_analytics_junit
@@ -1,21 +1,21 @@
 #!/bin/bash -eu
 
 if [ -z "${1: }" ]; then
-	echo "You must pass a path to the file you want to upload to Test Analytics"
-	exit 1
+  echo "You must pass a path to the file you want to upload to Test Analytics"
+  exit 1
 fi
 
 if [ -z "${2: }" ]; then
-	echo "You must pass a Buildkite Analytics Token"
-	exit 1
+  echo "You must pass a Buildkite Analytics Token"
+  exit 1
 fi
 
 FILE_PATH=$1
 ANALYTICS_TOKEN=$2
 
 if [ ! -f "$FILE_PATH" ]; then
-	echo "No file exists at $FILE_PATH. Cancelling Upload"
-	exit 0
+  echo "No file exists at $FILE_PATH. Cancelling Upload"
+  exit 0
 fi
 
 echo "Uploading ${FILE_PATH} to Test Analytics"
--- a/bin/upload_buildkite_test_analytics_junit
+++ b/bin/upload_buildkite_test_analytics_junit
@@ -15,6 +15,8 @@ ANALYTICS_TOKEN=$2
 
 if [ ! -f "$FILE_PATH" ]; then
   echo "No file exists at $FILE_PATH. Cancelling Upload"
+
+  # Clean exit, to avoid invalidating an entire run because of a missing artifact
   exit 0
 fi
 
--- a/bin/upload_buildkite_test_analytics_junit
+++ b/bin/upload_buildkite_test_analytics_junit
@@ -1,0 +1,35 @@
+#!/bin/bash -eu
+
+FILE_PATH=$1
+ANALYTICS_TOKEN=$2
+
+if [ -z "$FILE_PATH" ]; then
+	echo "You must pass a path to the file you want to upload to Test Analytics"
+	exit 1
+fi
+
+if [ -z "$ANALYTICS_TOKEN" ]; then
+	echo "You must pass a Buildkite Analytics Token"
+	exit 1
+fi
+
+if [ ! -f "$FILE_PATH" ]; then
+	echo "No file exists at $FILE_PATH. Cancelling Upload"
+	exit 0
+fi
+
+echo "Uploading ${FILE_PATH} to Test Analytics"
+
+curl -X POST \
+  -H "Authorization: Token token=$ANALYTICS_TOKEN" \
+  -F "format=junit" \
+  -F "data=@$FILE_PATH" \
+  -F "run_env[CI]=buildkite" \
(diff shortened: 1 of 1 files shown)


bin/upload_buildkite_test_analytics_junit
```diff
@@ -0,0 +1,35 @@
...
+fi
+
+if [ -z "$ANALYTICS_TOKEN" ]; then
+	echo "You must pass a Buildkite Analytics Token"
+	exit 1
+fi
+
+if [ ! -f "$FILE_PATH" ]; then
+	echo "No file exists at $FILE_PATH. Cancelling Upload"
+	exit 0
```
oguzkocer: What is the reason you picked `0` for the exit status? I assume it was a conscious choice, so I think it'd be good to document that to avoid confusion - like the one I am having now 😅

oguzkocer (approved):
@jkmassel The script seems straightforward enough that if the request succeeds, we should be fine to merge it. However, I had one question about the exit status, which would be good to take a look at before merging this in.

bin/upload_buildkite_test_analytics_junit
jkmassel: Hmm so there are two questions here I guess – the reason we're exiting cleanly was a decision that I went back-and-forth on, but eventually figured that it might be nicer to not have to have file existence logic in the parent script – we could just skip it if it's not found. 

The reason I used `exit 0` to indicate "exit cleanly" was just habit – sure it's identical to `exit` but when reading it's IMHO extra-clear that there's no exit error code.

Is this a good way to do it? Maybe not?

I could document it, but like...it's a style thing not functional thus feels sorta weird? 

Anyway, if the question you were asking is the first one LMK if you disagree – I can see good arguments both ways, so we could change it for sure.

oguzkocer: My question was the first one and I think a comment above the exit statement would be enough. I don't have a strong preference either way on that decision.

bin/upload_buildkite_test_analytics_junit
```diff
@@ -0,0 +1,35 @@
...
+  -F "data=@$FILE_PATH" \
+  -F "run_env[CI]=buildkite" \
+  -F "run_env[key]=$BUILDKITE_BUILD_ID" \
+  -F "run_env[number]=$BUILDKITE_BUILD_NUMBER" \
+  -F "run_env[job_id]=$BUILDKITE_JOB_ID" \
+  -F "run_env[branch]=$BUILDKITE_BRANCH" \
+  -F "run_env[commit_sha]=$BUILDKITE_COMMIT" \
+  -F "run_env[message]=$BUILDKITE_MESSAGE" \
+  -F "run_env[url]=$BUILDKITE_BUILD_URL" \
+  https://analytics-api.buildkite.com/v1/uploads
```
mokagio, line 37: @oguzkocer left a thoughtful comment and question.

Instead, I'm here to point out this command uses two spaces indentation while the rest of the file uses tabs. 🙃

mokagio (approved):
Pretty cool to see us using the raw APIs like real hackers.

Just a reference note to say the action did it's job nicely [in the build it was used](https://buildkite.com/automattic/wordpress-fluxc-android/builds/1509):

image

bin/upload_buildkite_test_analytics_junit
```diff
@@ -0,0 +1,35 @@
...
+
+if [ -z "$FILE_PATH" ]; then
+	echo "You must pass a path to the file you want to upload to Test Analytics"
+	exit 1
+fi
+
+if [ -z "$ANALYTICS_TOKEN" ]; then
+	echo "You must pass a Buildkite Analytics Token"
+	exit 1
+fi
```
mokagio: I don't think these `-z` checks will run, because of the `-u` in the shebang:

image

bin/upload_buildkite_test_analytics_junit
```diff
@@ -0,0 +1,35 @@
...
+
+if [ ! -f "$FILE_PATH" ]; then
+	echo "No file exists at $FILE_PATH. Cancelling Upload"
+	exit 0
+fi
+
+echo "Uploading ${FILE_PATH} to Test Analytics"
+
+curl -X POST \
+  -H "Authorization: Token token=$ANALYTICS_TOKEN" \
```
mokagio, line 26: Have you thought of calling this `BUILDKITE_ANALYTICS_TOKEN`? 

We're the ones who set the value, and I can see how it might be good to avoid the `BUILDKITE_` prefix to differentiate from the other env vars that come from the agent. On the other hand, all the documentation I interacted with so far, and the source of the Test Analytics Swift library all refer to this as `BUILDKITE_ANALYTICS_TOKEN` 🤷‍♂️

jkmassel: I did, but I actually avoided it intentionally so that if `BUILDKITE_ANALYTICS_TOKEN` was set, it wouldn't just automatically be used. This is because some projects have multiple test suites, and I wanted to avoid any "magic".

That said, we can always adjust this later without it being a breaking change if we want to adjust the approach

bin/upload_buildkite_test_analytics_junit
jkmassel: Ah yep – addressed in https://github.com/Automattic/bash-cache-buildkite-plugin/pull/29/commits/cc0975357ce1036cd31cd76a263ba3e4df55bf1e

bin/upload_buildkite_test_analytics_junit
jkmassel: It's a good comment! Fixed in dd80faf.

Now we just need a linter...

PR #30, opened 2022-10-03
Add a script to install and use a Node JS using nvm

rynaardb:
## What it does

Adds a script to install and use a Node JS using `nvm`.

## How to test

- Ensure that you have a working installation of `nvm`
- Ensure that `nvm` is correctly configured: `command -v nvm`
- Ensure that you have a valid `.nvmrc` file in place, containing the node version to install. For example: `echo "lts/*" > .nvmrc`
- Execute the script (from the same location as `.nvmrc`)  to install the version of Node.js as specified in `.nvmrc` : `sh nvm_install`

## Who

@Automattic/apps-infrastructure

### DIFF
--- a/bin/install_nvm_node_version
+++ b/bin/install_nvm_node_version
@@ -1,9 +1,6 @@
-NODE_VERSION=$1
+#!/bin/bash
 
-if [ -z "$1" ]; then
-	echo "You must specify a Node.js version"
-	exit 1
-fi
+# This script expects a valid .nvmrc at the project's root
 
-nvm install $NODE_VERSION
-nvm use $NODE_VERSION
+nvm install
+nvm use
--- a/bin/install_nvm_node_version
+++ b/bin/install_nvm_node_version
@@ -1,0 +1,9 @@
+NODE_VERSION=$1
+
+if [ -z "$1" ]; then
+	echo "You must specify a Node.js version"
+	exit 1
+fi
+
+nvm install $NODE_VERSION
+nvm use $NODE_VERSION
--- a/bin/nvm_install
+++ b/bin/nvm_install
@@ -1,6 +1,10 @@
 #!/bin/bash
 
 # This script expects a valid .nvmrc at the project's root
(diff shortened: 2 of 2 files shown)


bin/install_nvm_node_version
```diff
@@ -0,0 +1,6 @@
+#!/bin/bash
+
+# This script expects a valid .nvmrc at the project's root
```
jkmassel: If we expect this file to be present, we should test for that and fail if it's not there

jkmassel (changes requested):
Left a note about validation, but one thought about the naming as well – this command doesn't install any particular version. WDYT about just using something like `nvm_install`?

bin/install_nvm_node_version
rynaardb: > Left a note about validation, but one thought about the naming as well – this command doesn't install any particular version. WDYT about just using something like `nvm_install`?

Thanks for the naming suggestion. I agree it is more fitting. I came up with the current naming before switching to the `.nvmrc` file.

rynaardb: Command renamed to `nvm_install` and added a check for `.nvmrc` file in b71b2405e70c2e0b21e24e76d893bd9ba732dd71

bin/nvm_install
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash
```
mokagio: In most recent scripts, we've been adding the `e` flag to fail the script if a command fails and the `u` flag to fail the script if there's an unassigned variable usage. This script defines no variables, but it might in the future.

```suggestion
#!/bin/bash -eu
```

rynaardb: > In most recent scripts, we've been adding the `e` flag to fail the script if a command fails and the `u` flag to fail the script if there's an unassigned variable usage. This script defines no variables, but it might in the future.

Thanks @mokagio , this is indeed a great suggestion!

jkmassel (approved):

PR #31, opened 2022-10-04
Add a script to install node packages using npm

rynaardb:
## What it does

Add a script to install Node JS packages using `npm`.

## How to test

- Ensure that you have a working installation of Node JS
- Execute the script (from the same location as `package.json`)  to install the node packages.

## Who

@Automattic/apps-infrastructure


jkmassel (changes requested):
What's the benefit to using this over just running `npm install`?

rynaardb:
> What's the benefit to using this over just running `npm install`?

I am glad you are asking 😄  I kind of asked the same [question](https://a8c.slack.com/archives/C020Q0Z579V/p1664881059376529), but in a different way on Slack earlier today:

> Is this the approach we want to follow “as a rule” for repetitive commands like these, to add them to this plugin?

From my perspective, I also don't see any real benefit other than maybe having a more verbose command name: `npm install` vs. `install_node_packages`, and maybe some consistency should we ever decide to use something else other than `npm`? If none of this matters then I would say there is probably no real benefit and we can go ahead and close this PR.

jkmassel:
In this repo are some other examples of invoking a package manager – for instance [`install_gems`](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/trunk/bin/install_gems) will transparently take care of caching for us. That would add some benefit for us.

One other suggestion while I'm at it – in this case it seems we're invoking `npm`, so we should probably call that out. We use `install_$foo` for out command name, so it makes sense to have `install_npm_packages` – I could see us also having an `install_yarn_packages` command to invoke that package manager.

rynaardb:
> In this repo are some other examples of invoking a package manager – for instance [`install_gems`](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/trunk/bin/install_gems) will transparently take care of caching for us. That would add some benefit for us.
> 
> One other suggestion while I'm at it – in this case it seems we're invoking `npm`, so we should probably call that out. We use `install_$foo` for out command name, so it makes sense to have `install_npm_packages` – I could see us also having an `install_yarn_packages` command to invoke that package manager.

I've renamed the command to `install_npm_packages` and also added support for caching in 5a4e5b38611f185b91b2228e73ff5fd9b75325a3

bin/install_npm_packages
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -eu
+
+# This script expects a valid package.json at the project's root
+if [ ! -f package.json ]; then
```
jkmassel, line 4: We should instead check for the presence of `package-lock.json` – if that's not present, it's not really safe to run `npm install` (because this could be a yarn repo with a `yarn.lock` file).

It'd be fine to check for `package.json` as well, but the `lock` file is the more important thing.

jkmassel (changes requested):
Have you had a chance to test this on a PR under development? I'd like to see it used in a project before we land this, if we can!

bin/install_npm_packages
rynaardb: Oh yes, that makes a lot of sense! I've added the check for the package-lock.json file in b95b0639794068b7715dea0474893da0bbceea60

jkmassel: I'm not sure this is correct – these should be two separate operations:

1. Ensure package.json is present (ie – is this a node project?)
2. Ensure package-lock.json is present (ie – is this an npm project (as opposed to a yarn project))

rynaardb: Assuming you meant that we should check for **both** conditions, not just one or the other. e6e45de15a28ccd21f2fd8ef25e9ef60ab9b5697 now splits the check into two separate if-statements to ensure **both** files are present.

jkmassel (approved):

PR #32, opened 2022-10-06
Install Bundler version from `Gemfile.lock` + `add_ruby_env_info`

mokagio:
Despite the `gem install bundler` call in `install_gems` we recently see a come back of the `Gem::LoadError: You have already activated bundler...` error from https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16 .

~~I don't know how to fix the issue, but this PR adds info dumps that will hopefully help understand more about its cause.~~

Update: The current best solution is to install the same bundler version as the one tracked in `Gemfile.lock` (original suggestion from #16). The approach has been implemented here and verified [in this Woo demo PR](https://github.com/woocommerce/woocommerce-ios/pull/7820) were the installing the Bundler version from `Gemfile.lock` didn't result in the `Gem::LoadError...` error. See build [here](https://buildkite.com/automattic/woocommerce-ios/builds/8248#0183aeb7-19b5-4ab7-a5be-ec909228ed4c).

image

These debug calls already revealed something interesting: CI has _two_ `default` Bundler versions, which seems like a probable cause for the "you have already activated..." version issue.

I couldn't find an explanation for that `default` in the `gem list --help`, so I [posted a StackOverflow question](https://stackoverflow.com/questions/73967910/how-can-gem-list-show-multiple-default-versions-for-a-gem) about it.

### DIFF
--- a/bin/dump_ruby_env_info
+++ b/bin/dump_ruby_env_info
@@ -1,0 +1,14 @@
+#!/bin/bash -eu
+
+echo "--- :ruby: Ruby environment info dump"
+set -x
+which ruby
+ruby --version
+
+which gem
+gem --version
+
+which bundle
+gem list bundler
+bundle version
+set -x
--- a/bin/dump_ruby_env_info
+++ b/bin/dump_ruby_env_info
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
 
-echo "--- :ruby: Ruby environment info dump"
+echo "--- :ruby: Ruby environment info dump ${1:-}"
 set -x
 which ruby
 ruby --version
--- a/bin/dump_ruby_env_info
+++ b/bin/dump_ruby_env_info
@@ -1,6 +1,13 @@
 #!/bin/bash -eu
 
 echo "--- :ruby: Ruby environment info dump ${1:-}"
+
+if [[ -f .ruby-version ]]; then
+  cat .ruby-version
+else
+  echo 'Could not find .ruby-version in current folder'
+fi
+
 set -x # debug mode: print the commands we call
 
 which ruby
--- a/bin/dump_ruby_env_info
+++ b/bin/dump_ruby_env_info
@@ -1,7 +1,8 @@
 #!/bin/bash -eu
 
 echo "--- :ruby: Ruby environment info dump ${1:-}"
-set -x
+set -x # debug mode: print the commands we call
+
 which ruby
 ruby --version
 
@@ -11,4 +12,5 @@ gem --version
 which bundle
 gem list bundler
 bundle version
-set -x
+
+set +x # exit debug mode
--- a/bin/install_gems
+++ b/bin/install_gems
@@ -9,8 +9,6 @@ CACHEKEY="$BUILDKITE_PIPELINE_SLUG-$ARCHITECTURE-ruby$RUBY_VERSION-$GEMFILE_HASH
 
 restore_cache "$CACHEKEY"
 
-dump_ruby_env_info 'after running restore_cache'
-
 # Install the same Bundler version as the one in the Gemfile.lock.
 # This should prevent runtime issues where Ruby "gets confused" on which Bundler version to use.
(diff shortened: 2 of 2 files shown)


bin/dump_ruby_env_info
```diff
@@ -0,0 +1,23 @@
...
+
+which ruby
+ruby --version
+
+which gem
+gem --version
+
+which bundle
+gem list bundler
+bundle version
```
mokagio, line 21: Feel free to add additional info dump commands if you can think of any 🙇‍♂️

bin/install_gems
```diff
@@ -1,16 +1,22 @@
...
 restore_cache "$CACHEKEY"
 
+dump_ruby_env_info 'after running restore_cache'
+
 # Ensure we're using the latest version of Bundler before running bundle install.
 #
-# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16  for more details on why this is recommended.
-gem install bundler
+# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16 for more details on why this is recommended.
+gem install bundler:"$( (tail -1 | sed 's/ *//g') < Gemfile.lock)"
```
AliSoftware:
```suggestion
GEMFILE_LOCK_BUNDLER_VERSION=$(sed -n -e '/BUNDLED WITH/{n;s/ *//p;}' Gemfile.lock)
gem install bundler -v "$GEMFILE_LOCK_BUNDLER_VERSION"
```

mokagio: Done in [`c9b9d89` (#32)](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/32/commits/c9b9d89de57896e7567a58a666a48a090d6c689f)

bin/install_gems
```diff
@@ -1,16 +1,23 @@
...
-# Ensure we're using the latest version of Bundler before running bundle install.
-#
-# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16  for more details on why this is recommended.
-gem install bundler
+# Install the same Bundler version as the one in the Gemfile.lock.
+# This should prevent runtime issues where Ruby "gets confused" on which Bundler version to use.
+# See https://github.com/Automattic/bash-cache-buildkite-plugin/issues/16 for more details.
+GEMFILE_LOCK_BUNDLER_VERSION=$(sed -n -e '/BUNDLED WITH/{n;s/ *//p;}' Gemfile.lock)
+# The command will fail if the `--version` parameter is empty; we can omit checking for that ourselves.
+gem install bundler --version "$GEMFILE_LOCK_BUNDLER_VERSION"
```
mokagio, line 15: @crazytonyli 's work in 114-gh-Automattic/buildkite-ci should address the issue this change works around at the root, but I think it's still useful to respect the Bundler version set in the Gemfile.lock.

bin/install_gems
```diff
@@ -1,16 +1,23 @@
 #!/bin/bash -eu
 
+dump_ruby_env_info "before running $(basename "$0")"
+
```
AliSoftware: I wonder if we still want to keep those debug dumps in the final `install_gems` action there?

Personally I'd remove both calls (before + after the `gem install bundler --version …`) to reduce verbosity — one can always temporarily add a call to `dump_ruby_env_info` in their pipeline directly if they need to debug Ruby issues on a given build later.

And if you still think it's useful to keep the dump of the whole Ruby env on every builds and steps calling `install_gem`, I'd still remove that initial one to keep only the one after we did the `gem install`

mokagio: My rationale for having them both is to see the before-after state. It was useful to me when trying to understand the issue, to see which was the selected Bundler as the pipeline started.

But given the fix in [114-gh-Automattic/buildkite-ci](https://href.li/?https://github.com/Automattic/buildkite-ci/issues/114) it's probably safe to remove the info logs, even though I don't think they made the logs particularly more verbose, thanks to Buildkite's logs grouping syntax–although I guess one could argue they _would_ make the raw logs more verbose.

AliSoftware (approved):

PR #36, opened 2022-11-18
Add command to run `.strings` file generation

mokagio:
Extracts the script developed in https://github.com/wordpress-mobile/WordPress-iOS/pull/19553 to this plugin so it can be used across multiple repos and iterated upon in isolation.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/19624 for a live demo of the happy behavior and https://github.com/wordpress-mobile/WordPress-iOS/pull/19544 for when it catches errors.

### DIFF
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -22,10 +22,10 @@ LOGS=logs.txt
 set +e
 set -o pipefail
 if [[ $# -eq 0 ]]; then
-  bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
-else
-  bundle exec fastlane "$@" | tee $LOGS
+  # If no argument provided, update $@ with default values
+  set -- generate_strings_file_for_glotpress skip_commit:true
 fi
+bundle exec fastlane "$@" | tee $LOGS
 EXIT_CODE=$?
 set -e
 
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -1,0 +1,48 @@
+#!/bin/bash -eu
+
+# Attempts to generate the frozen `.strings` file to gets sent to GlotPress
+# after a new release code freeze.
+#
+# If the process fails, annotates the build with the errors.
+
+# The strings generation happens via Fastlane, so we need to install the
+# Ruby gems.
+echo "--- :rubygems: Setting up Gems"
+install_gems
+
+# The strings generation also picks up files from our first party libraries,
+# some of which are installed via CocoaPods
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- :sleuth_or_spy: Lint Apple Localized Strings Format"
+# A next step improvement is to move the logs management within the release
+# toolkit
+LOGS=logs.txt
+set +e
+set -o pipefail
+bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+EXIT_CODE=$?
+set -e
+
+if [[ $EXIT_CODE -ne 0 ]]; then
+  # Strings generation finished with errors, extract the errors in an easy-to-find section
+  echo "+++ :x: Strings Generation Failed"
+  ERRORS=errors.txt
+  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`*LocalizedStrings\` calls:" | tee $ERRORS
+  echo '' >> $ERRORS
+  # Print the errors inline.
+  #
+  # Notice the second `sed` call that removes the ANSI escape sequences that
+  # Fastlane uses to color the output.
+  grep -e "\[.*\].*genstrings:" $LOGS \
+    | sed -e 's/\[.*\].*genstrings: error: /- /' \
+    | sed -e $'s/\x1b\[[0-9;]*m//g' \
+    | sort \
+    | uniq \
+    | tee -a $ERRORS
+  # Annotate the build with the errors
+  cat $ERRORS | buildkite-agent annotate --style error --context genstrings
+fi
+
+exit $EXIT_CODE
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -29,14 +29,19 @@ if [[ $EXIT_CODE -ne 0 ]]; then
   # Strings generation finished with errors, extract the errors in an easy-to-find section
   echo "+++ :x: Strings Generation Failed"
   ERRORS=errors.txt
-  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:" | tee $ERRORS
-  echo '' >> $ERRORS
+  printf "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:\n\n" | tee $ERRORS
   # Print the errors inline.
   #
   # Notice the second `sed` call that removes the ANSI escape sequences that
-  # Fastlane uses to color the output.
-  grep -e "\[.*\].*genstrings:" $LOGS \
-    | sed -e 's/\[.*\].*genstrings: error: /- /' \
+  # Fastlane uses to color the output. We need to do this at this point to
+  # account for Fastlane printing the errors multiple times (inline and in the
+  # end of the lane summary) but with different escape sequences. Without it,
+  # we would get multiple hits for the same error.
+  #
+  # See discussion at
+  # https://github.com/wordpress-mobile/WordPress-iOS/pull/19553#discussion_r1017743950
+  cat $LOGS \
+    | sed -ne 's/\[.*\].*genstrings: error: /- /p' \
     | sed -e $'s/\x1b\[[0-9;]*m//g' \
     | sort \
     | uniq \
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -29,7 +29,7 @@ if [[ $EXIT_CODE -ne 0 ]]; then
   # Strings generation finished with errors, extract the errors in an easy-to-find section
   echo "+++ :x: Strings Generation Failed"
   ERRORS=errors.txt
-  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`*LocalizedStrings\` calls:" | tee $ERRORS
+  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:" | tee $ERRORS
   echo '' >> $ERRORS
   # Print the errors inline.
   #
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -28,8 +28,8 @@ set -e
 if [[ $EXIT_CODE -ne 0 ]]; then
   # Strings generation finished with errors, extract the errors in an easy-to-find section
   echo "+++ :x: Strings Generation Failed"
-  ERRORS=errors.txt
-  printf "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:\n\n" | tee $ERRORS
+  ERRORS_FILE=errors.txt
+  printf "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:\n\n" | tee $ERRORS_FILE
   # Print the errors inline.
   #
   # Notice the second `sed` call that removes the ANSI escape sequences that
@@ -45,9 +45,9 @@ if [[ $EXIT_CODE -ne 0 ]]; then
     | sed -e $'s/\x1b\[[0-9;]*m//g' \
     | sort \
     | uniq \
-    | tee -a $ERRORS
+    | tee -a $ERRORS_FILE
   # Annotate the build with the errors
-  cat $ERRORS | buildkite-agent annotate --style error --context genstrings
+  cat $ERRORS_FILE | buildkite-agent annotate --style error --context genstrings
 fi
 
 exit $EXIT_CODE
--- a/bin/lint_localized_strings_format
+++ b/bin/lint_localized_strings_format
@@ -21,7 +21,11 @@ echo "--- :sleuth_or_spy: Lint Apple Localized Strings Format"
 LOGS=logs.txt
 set +e
 set -o pipefail
-bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+if [[ $# -eq 0 ]]; then
+  bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+else
+  bundle exec fastlane "$@" | tee $LOGS
+fi
 EXIT_CODE=$?
 set -e
 


bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+echo "--- :cocoapods: Setting up Pods"
+install_cocoapods
+
+echo "--- :sleuth_or_spy: Lint Apple Localized Strings Format"
+# A next step improvement is to move the logs management within the release
+# toolkit
+LOGS=logs.txt
+set +e
+set -o pipefail
+bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
```
mokagio: I though of checking whether the lane was available, but there's no out of the box command for that.

If the lane is not there, `fastlane` will return `1`. That will make the lane fail and we'll print the error via the code below.

```
$ bundle exec fastlane lane_name_thats_not_available

[...]
[!] Could not find lane 'ios lane_name_thats_not_available'. Available lanes: ios build_for_testing, ios build_jetpack_for_testing, ios test_without_building, ...
```

The build annotation will likely end up being empty. 

This should be an edge case that would occur only when setting up a new project or in the unlikely occurrence that we'll change the `generate_strings_file_for_glotpress` lane name. Having refined error handling for this occurrence seems hitting diminishing returns.

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -eu
+
+# Attempts to generate the frozen `.strings` file to gets sent to GlotPress
+# after a new release code freeze.
```
AliSoftware, line 4:
```suggestion
# Attempts to generate the frozen `.strings` file—which usually gets sent to GlotPress
# after a new release code freeze—to catch potential warnings early on.
```

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+#
+# If the process fails, annotates the build with the errors.
+
+# The strings generation happens via Fastlane, so we need to install the
+# Ruby gems.
```
AliSoftware, line 9: It's me again with my arguments against that 80-columns thing 😄 

I'm not against splitting comments into multiple lines if they are really long (like on line 3-4), but this really starts to look odd to me when we start doing that _just_ for line-wrapping… a single word or two.

I don't have a strict and explicit rule about this though, but instinctively I think that my unconscious reasoning is, if a comment would be close to fit on a single line, and wrapping it at 80 cols would lead to the 2nd line be less than, say 20% of the width of the first line, that _feels_ too unbalanced and odd-looking to me?
So for example, I'm fine with your case on lines 13-14, but find this wrapping on line 8-9 or the one for 19-20 to be a bit unnecessary? 🤷 

Anyway, it's just a hot take, nothing blocking, sorry for the rant 😅

bin/lint_localized_strings_format
AliSoftware: I agree.

Though one easy thing we could consider doing, is to provide this as a default-but-overridable parameter, [like how we do it for commands like `build_and_test_pod`](https://github.com/Automattic/bash-cache-buildkite-plugin/blob/trunk/bin/build_and_test_pod#L27) that uses `test` as the lane name by default but allows to override it by providing it as `$1` (well, `$@` in this case)

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+set -o pipefail
+bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+EXIT_CODE=$?
+set -e
+
+if [[ $EXIT_CODE -ne 0 ]]; then
+  # Strings generation finished with errors, extract the errors in an easy-to-find section
+  echo "+++ :x: Strings Generation Failed"
+  ERRORS=errors.txt
+  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`*LocalizedStrings\` calls:" | tee $ERRORS
```
AliSoftware: Opinion/Nit: What do you think about using `{NS,App}` instead of `*` here in order to potentially be clearer?

```suggestion
  echo "Found errors when trying to run \`genstrings\` to generate the \`.strings\` files from \`{NS,App}LocalizedStrings\` calls:" | tee $ERRORS
```

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+  echo '' >> $ERRORS
+  # Print the errors inline.
+  #
+  # Notice the second `sed` call that removes the ANSI escape sequences that
+  # Fastlane uses to color the output.
+  grep -e "\[.*\].*genstrings:" $LOGS \
+    | sed -e 's/\[.*\].*genstrings: error: /- /' \
```
AliSoftware: Any reason why you didn't apply the fix that you applied on the other PR, namely replace `grep` + first `sed` to use `sed -ne 's/…/- /p "$LOGS"` instead?

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+set +e
+set -o pipefail
+bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+EXIT_CODE=$?
+set -e
+
+if [[ $EXIT_CODE -ne 0 ]]; then
+  # Strings generation finished with errors, extract the errors in an easy-to-find section
+  echo "+++ :x: Strings Generation Failed"
+  ERRORS=errors.txt
```
AliSoftware: 💄  Nit: maybe name this `ERROR_FILE`? 🤷

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,48 @@
...
+  # Notice the second `sed` call that removes the ANSI escape sequences that
+  # Fastlane uses to color the output.
+  grep -e "\[.*\].*genstrings:" $LOGS \
+    | sed -e 's/\[.*\].*genstrings: error: /- /' \
+    | sed -e $'s/\x1b\[[0-9;]*m//g' \
+    | sort \
+    | uniq \
+    | tee -a $ERRORS
+  # Annotate the build with the errors
+  cat $ERRORS | buildkite-agent annotate --style error --context genstrings
```
AliSoftware: Even if, in our case, we're guaranteed that `$ERRORS` doesn't contain any space or special characters and that won't cause any trouble, I'm surprised that `shellcheck` doesn't complain about quotes missing around that `$ERRORS` here 🤷  Or is it smart enough to realize that the value of `ERRORS=` is a literal guaranteed to have no space, and is also never changed/updated by subsequent lines in the scripts, and thus why it doesn't warn? 🤯

bin/lint_localized_strings_format
AliSoftware:
```suggestion
bundle exec fastlane "${@:-generate_strings_file_for_glotpress}" skip_commit:true | tee $LOGS
```

bin/lint_localized_strings_format
mokagio: I think it's the latter.

bin/lint_localized_strings_format
mokagio: @AliSoftware great suggestion, thanks!

I was curious about the `$@` usage, tried it locally, and run in some issues. 

If we use `"${@...}"`, that is wrap the expansion in double quote to avoid re-splitting elements like ShellCheck suggests, and also pass more than one argument to the script invocation, the double quotes will result in Fastlane not finding our lane.

I noticed this when I tried to be smart and move the `skip_commit:true` in the fallback:

```
bundle exec fastlane "${@:-generate_strings_file_for_glotpress skip_commit:true}"
```

image

The approach works fine without the double quotes. However, ShellCheck raises a warning, `SC2068`.


```
bundle exec fastlane ${@:-generate_strings_file_for_glotpress skip_commit:true}
```

image

At this point, I'd like to use the following, what do you think?

```suggestion
# We cannot wrap `${@:-...}` in double quotes because of Fastlane.
# If the caller passes more than one parameter, Fastlane will consider the resulting string with spaces as the name of the lane to run failing the script.
# That's the same if our fallback, which constains more than one word, takes place.
#
# shellcheck disable=2068
bundle exec fastlane ${@:-generate_strings_file_for_glotpress skip_commit:true} | tee $LOGS
```

bin/lint_localized_strings_format
mokagio: Does `genstrings` emit warnings? Or errors only? The manpage has no mentions of "error" and only one of "warnings", yet the output we filter for is `error:`. 🤔 

```suggestion
# Attempts to generate the frozen `.strings` file—which usually gets sent to GlotPress
# after a new release code freeze—to catch potential errors early on.
```

I'm mostly asking because, if it does also print warnings, it might be useful to catch them too (and treat them as errors if possible).

bin/lint_localized_strings_format
mokagio: 🤦 must have copy pasted from an out-of-date branch.

mokagio: Addressed in [`2d04b1f` (#36)](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/36/commits/2d04b1f70e78e7c3e09058a4cfee0583f2b71b8c)

bin/lint_localized_strings_format
mokagio: 👨‍🍳 👌  [`08f8a0a` (#36)](https://github.com/Automattic/bash-cache-buildkite-plugin/pull/36/commits/08f8a0a65f68d5c2aa6e0d7b0abd997a08ed2471)

bin/lint_localized_strings_format
AliSoftware: `"${@:-default}"` will wrap each individual argument from $@ in quotes, which is exactly what we want: if we use that and then  pass `lint_localized_strings_format a b` then `"${@:-default}"` will be substituted with `"a" "b"` (as opposed to `"a b"` like one might think at first). So I still think that's what we need to use.

But when using this syntax, `default` needs to be a single default argument. That means that if you use `"${@:-d1 d2}"` then pass no argument, it will be substituted with the single default argument `"d1 d2"` this time, not `"d1" "d2"`. This is why when you try to include `skip_commit:true` inside it it will not find the lane as it will try to find a lane whose name is `"ios_generate_strings_file_for_glotpress skip_commit:true"` which doesn't exists.

---

I see two solutions for this:

1. Use my initial suggestion. That will always include `skip_commit:true` as part of the parameters passed to the lane. But I don't think that's a problem because (1) I don't see a use case why we'd want `skip_commit:false` instead and (2) if the repo we're running this in uses a different name for that option (that we'd then pass as part of the `$@` parameters for this case), including an unused option to the lane call will not have any side effects and will just be ignored (that lane from that repo will be unlikely to use `options[:skip_commit]` for anything other than deciding if we need to skip the commit step or not)

2. Be more clever and inject `skip_commit:true` into the `$@` parameters list (via `unshift` or something like that?) only `if [[ $# -eq 0 ]]`, before the line that then uses `"$@"`. In other words, don't use the `:-` fallback mechanism in the `"$@"` syntax but instead modify the content of `$@` before if no parameter is passed.

I don't really think solution 2 is worth it though, and I think solution 1 is totally enough for our use case, allowing to overwrite the lane name if needed and pass _additional_ arguments to it in addition to `skip_commit:true` that will always be passed.

bin/lint_localized_strings_format
AliSoftware: That is a very good question… that I don't have the answer to 😅

bin/lint_localized_strings_format
mokagio: > `"${@:-default}"` will wrap each individual argument from `$@` in quotes, which is exactly what we want

I should have tested that scenario better.

> Be more clever and inject `skip_commit:true` into... only `if [[ $# -eq 0 ]]`...

This suggestion made me think of a maybe "less clever" but equally effective alternative:

```sh
if [[ $# -eq 0 ]]; then
  bundle exec fastlane update_appstore_strings skip_commit:true
else
  bundle exec fastlane "$@"
fi
```

The reason I'm keen to have either `$@` or the lane name including all its parameters is because I think if we provide a way for the user to override the default, we should go all in with it.

You made a compelling case for why having `skip_commit:true` as an argument in both scenarios is not an issue and I think it would be a reasonable compromise.

However, I can see myself reading the code in the future and asking why `skip_commit:true` I outside the `${@:-...}` variable substitution. We could easily link to this conversation or leave a comment that summarizes it, but the code above, unless I missed something in the way it will behave, speak for itself.

AliSoftware: I like that compromise suggestion of yours 👍  Let's go with it!

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,57 @@
...
+# A next step improvement is to move the logs management within the release
+# toolkit
+LOGS=logs.txt
+set +e
+set -o pipefail
+if [[ $# -eq 0 ]]; then
+  bundle exec fastlane generate_strings_file_for_glotpress skip_commit:true | tee $LOGS
+else
+  bundle exec fastlane "$@" | tee $LOGS
+fi
```
AliSoftware: Re: your comment about this on commit ae2b014a503ad9cb243a0f9ff5b056c8d7806dd1:

> You might notice the duplicated `| tee $LOGS` call. I considered DRYing the code by assigning `bundle exec fastlane ...` to a variable, e.g. `CMD`, and then do `$CMD | tee $LOGS`, but decided to keep the duplication because it's minimal and keeps the code, IMHO, clearer.

I found another (and more bash-idiomatic) way to avoid repeating the command in each branch: you can use `set -- …` to replace the content of the `$@` arguments list:

```suggestion
if [[ $# -eq 0 ]]; then
  # If no argument provided, update $@ with default values
  set -- generate_strings_file_for_glotpress skip_commit:true
fi
bundle exec fastlane "$@" | tee $LOGS
```

(Or alternatively a more compact version below, though I'm not sure it's worth the loss of readability):

```suggestion
# If no argument provided, update $@ with default values
[[ $# -eq 0 ]] && set -- generate_strings_file_for_glotpress skip_commit:true
bundle exec fastlane "$@" | tee $LOGS
```

mokagio: TIL `set --`. It looks odd at first, but I guess it's just because I'm don't know much idiomatic bash. It certainly achieves the goal of DRYing both `bundle exec fastlane` and `| tee $LOGS` and in a way that's neat.

AliSoftware: TBH I didn't know about it either, I tried `unshift` at first (thinking it was a bash command that existed, as the opposite of `shift`… but no), then I Googled a bit and learned about `set --` in the process… so TIL too 😛

bin/lint_localized_strings_format
```diff
@@ -0,0 +1,57 @@
...
+  # Fastlane uses to color the output. We need to do this at this point to
+  # account for Fastlane printing the errors multiple times (inline and in the
+  # end of the lane summary) but with different escape sequences. Without it,
+  # we would get multiple hits for the same error.
+  #
+  # See discussion at
+  # https://github.com/wordpress-mobile/WordPress-iOS/pull/19553#discussion_r1017743950
+  cat $LOGS \
+    | sed -ne 's/\[.*\].*genstrings: error: /- /p' \
+    | sed -e $'s/\x1b\[[0-9;]*m//g' \
```
AliSoftware, line 49: 💄  Nitpick: `cat $LOGS | sed -ne $CMD` creates two shell processes (one for `cat` then one for `sed`) so it's usually a better practice to use `sed -ne $CMD $LOGS` to use a single shell process (and let `sed` do the file reading directly, instead of `cat` doing the file IO and `sed` doing stdin-piping) instead 🙃 

I'm actually kinda surprised that `shellcheck` didn't automatically suggest you that (is [it does warn for similar cases of unnecessary uses of `cat`](https://www.shellcheck.net/wiki/SC2002)) 🤷 

```suggestion
  sed -ne 's/\[.*\].*genstrings: error: /- /p' $LOGS \
    | sed -e $'s/\x1b\[[0-9;]*m//g' \
```

(PS: If we want to go even further, I think you could also collapse the to `sed` invocations into a single invocation with a unified sed command/script doing the replacements all at once… but I'm not convinced that will lead to something super readable, so I'm ok to keep those separate)

AliSoftware (approved):
Left a small nit about good practices around overuse of `cat`, but not a blocker.
Disabled auto-merge though just in case you want to address it (or not) before merging.

PR #38, opened 2023-01-25
Fix nvm_install issues

rynaardb:
## What it does

Fixes the following issues with the `nvm_install` script:

- Add the missing line to source `nvm`
- Set the correct file permissions on the `nvm_install.sh` script (`0755`) to make it executable
- Remove `nvm use` since the correct versions specified in the `.nvmrc` file will be used once you do a `nvm install`


AliSoftware (commented):
Quoting @jkmassel from the recent conversation in #apps-ci channel:

> So if you `source ~/.nvm/nvm.`sh it’ll automatically install the version of `node` specified by the `.nvmrc` file in the `pwd`
> Running `nvm install` is the wrong way to do it (maybe not officially, but at least in terms of “making this thing actually work right”)

I'm wondering if we are expected to remove `nvm install` and `nvm use` from the end of that `bin/nvm_install` bash-cache script? (especially as the script clearly expects a `.nvmrc` file to be present, and `exit 1` if not)

rynaardb:
> Quoting @jkmassel from the recent conversation in #apps-ci channel:
> 
> > So if you `source ~/.nvm/nvm.`sh it’ll automatically install the version of `node` specified by the `.nvmrc` file in the `pwd`
> > Running `nvm install` is the wrong way to do it (maybe not officially, but at least in terms of “making this thing actually work right”)
> 
> I'm wondering if we are expected to remove `nvm install` and `nvm use` from the end of that `bin/nvm_install` bash-cache script? (especially as the script clearly expects a `.nvmrc` file to be present, and `exit 1` if not)

Upon checking and testing it myself, I'm not convinced the statement made by @jkmassel is entirely accurate. Sourcing the nvm script **does not** automatically **install** the version specified in the `.nvmrc` file.

Here's two examples backing my findings:

This [build](https://buildkite.com/automattic/tumblr-android/builds/11975#0185e4f9-f7ec-4bab-a719-4fe3e0750d3b) failed with `node command not found` because node didn't get installed, even though the `.nvmrc` file was present and nvm got sourced.

This [build](https://buildkite.com/automattic/tumblr-android/builds/11977#0185e4fe-1754-4801-a711-ee7571acf125) succeeded and installed node correctly after adding `nvm install` back.

I also tested it locally and came to the same conclusion.

However, what is true is that you don't need to run `nvm use`, as that gets taken care by the version you specified in the `.nvmrc` file once you do the `nvm install`. That has now been removed in 3a52c7a5da67a9fada87b9d04a509dcfcebd2b4f.

bin/nvm_install
```diff
@@ -1,10 +1,12 @@
...
+
 # This script expects a valid .nvmrc at the project's root
 if [ ! -f .nvmrc ]; then
     echo "No valid .nvmrc file found"
     exit 1
 fi
 
-nvm install
-nvm use
+nvm install
```
AliSoftware: Please add a newline at end of file to avoid issues with `cat` and other POSIX commands (also most IDEs will add one automatically next time we open the file in an editor, and will thus mark the file as modified next time it gets opened because of this CRLF at EOF being added by them automatically for POSIX conventions)

AliSoftware (approved):
I'd love to have a final approval by Jeremy—who seems more versed in Nodejs / nvm than I am—for confirmation; but given your tests and findings I'm going to approve to at least unblock you

jkmassel:
> Upon checking and testing it myself, I'm not convinced the statement made by @jkmassel is entirely accurate. Sourcing the nvm script **does not** automatically **install** the version specified in the `.nvmrc` file.

Sorry, that's correct – [_installing_ NVM will auto-install the node version](https://github.com/nvm-sh/nvm/blob/552db40622bb7a82d9c6d67d2d6bcf3694b47e30/install.sh#L455), my mistake.

PR #41, opened 2023-03-01
Add Makefile

jkmassel:
Slightly simplify tasks and aligns with our other repos


AliSoftware (approved):
image

AliSoftware:
Mmmh actually I see there's also a `test.sh` file that contains `docker-compose run …` commands. Might be time to get rid of it to promote using the Makefile instead.

PR #42, opened 2023-03-01
Add Makefile round 2

jkmassel:
Replaces the `test.sh` file with `Make`, and also the CI invocations


Makefile
```diff
@@ -1,9 +1,12 @@
 .DEFAULT_GOAL := default
 
 lint:
-	docker-compose run --rm lint
+	docker run -it --rm -v `pwd`:/plugin:ro buildkite/plugin-tester
```
AliSoftware: Nit: (avoids spawning a shell and process just to get the working dir…)

```suggestion
	docker run -it --rm -v "$PWD":/plugin:ro buildkite/plugin-tester
```

Makefile
```diff
@@ -1,9 +1,12 @@
...
 
 test:
 	docker run -it --rm -v `pwd`:/plugin:ro buildkite/plugin-tester
```
AliSoftware: Ditto
```suggestion
	docker run -it --rm -v "$PWD":/plugin:ro buildkite/plugin-tester
```

Makefile
```diff
@@ -1,9 +1,12 @@
...
 
-default: lint test
+shellcheck:
+	docker run -it --rm -v `pwd`:/app -w /app koalaman/shellcheck hooks/** bin/** --exclude=SC1071
```
AliSoftware: Ditto
```suggestion
	docker run -it --rm -v "$PWD":/app -w /app koalaman/shellcheck hooks/** bin/** --exclude=SC1071
```

AliSoftware (approved):
Left a nit to simplify the `-v` params.
Otherwise 👍

Makefile
jkmassel: Addressed in 2f8e498f1810b0fc0058cc0e37fd619c0cb9ff23 – had to use the `${}` syntax to make it work

Makefile
jkmassel: Interestingly, we can't get rid of `docker-compose` here, because I can't figure out how the heck the local plugin parser works 😓

tests/test-that-all-files-are-executable.rb
```diff
@@ -0,0 +1,16 @@
+#!/usr/bin/env ruby
+
+require 'rspec'
+
+RSpec.configure do |config|
+  config.formatter = :documentation
+end
+
+context "All Commands Should Be Executable" do
```
AliSoftware: Might be worth adding a `# comment` to explain why this test was not done as part of `environment.bats` (and maybe one in `tests/environment.bats` as well to point to this file) and why this had to be moved to Ruby.

Otherwise I can totally see one of our future selves being tempted to move it back to `environment.bats` in order to mutualize all the Unit Tests in the same place, without remembering about the quirk we discovered about it.

jkmassel: Yep good call – added in 27d6c20f119c53574960b43353b59c162b0eb286

tests/test-that-all-files-are-executable.rb
```diff
@@ -7,6 +7,10 @@
   config.formatter = :documentation
 end
 
+# You might be thinking "why is this in Ruby instead of `environment.bats`?" Good question – for some reason,
+# it seems that running `[ -x ]` under `bats` returns invalid results, and this was more reliable.
```
AliSoftware: Well technically running `[ -x ]` under `bats` works as expected… when run in the context of our CI. But it fails to detect non-executable files only when running that command under `docker` locally from our Macs. 🤷 

Anyway, 👍  at leaving breadcrumbs 🙂

AliSoftware: Btw, cc @rynaardb for information, as we discussed that bug in Slack DMs last week

jkmassel: 52029ca1041c6e47137eff5827b0b3181734c4e4 😉

rynaardb: Yeah, my conclusion was also that there might be something weird going on with macOS's version of Docker.

PR #43, opened 2023-03-01
Add SPM Caching Action

spencertransier:
### Summary

This PR adds a caching action for Swift Package Manager to the CI toolkit. Some notes:
- It works for both workspace-based repos like WCiOS, DOiOS/Mac, etc. along with non-Xcode repos like `hostmgr` or Swift packages if they have SPM dependencies. The action automatically detects what kind of repo it is based on where the `Package.resolved` file is located. 
- I used the existing `install_swiftpm_dependencies` action because I didn't see it being used elsewhere. If that's incorrect (@jkmassel & @AliSoftware looks like you both worked on that action in the past), please let me know and I can move SPM caching into a separate action. 
- This is my first time diving into bash scripting. If there are conventions I'm not following or better ways of doing something, please let me know! 

### How to Test
You can see this CI build of Day One iOS/Mac that's using the latest caching code: 

Example Builds: 
- DOiOS/Mac: https://buildkite.com/automattic/day-one-apple/builds/7705
- WCiOS: https://buildkite.com/automattic/woocommerce-ios/builds/11347
- hostmgr: https://buildkite.com/automattic/hostmgr/builds/295

There is now a "Restoring SPM Cache" step: 

Screenshot 2023-03-01 at 3 39 13 PM

To try regenerating the cache, you can delete any of the `day-one-spm-cache...` files from the `a8c-ci-cache` and re-run the build to verify that the cache rebuilds.


spencertransier:
Question about the Shellcheck error I'm getting: 

> In bin/save_cache line 58:
[2023-03-01T23:35:12Z] 		tar $TAR_OPTIONS -czf "$CACHE_KEY" -C "$CACHE_FILE" .
[2023-03-01T23:35:12Z]                     ^----------^ SC2086 (info): Double quote to prevent globbing and word splitting.

It's referring to `$TAR_OPTIONS` not being surrounded in quotes. The issue is that if I add the quotes, the `tar` options that I'm passing to `save_cache` from `install_swiftpm_dependencies` don't actually get used by `tar` when archiving the cache. In this case, I'm passing folders that should be excluded when creating the cache archive. 



**Without** the quotes around `$TAR_OPTIONS` in `save_cache`, the folders are excluded from the archive as expected. **With** the quotes, the `--exclude` options are ignored and the folders are not excluded from the archive. 

I'm just not sure why adding the quotes is doing that. Any suggestions about what could be changed to allow for the quotes to be added? I know this is just an `info` issue with Shellcheck. I assume we have Shellcheck fail for `info` issues for a reason, so I didn't want to ignore this one even though it's not even a warning.

AliSoftware:
@spencertransier  Adding the quotes around `"$TAR_OPTIONS"` means that if that variable contains eg `--exclude Foo` then it will consider `"--exclude foo"` as a single argument (one which contains spaces) as opposed to two.

On the other hand, not quoting the `$TAR_OPTIONS` means that if it contains `--exclude foo` it'll see it as two individual arguments like you probably want… but that also means that if the folder you want to exclude contains a space, eg `--exclude my project` it will see it as three distinct arguments (`--exclude`, `my` and `project`)

The best solution for your case is probably to use the bash array syntax to build your list or tar options (whose syntax I always forget but a quick search on the net should help)

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,65 @@
...
+	echo "BUILD_TYPE is not set"
+fi
+
+# `checkouts` can be excluded because the system can quickly generate them 
+# instead of needing to download them in the cache each time.
+# 
+# `artifacts` should be excluded because it causes issues when downloading 
+# certain packages to have the artifacts already present after extracting 
+# cache
+TAR_OPTIONS="--exclude=./checkouts --exclude=./artifacts"
```
crazytonyli: Rather than fighting with bash syntax regarding using quote or not, I'm wondering if it'd be simpler to just delete these unwanted files from the SPM_CACHE_LOCATION?

crazytonyli: Or, copy the "SPM_CACHE_LOCATION", delete the unwanted files from the new copy, call `save_cache` on the new copy, and delete the new copy afterwards.

spencertransier: @AliSoftware Thank you for the explanation! The array solution makes a lot of sense after the behavior I saw while debugging this.

@crazytonyli Good idea! I think I'm going to go with deleting those directories before saving the cache. It removes the complexity of passing the tar options along with simplifying the changes to `save_cache`.

bin/save_cache
```diff
@@ -48,7 +48,17 @@ if ! aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/
 	echo "No existing cache entry for $CACHE_KEY – storing in cache"
 
 	echo "	Compressing"
-	tar -czf "$CACHE_KEY" "$CACHE_FILE"
+	TAR_OPTIONS=${4-foo}
```
AliSoftware: Why `-foo` here?

spencertransier: Oops, that was left over from some things I was trying out 🙂

spencertransier: Fixed with https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/43/commits/8a6377fc20ff835c4b2d4f9272d761548c110abc

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,67 @@
...
+  PACKAGE_RESOLVED_LOCATION=$CUSTOM_PACKAGE_RESOLVED_PATH
+elif  [[ -f ${WORKSPACE_PACKAGE_RESOLVED_PATH[0]} ]]; then
+  BUILD_TYPE="XCODEBUILD"
+  PACKAGE_RESOLVED_LOCATION=${WORKSPACE_PACKAGE_RESOLVED_PATH[0]}
+elif  [[ -f $ROOT_PACKAGE_RESOLVED_PATH ]]; then
+  BUILD_TYPE="SWIFT"
+  PACKAGE_RESOLVED_LOCATION=$ROOT_PACKAGE_RESOLVED_PATH
+else
+  echo "Unable to find Package.resolved"
+fi
```
AliSoftware: I'm surprised that `shellcheck` didn't make you add quotes around all of those to prevent issues if paths have spaces? 🤔 

```suggestion
if [[ -n "$CUSTOM_PACKAGE_RESOLVED_PATH" ]]; then
  PACKAGE_RESOLVED_LOCATION="$CUSTOM_PACKAGE_RESOLVED_PATH"
elif  [[ -f "${WORKSPACE_PACKAGE_RESOLVED_PATH[0]}" ]]; then
  BUILD_TYPE="XCODEBUILD"
  PACKAGE_RESOLVED_LOCATION="${WORKSPACE_PACKAGE_RESOLVED_PATH[0]}"
elif  [[ -f "$ROOT_PACKAGE_RESOLVED_PATH" ]]; then
  BUILD_TYPE="SWIFT"
  PACKAGE_RESOLVED_LOCATION="$ROOT_PACKAGE_RESOLVED_PATH"
else
  echo "Unable to find Package.resolved"
fi
```

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,67 @@
...
+
+PACKAGE_RESOLVED_HASH=$(hash_file "$PACKAGE_RESOLVED_LOCATION")
+CACHE_KEY="$BUILDKITE_PIPELINE_SLUG-spm-cache-$PACKAGE_RESOLVED_HASH"
```
AliSoftware: Nit for readability and better separation of variable name vs string literal parts (making sure part of the literal string isn't interpreted as the name of the variable preceding it):
```suggestion
CACHE_KEY="${BUILDKITE_PIPELINE_SLUG}-spm-cache-${PACKAGE_RESOLVED_HASH}"
```

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,67 @@
 #!/bin/bash -eu
 
+CUSTOM_PACKAGE_RESOLVED_PATH=${1-}
+BUILD_TYPE=${2-}
+SPM_CACHE_LOCATION="$HOME/spm"
```
AliSoftware: There exists a `$HOME/.swiftpm/cache` folder on my Mac that was probably created by SPM itself at some point.

I wonder if we shouldn't use that folder (for conventions) as opposed to `~/spm` here?

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,67 @@
...
 sudo defaults write com.apple.dt.Xcode IDEPackageSupportUseBuiltinSCM YES
 
 # Trust all GitHub.com and BitBucket.org keys – this allows checking out dependencies via SSH
 add_host_to_ssh_known_hosts bitbucket.org
 add_host_to_ssh_known_hosts github.com
 
-xcodebuild -resolvePackageDependencies
+# Based on the project type, resolve the packages using the correct build type
+if [[ "$BUILD_TYPE" == "XCODEBUILD" ]]; then
+  echo "	Resolving packages with \`xcodebuild\`"
```
AliSoftware, line 42: What do you think about [using the `~~~` "section" syntax that Buildkite supports](https://buildkite.com/docs/pipelines/managing-log-output) to control log output here?
(And probably same on the lines below (L46, L60, L64) too?)

On the other hand, that will create a new section while maybe we want all of those sub-messages to be grouped under the same parent… 🤷

bin/save_cache
```diff
@@ -48,7 +48,17 @@ if ! aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/
 	echo "No existing cache entry for $CACHE_KEY – storing in cache"
 
 	echo "	Compressing"
-	tar -czf "$CACHE_KEY" "$CACHE_FILE"
+	TAR_CONFIG=${4-}
+	if [[ "$TAR_CONFIG" == '--use_alternate_tar_config' ]]; then
```
AliSoftware: Nit but I'm not a fan of the name of this custom flag… what about `--use_relative_path_in_tar` or something like that instead?

---

Also, I'm wondering if it wouldn't make sense to adopt that `-C` behavior… in all circumstances after all (seems a nicer solution than including `/User/builder` in the cache path and risking the un-tar-ing to fail if used on a different builder which might not have a `/User/builder` folder (all our Mac VMs currently do, but we shouldn't need to make that assumption, and also nothing says the cache would be restored on a Mac VM vs an Android AMI or Docker container or whatnot, so…).

That one might be a broader investigation to verify that changing that behavior that `save_cache` was using so far (and you kept in the `else`) to instead use `-C "$CACHE_FILE" .` in all cases would not break existing setups depending on how they use `restore_cache` on their end; so could be saved for a subsequent PR if you're not sure about that. But would be a nice simplification if we can confirm it and go in that direction.

bin/install_swiftpm_dependencies
spencertransier: I thought about that and decided to keep them grouped under the "Restoring SPM Packages" parent for now. I feel like they aren't important enough to surface higher, but we can always easily change that if we decide differently in the future 🙂

bin/install_swiftpm_dependencies
spencertransier: Ah, good point! Jeremy and I had discussed using the cache that's within DerivedData, but I wasn't able to make that work because there's not a consistent path name within DerivedData. But it does work great with making the cache location in the script point to `$HOME/Library/Caches/org.swift.swiftpm`. And that has the added benefit of client repos not needing to specify a specific SPM cache folder in the Fastfile. Package resolution will pull from that `org.swift.swiftpm` folder automatically when resolving the packages. 

Added in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/43/commits/44785e46a581287c90a5f3d2f021440a33cec2bf

bin/save_cache
spencertransier: Agreed, I like `--use_relative_path_in_tar` better, too. Changed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/43/commits/b919255f6778f69af21855091369fc4caa96dad6

spencertransier: > I'm wondering if it wouldn't make sense to adopt that -C behavior in all circumstances. 

I agree that would be nice to use the same behavior for all cases if possible. Because I saw different behavior between the existing `tar` configuration versus the one I added in this PR, I had assumed the projects that use `save_cache` would need to be updated and cause a breaking change. I haven't tested that out yet, but I think it'd be outside the scope for this project. I'll create a separate issue for it.

jkmassel (approved):

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,63 @@
...
+# 
+# `artifacts` should be removed because it causes issues when downloading 
+# certain packages to have the artifacts already present after extracting 
+# cache
+echo "	Cleaning up cache files before saving cache"
+rm -rf "$SPM_CACHE_LOCATION/checkouts" "$SPM_CACHE_LOCATION/artifacts"
+
+# If this is the first time we've seen this particular cache key, save it for the future
+echo "	Saving SPM Cache"
+save_cache "$SPM_CACHE_LOCATION" "$CACHE_KEY" false --use_relative_path_in_tar
```
crazytonyli, line 63: Not a blocker. This sequence of argument feels prone to errors, it's probably worth looking into parsing the arguments and options, using `getopts` for example. Statement as below would look much nicer 😸 

```
save_cache --cache-key "$CACHE_KEY" --overwrite --use_relative_path_in_tar "$SPM_CACHE_LOCATION"
```

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,63 @@
...
+if [[ -n "$CUSTOM_PACKAGE_RESOLVED_PATH" ]]; then
+  PACKAGE_RESOLVED_LOCATION="$CUSTOM_PACKAGE_RESOLVED_PATH"
+elif  [[ -f "${WORKSPACE_PACKAGE_RESOLVED_PATH[0]}" ]]; then
+  BUILD_TYPE="XCODEBUILD"
+  PACKAGE_RESOLVED_LOCATION="${WORKSPACE_PACKAGE_RESOLVED_PATH[0]}"
+elif  [[ -f "$ROOT_PACKAGE_RESOLVED_PATH" ]]; then
+  BUILD_TYPE="SWIFT"
+  PACKAGE_RESOLVED_LOCATION="$ROOT_PACKAGE_RESOLVED_PATH"
+else
+  echo "Unable to find Package.resolved"
```
crazytonyli, line 20: Should we exit with error here?

bin/install_swiftpm_dependencies
```diff
@@ -1,9 +1,63 @@
...
+CACHE_KEY="${BUILDKITE_PIPELINE_SLUG}-spm-cache-${PACKAGE_RESOLVED_HASH}"
+
+# Restore SPM cache if it's available
+mkdir -p "$SPM_CACHE_LOCATION"
+cd "$SPM_CACHE_LOCATION"
+restore_cache "$CACHE_KEY"
+cd -
+
+# This will let Xcode use the system SSH config for downloading packages
 sudo defaults write com.apple.dt.Xcode IDEPackageSupportUseBuiltinSCM YES
```
crazytonyli, line 34: Not part of this PR. Is this preference same as `xcodebuild -scmProvider system`?

crazytonyli (changes requested):

bin/install_swiftpm_dependencies
spencertransier: Yep 🙂 Resolved in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/commit/9d49421aeab284db1e3672b83bb19ec4eadad162

crazytonyli (approved):

bin/install_swiftpm_dependencies
spencertransier: @crazytonyli I _think_ so but I'm not sure. I will look into this and see about adding in the next update of the SPM caching action.

bin/install_swiftpm_dependencies
spencertransier: Thank you for the suggestion! I am going to land this PR and work on improving this in my next iteration of the SPM caching and `save_cache` improvements.

mokagio: See https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/54

PR #44, opened 2023-03-02
Try fixing local vs CI discrepancy on remaining bats test

AliSoftware:
## Why?

I noticed that the remaining test that we have in `environment.bats` still appears to be affected by the same issue about "properly goes red on CI but stays green when run locally" (the one we solved on the previous PR by moving the "executable bit" test to Ruby / rspec).

In other words, running `make buildkite-plugin-test` will always succeed locally even if some commands are supposed have `which` respond in failure, while it will fail as expected on CI.

## Failed Attempt at Fixing

At that point that made me wonder if we should keep the `plugin-tester` thing vs moving everything to Ruby tests. But then when attempting to do that, I noticed that even when running the test in Ruby (`expect(system(%Q(/bin/bash -c 'source "$PWD/hooks/environment" >/dev/null; command -v "#{cmd}"'))).to be true`) always passes locally, so this time the problem isn't solved by migrating to Ruby.

But after running `docker run -it …` to test the commands interactively in the container (see spoiler blocks below), I ended up confirming that running `which` (as well as `command -v`) in those containers locally (from my Mac) **always returns the path** to files that can be found in `$PATH`… **even when those are confirmed to not have the executable bit**. While **those same commands fail on CI for the same case**.

<details><summary>🕵️  Apparently `which` (and `command -v`) don't care about the executable bit when run on Docker Mac locally (while they do on CI)</summary>

```
$ docker run -it --rm -v "/Users/olivierhalligon/Documents/Dev/tools/a8c-ci-toolkit-buildkite-plugin":/plugin -w /plugin ruby:2.7.4 /bin/bash
root@b25828ad31a5:/plugin# which save_cache
root@b25828ad31a5:/plugin# which cache_repo
root@b25828ad31a5:/plugin# source hooks/environment 
~~~ :file_cabinet: Loaded Caching Plugin
root@b25828ad31a5:/plugin# which save_cache
./bin/save_cache
root@b25828ad31a5:/plugin# which cache_repo
./bin/cache_repo
root@b25828ad31a5:/plugin# ls -l $(which cache_repo)
-rw-r--r-- 1 root root 496 Mar  2 11:11 ./bin/cache_repo
```
</details>
<details><summary>🕵️ The same is true when I try to run it in the container used to run <code>bats</code></summary>

```
$ docker run -it --rm -v "${PWD}":/plugin buildkite/plugin-tester /bin/bash
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
bash-5.1# which cache_repo
bash-5.1# which save_cache
bash-5.1# source hooks/environment 
~~~ :file_cabinet: Loaded Caching Plugin
bash-5.1# which save_cache
./bin/save_cache
bash-5.1# which cache_repo
./bin/cache_repo
bash-5.1# ls -l $(which cache_repo)
-rw-r--r--    1 root     root           496 Mar  2 11:11 ./bin/cache_repo
```
</details>

➡️  So it's really the `which` / `command -v` implementations that differs when running that docker container from a Mac vs from CI… 😒 

## What's left in this PR

I ended up bailing on this for now and I guess we'll have to consider living with the fact that those "check that all commands are exposed in $PATH" tests will never to be trusted when running `make test` locally and will never behave the same on local vs CI. So in the end I didn't change anything in that part…

But, while I was working on this, I made some tiny improvements on the Ruby test about executable bit to make it more Ruby-esque… so I ended up commiting that change anyway, and that's all what that PRs ends up being after all.

## To Test

 - Confirm that [the CI build for intentional failure introduced in 24dd1da293c8d2b26bce73af038fc38485600732 failed as expected](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/374#0186a202-1fc7-42d3-9f15-0a7fefd0d64e)
 - Introduce a failure by running `chmod -x bin/cache_repo`, then run `make test` locally and validate it goes red
 - Revert the failure (`git restore bin/cache_repo` and validate that running `make test` locally passes and goes back green


AliSoftware:
🚨  Interestingly, https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/374#0186a202-1fc7-42d3-9f15-0a7fefd0d64e fails when running the `environment.bats` test, while running the same test locally… goes green.

So I guess the glitch on those tests failing on CI but passing locally… was not just on the test about executable bit after all, but also about the one about `which` and if the command was exposed.

Maybe we should get rid of `environment.bats` after all and move everything to Ruby? Though not sure how to replicate the result of `source hooks/environment` from a Ruby test…? wdyt?

AliSoftware:
Ok, I bail, apparently `which` (and `command -v`) don't care about the executable bit when run on Docker Mac locally. As you can see from the run below, running `which cache_repo` in the ruby container returns a path… even if `cache_repo` does not have the executable bit.

So that means that even implementing that test (about commands being exposed in `$PATH`) in Ruby instead of bats wouldn't solve the issue, we'd still have a difference between local and CI 😒 

```
$ docker run -it --rm -v "/Users/olivierhalligon/Documents/Dev/tools/a8c-ci-toolkit-buildkite-plugin":/plugin -w /plugin ruby:2.7.4 /bin/bash
root@b25828ad31a5:/plugin# which save_cache
root@b25828ad31a5:/plugin# which cache_repo
root@b25828ad31a5:/plugin# source hooks/environment 
~~~ :file_cabinet: Loaded Caching Plugin
root@b25828ad31a5:/plugin# which save_cache
./bin/save_cache
root@b25828ad31a5:/plugin# which cache_repo
./bin/cache_repo
root@b25828ad31a5:/plugin# ls -l $(which cache_repo)
-rw-r--r-- 1 root root 496 Mar  2 11:11 ./bin/cache_repo
```

The same is true when I try to run it in the container in which `bats` is currently executed, of course:
```
$ docker run -it --rm -v "${PWD}":/plugin buildkite/plugin-tester /bin/bash
WARNING: The requested image's platform (linux/amd64) does not match the detected host platform (linux/arm64/v8) and no specific platform was requested
bash-5.1# which cache_repo
bash-5.1# which save_cache
bash-5.1# source hooks/environment 
~~~ :file_cabinet: Loaded Caching Plugin
bash-5.1# which save_cache
./bin/save_cache
bash-5.1# which cache_repo
./bin/cache_repo
bash-5.1# ls -l $(which cache_repo)
-rw-r--r--    1 root     root           496 Mar  2 11:11 ./bin/cache_repo
```

So that still doesn't explain neither why there's such a different between local and CI, nor how we could solve it for the "check it's exposed in `$PATH`" test that is still implemented in `bats`. I guess we'll have to live with it and there's no way to solve that discrepancy case for the "check command is exposed" test then?

PR #46, opened 2023-03-09
Fix `annotate_test_failures` for flaky retries which ultimately succeed

AliSoftware:
## What?

Fixes `annotate_test_failures` command, to ignore cases of flaky retries which ultimately succeeded.

This bug was the cause for some CI builds to still get an annotation mentioning test failures… even when the corresponding CI step ended up green (See [this example build](https://buildkite.com/automattic/wordpress-ios/builds/12940)):

test-failure-annotation-but-step-green

_(cc @mokagio & @crazytonyli  as I talked about this bug very recently with both of you, in P2s and PR comments respectively)_

## Why?

When a flaky test failed but Xcode is configured to auto-retry tests multiple times on failure, all the intermediate failures are recorded in the `report.junit` alongside the final success (if any). For example this is an extract of such `.junit` report:

```xml
    <testcase classname='WordPressTest.TenorDataSourceTests' name='testDataSourceReceivesRequestedCount'>
      <failure message='Asynchronous wait failed: Exceeded timeout of 2 seconds, with unfulfilled expectations: &quot;Waiting&quot;.'>&lt;unknown&gt;:0</failure>
    </testcase>
    <testcase classname='WordPressTest.TenorDataSourceTests' name='testDataSourceReceivesRequestedCount'>
      <failure message='XCTAssertEqual failed: (&quot;Optional(0)&quot;) is not equal to (&quot;Optional(3)&quot;)'>WordPress/WordPressTest/MediaPicker/Tenor/TenorDataSouceTests.swift:37</failure>
    </testcase>
    <testcase classname='WordPressTest.TenorDataSourceTests' name='testDataSourceReceivesRequestedCount' time='1.042'/>
```

The previous version of our script were not smart enough to detect that, and simply extracted all `testcase` nodes that had a `failure` subnode, and built the annotation with the list of found failures from that. Which is why it erroneously included the flaky failures

## How?

To solve this, during the iteration on the `testcase[failure]` node candidates, we are now finding all the sibling nodes to that `testcase` which happens to have the same `classname` and `name` attributes, i.e. nodes reporting all the assertion failures for the same test. This list will thus include the current candidate being iterated on, but also potentially all other retries of the same test.

Once we got that list of nodes, we check if the last one of those nodes ends up being a failure or success XML node:
 - If it ended up in success, that means this was a flaky test that ultimately ended up being marked green by Xcode and was considered passing (even if it failed first and was retried)
 - While if that last instance ended up with a failure, that means the test never succeeded even during potential subsequent retries, so that's a true failure that needs reporting

### WPiOS Demo build with ❌ failed + ⚠️ flaky tests

 - Check [this Buildkite build](https://buildkite.com/automattic/wordpress-ios/builds/13022) — which corresponds to [the build of the test branch I created in WPiOS specifically to create flaky tests and point to this branch of the plugin](https://github.com/wordpress-mobile/WordPress-iOS/compare/bcfad40714)
 - Confirm that the build is red / marked as failed
 - Confirm that both the error and warning annotations are present under the "All Jobs" tab, and that both look ok
    - Note: if Buildkite shows you the "Issues" tab instead of "All Jobs" tab by default, that will only show you the error annotation; be sure to check the "All Jobs" tab to see the warning annotation too.

image

### WPiOS Demo build with ⚠️ only flaky tests

 - Check [this Buildkite build](https://buildkite.com/automattic/wordpress-ios/builds/13026) — which [disables all failing tests and only keep flaky ones](https://github.com/wordpress-mobile/WordPress-iOS/commit/bdc248c9fe854804392b9323c43abd71e9087d70)
 - Confirm that the build is green / passed
 - Confirm that there is no error annotation but only a warning annotation, and that it looks ok

image

### WPiOS Demo build with ❌ only failed tests

 - Check [this Buildkite build](https://buildkite.com/automattic/wordpress-ios/builds/13024) — which [disables all flaky tests and only keep failing ones](https://github.com/wordpress-mobile/WordPress-iOS/commit/2dbc20cc30489f66ed992695913faed9184922f2)
 - Confirm that the build is red / marked as failed
 - Confirm that there is no warning annotation but only an error annotation, and that it looks ok

image

### DIFF
--- a/bin/annotate_test_failures
+++ b/bin/annotate_test_failures
@@ -17,28 +17,46 @@ end
 file = File.open(junit_path)
 xmldoc = REXML::Document.new(file)
 
+# Here's how report.junit XML files look like for failure vs success:
+#
+# <testcase classname='WordPressTest.MediaServiceTests' name='testDeletingLocalMediaThatDoesntExistInCoreData'>
+#   <failure message='Asynchronous wait failed: Exceeded timeout of 0.1 seconds, with unfulfilled expectations: &quot;The delete call succeeds even if the media object isn&apos;t saved.&quot;.'>&lt;unknown&gt;:0</failure>
+# </testcase>
+# <testcase classname='WordPressTest.MediaServiceTests' name='testDeletingLocalMediaThatDoesntExistInCoreData' time='0.145'/>
+#
 failures = []
 REXML::XPath.each(xmldoc, '//testcase[failure]') do |testcase|
-  failure = testcase.elements['failure']
-  failures.append <<~ENTRY
-  <details><summary><tt>#{testcase['name']}</tt> in <tt>#{testcase['classname']}</tt></summary>
-  
-  #{failure['message']}
+  test_class = testcase['classname']
+  test_name = testcase['name']
+  failure_node = testcase.elements['failure']
+  failure_message = failure_node['message']
 
-  ```
-  #{failure.text}
-  ```
-  </details>
-  ENTRY
+  # Find all nodes that are about the same test, to detect false positives
+  # (in case there were multiple retries of the same flaky test with the last one ultimately passing)
+  retries = testcase.parent.get_elements("testcase[@classname='#{testcase['classname']}' and @name='#{testcase['name']}']")
+  if retries.last.elements['failure']
+    failures.append <<~ENTRY
+      <details><summary><tt>#{test_name}</tt> in <tt>#{test_class}</tt></summary>
+      #{failure_message}
+  
+      ```
+      #{failure_node.text}
+      ```
+      </details>
+    ENTRY
+  else
+    puts "- Note: Flaky test \`#{test_class}`.`#{test_name}`: #{failure_message}"
+  end
 end
 
 if failures.empty?
-  puts "No test failure found. Removing any previous `#{title}` Buildkite annotation if any."
+  puts "No test failure found (after eliminating intermediate failed retries). Removing any previous `#{title}` Buildkite annotation if any."
   system('buildkite-agent', 'annotation', 'remove', '--context', title)
   exit 0 # We don't want to fail if the `annotation remove` command failed because there was no previous annotation
 else
-  puts "Found #{failures.count} test failure(s). Reporting them as a `#{title}` Buildkite annotation."
+  puts "Ultimately found #{failures.count} real test failure(s). Reporting them as a `#{title}` Buildkite annotation."
   markdown = "\#\#\#\# #{title}: #{failures.length} failure(s)\n\n" + failures.join("\n")
+  puts markdown
   system('buildkite-agent', 'annotate', markdown, '--style', 'error', '--context', title)
   exit $?.exitstatus
 end
--- a/bin/annotate_test_failures
+++ b/bin/annotate_test_failures
@@ -5,6 +5,10 @@
 #
 require 'rexml/document'
 
+###################
+# Parse arguments
+###################
+
 junit_path = ARGV.first || File.join('build', 'results', 'report.junit')
 title = ENV.fetch('BUILDKITE_LABEL', 'Tests')
 
@@ -14,8 +18,9 @@ unless File.exist?(junit_path)
   exit 0
 end
 
-file = File.open(junit_path)
-xmldoc = REXML::Document.new(file)
+###################
+# Helper methods
+###################
 
 # Here's how report.junit XML files look like for failure vs success:
 #
@@ -24,39 +29,91 @@ xmldoc = REXML::Document.new(file)
 # </testcase>
 # <testcase classname='WordPressTest.MediaServiceTests' name='testDeletingLocalMediaThatDoesntExistInCoreData' time='0.145'/>
 #
-failures = []
-REXML::XPath.each(xmldoc, '//testcase[failure]') do |testcase|
-  test_class = testcase['classname']
-  test_name = testcase['name']
-  failure_node = testcase.elements['failure']
-  failure_message = failure_node['message']
+class TestFailure
+  attr_reader :classname, :name, :message, :details
+  attr_accessor :count
 
-  # Find all nodes that are about the same test, to detect false positives
-  # (in case there were multiple retries of the same flaky test with the last one ultimately passing)
-  retries = testcase.parent.get_elements("testcase[@classname='#{testcase['classname']}' and @name='#{testcase['name']}']")
-  if retries.last.elements['failure']
-    failures.append <<~ENTRY
-      <details><summary><tt>#{test_name}</tt> in <tt>#{test_class}</tt></summary>
-      #{failure_message}
-  
+  def initialize(node)
+    @classname = node['classname']
+    @name = node['name']
+    failure_node = node.elements['failure']
+    @message = failure_node['message']
+    @details = failure_node.text
+    @count = 1
+  end
+
+  def to_s
+    times = @count > 1 ? " (#{@count} times)" : ''
+    <<~ENTRY
+      <details><summary><tt>#{@name}</tt> in <tt>#{@classname}</tt>#{times}</summary>
+      #{@message}
+
       ```
-      #{failure_node.text}
+      #{@details}
       ```
       </details>
     ENTRY
-  else
-    puts "- Note: Flaky test \`#{test_class}`.`#{test_name}`: #{failure_message}"
+  end
+
+  def key
+    "#{@classname}.#{@name}"
+  end
+
+  def ultimately_succeeds?(parent_node:)
+    all_same_test_nodes = parent_node.get_elements("testcase[@classname='#{@classname}' and @name='#{@name}']")
+    !all_same_test_nodes.last.elements['failure'] # If last node found for that test doesn't have a <failure> child, then test ultimately succeeded.
+  end
+
+  def ==(other)
+    @classname == other.classname && @name == other.name && @message == other.message && @details == other.details
   end
 end
 
-if failures.empty?
-  puts "No test failure found (after eliminating intermediate failed retries). Removing any previous `#{title}` Buildkite annotation if any."
-  system('buildkite-agent', 'annotation', 'remove', '--context', title)
-  exit 0 # We don't want to fail if the `annotation remove` command failed because there was no previous annotation
-else
-  puts "Ultimately found #{failures.count} real test failure(s). Reporting them as a `#{title}` Buildkite annotation."
-  markdown = "\#\#\#\# #{title}: #{failures.length} failure(s)\n\n" + failures.join("\n")
-  puts markdown
-  system('buildkite-agent', 'annotate', markdown, '--style', 'error', '--context', title)
-  exit $?.exitstatus
+# Add a test_failure into a given array, or increment its count if it already exists
+#
+def record(test_failure, into:)
+  existing = into.find { |f| f == test_failure }
+  if existing.nil?
+    into.append(test_failure)
+  else
+    existing.count += 1
+  end
 end
+
+# Given a list of failures for a given step and state, update the corresponding annotation
+#
+def update_annotation(title, list, style, state)
+  annotation_context = "#{title} (#{state})"
+  if list.empty?
+    puts "No test #{state}. Removing any previous `#{annotation_context}` Buildkite annotation if any.\n\n"
+    system('buildkite-agent', 'annotation', 'remove', '--context', annotation_context)
+  else
+    tests_count = list.map(&:key).uniq.count # Count the number of tests that failed, even if some tests might have multiple assertion failures
+    puts "#{tests_count} test(s) #{state} (#{list.count} distinct assertion failures in total). Reporting them as a `#{annotation_context}` Buildkite #{style} annotation.\n\n"
+    markdown = "\#\#\#\# #{title}: #{tests_count} tests #{state} (#{list.count} distinct assertion failure(s) in total)\n\n" + list.map(&:to_s).join("\n")
+    puts markdown
+    system('buildkite-agent', 'annotate', markdown, '--style', style, '--context', annotation_context)
+  end
+end
+
+###################
+# Main
+###################
+
+file = File.open(junit_path)
+xmldoc = REXML::Document.new(file)
+
+failures = []
+flakies = []
+REXML::XPath.each(xmldoc, '//testcase[failure]') do |node|
+  test_failure = TestFailure.new(node)
+
+  if test_failure.ultimately_succeeds?(parent_node: node.parent)
+    record(test_failure, into: flakies)
+  else
+    record(test_failure, into: failures)
+  end
+end
+
+update_annotation(title, failures, 'error', 'have failed')
+update_annotation(title, flakies, 'warning', 'were flaky')


bin/annotate_test_failures
```diff
@@ -17,28 +17,46 @@ end
...
+    failures.append <<~ENTRY
+      <details><summary><tt>#{test_name}</tt> in <tt>#{test_class}</tt></summary>
+      #{failure_message}
+  
+      ```
+      #{failure_node.text}
+      ```
+      </details>
+    ENTRY
+  else
```
AliSoftware: Side note that if a test case was retried, say, 3 times, but failed in all 3 attempts (so, _not_ ultimately succeeding), then all 3 failures for that same test will be reported (as opposed to, say, the last failure only)

This is intentional, because it's not so uncommon that flaky tests fail at different points in the test and/or with different error messages during their various retries.

<details><summary>That's actually exactly what happened for <code>testDeleteFollowerFailure</code> in <a href="https://buildkite.com/automattic/wordpress-ios/builds/12940#0186c2cf-d9f3-440b-b8db-67ca28a271a9">this CI build</a>, and why we'd want to log <i>all</i> intermediate failures for those cases that never succeeded even after multiple retries.</summary>
image

Even if, in this specific case, it ultimately passed on 4th retry (which is why the whole CI step ended up green, and you can also see that by looking in the `report.junit`), you can imagine a case where it wouldn't have ultimately succeeded, yet it'd still be interesting to see the 3 different error messages for each failure instead of only the last one. Especially to avoid accidentally misleading the person who will try to debug that test failure into thinking that the test always fails at the same point in code and with the same error, when the reality is that it's happening more randomly.
</details>

bin/annotate_test_failures
```diff
@@ -17,28 +17,46 @@ end
...
 end
 
 if failures.empty?
-  puts "No test failure found. Removing any previous `#{title}` Buildkite annotation if any."
+  puts "No test failure found (after eliminating intermediate failed retries). Removing any previous `#{title}` Buildkite annotation if any."
   system('buildkite-agent', 'annotation', 'remove', '--context', title)
   exit 0 # We don't want to fail if the `annotation remove` command failed because there was no previous annotation
 else
-  puts "Found #{failures.count} test failure(s). Reporting them as a `#{title}` Buildkite annotation."
+  puts "Ultimately found #{failures.count} real test failure(s). Reporting them as a `#{title}` Buildkite annotation."
```
mokagio: What do you think of adopting a wording similar to that of the `failures.empty? == true` branch?

```suggestion
  puts "Found #{failures.count} test failure(s) (multiple retries that failed in the same way count as one failure). Reporting them as a `#{title}` Buildkite annotation."
```

bin/annotate_test_failures
mokagio: > Side note that if a test case was retried, say, 3 times, but failed in all 3 attempts (so, not ultimately succeeding), then all 3 failures for that same test will be reported (as opposed to, say, the last failure only)
>
> This is intentional, because it's not so uncommon that flaky tests fail at different points in the test and/or with different error messages during their various retries.

Are we sure that the number of `testcase.failure` nodes is the number of retries?

WordPress iOS is configured to retry on failure up to 3 times, but `testDeleteFollowerFailure` is not marked as a consistent failure despite having 3 failures

image

_retry on failure, maximum three times_

image

_3 failure, 1 success... 4 repetitions?_ 🤔 

This _could_ be a bug in Xcode, but [the way `testDeleteFollowerFailure` is written](https://github.com/wordpress-mobile/WordPress-iOS/blob/e59cda935c7e02b68c8c03afcbcf2cb9b58e6728/WordPress/WordPressTest/PeopleServiceTests.swift#L154-L203) also leaves the door open for the test reporting multiple failures in a single run. After all, it has multiple assertions.

In particular, notice how a failure in the Nimble async code doesn't prevent the test from finishing running and failing again later on (the async code is just how the test is shaped, Nimble `waitUntil` ... `done()` makes the execution "linear"):

image

This is, as far as I know, the expected behavior. Here's a dumb example:

image

mokagio: I run those tests locally, this was the JUnit output:

```
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertTrue failed'>WordPress/WordPressTest/PeopleServiceTests.swift:226</failure>
    </testcase>
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertFalse failed'>WordPress/WordPressTest/PeopleServiceTests.swift:227</failure>
    </testcase>
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertTrue failed'>WordPress/WordPressTest/PeopleServiceTests.swift:226</failure>
    </testcase>
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertFalse failed'>WordPress/WordPressTest/PeopleServiceTests.swift:227</failure>
    </testcase>
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertTrue failed'>WordPress/WordPressTest/PeopleServiceTests.swift:226</failure>
    </testcase>
    <testcase classname='WordPressTest.PeopleServiceTests' name='testMultipleFailuresInOneTest'>
      <failure message='XCTAssertFalse failed'>WordPress/WordPressTest/PeopleServiceTests.swift:227</failure>
    </testcase>
```

That refutes the `testcase.failure` count equals retries hypothesis 😞

mokagio: Running that test report through the changes in this PR (as of https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/46/commits/0727e65e25415f5f1f13bd0255edb6deceb0f3c7) results in the following output

---

Ultimately found 6 real test failure(s). Reporting them as a `Tests` Buildkite annotation.
#### Tests: 6 failure(s)

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertTrue failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:225
```
</details>

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertFalse failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:226
```
</details>

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertTrue failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:225
```
</details>

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertFalse failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:226
```
</details>

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertTrue failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:225
```
</details>

<details><summary><tt>testMultipleFailuresInOneTest</tt> in <tt>WordPressTest.PeopleServiceTests</tt></summary>
XCTAssertFalse failed

```
WordPress/WordPressTest/PeopleServiceTests.swift:226
```
</details>

---

mokagio: I'm not sure where to go from here... I guess we could change the way tests are grouped for the purpose of reporting?

Xcode knows how to differentiate between multiple assertion failures in the same `func test...` and a whole `func test...` failing:

image

That makes me think the `xcresult` output would contain that information, too, and that it is the translation to JUnit XML format that looses it. But, I haven't had a chance to verify it.

AliSoftware: Ahhh damn, you make a good point in the fact that a given test case can have multiple assertion failures surfaced in a single execution… didn't think of that.

That kinda throws a wrench in the whole logic of my implementation… or does it? Because even if a single run can have multiple failures, would it then end up having a non-failing XML node at the end if the test didn't pass?

In other words, yes, if a test case reports 3 failure nodes in the junit report, maybe indeed it doesn't mean 3 retries (each failing once), but instead 3 failures reported in the first time running the test. But even then, I wouldn't expect a `<testcase>` node _without_ a `<failure>` child node… showing up at the end of the list, unless if the test ultimately succeeded. Because while a `XCTAssert` failing would call `XCTestCase.recordFailure(…)` and ultimately report a failure entry in the `xcresult` and `.junit` reports… a non-failing `XCTAssert` assertion does _not_ report a success entry (there's no underlying `XCTestCase.recordSuccess(…)`) on its own.

So I think that, while there are multiple possible scenarios that would lead to having multiple failure nodes (either multiple failures in a single run, or failures happening during multiple retries, or a mix of both), the only way to get a _success_ node (aka a `testcase` node without a child `failure` node) for a test in the `junit` report is if the whole test ultimately passed?

AliSoftware: Said otherwise, even if I agree that the "`testcase.failure` count equals retries" hypothesis is probably wrong after all, the "test case ends up with a success means Xcode ultimately considered it green and marked as passed" hypothesis still holds, making the logic still work (even if we should maybe rename the `retries` variable in the implementation to something like `entries_for_same_test` or similar to avoid misleading the reader)?

---

My new theory is thus that, given the following:

```swift
func testFlaky() {
  XCTAssertTrue(Bool.random(), message: "1")
  XCTAssertTrue(Bool.random(), message: "2")
}
```
And Xcode configured for 3 retries:
 - first try failing in both assertions, second try failing only on `2`, but third try passing, will lead to nodes representing {`failure:1`, `failure:2`,`failure:2`,`success`}
 - first and second try failing on `2`, third and last try failing on `1`, and the test never succeeding despite retries… will lead to {`failure:2`,`failure:2`,`failure:1`}, with no success node.

Still making the "if set ends with success node, then whole test was ultimately marked green, and if not then it was red, whether that was due to multiple retries or multiple assertions in single try or mix of both" logic work…?

AliSoftware:
@mokagio I spent some time (most of my afternoon, tbh 😅 ) testing things around and improved the changes significantly:
 - I [created a `test/failure-retry-junit-report` branch on the WPiOS repo for testing purposes](https://github.com/wordpress-mobile/WordPress-iOS/compare/test/failure-retry-junit-report):
    - Added a new "FlakyTestsDemo" Xcode Unit Test target
    - Added a test plan configured with 3 retries
    - [Created unit tests specifically designed to cover the various cases (always fail, flaky but ultimately pass, skipped, …)](https://github.com/wordpress-mobile/WordPress-iOS/compare/test/failure-retry-junit-report#diff-22ecaed2c611eec8f4b43cf3d735c1a0e7ad8ac8fee7714a1a9e37facc7a46e5R11-R40)
    - Adjusted the fastlane lanes running Unit Tests to run that test plan instead
    - Commented most of the `pipeline.yml` and make it point to this `fix/annotate_test_failures` branch of the `a8c-ci-toolkit-buildkite-plugin`
  - That allowed me to get a `report.junit` XML file that represented a typical report for the various cases you raised, which I then used as a base to test and improve my `annotate_test_failures` script here.

The improvements I made include:
 - Created a dedicated `TestFailure` ruby class to make handling of failure nodes easier to manipulate
 - Detect exact duplicate failures—differentiating between cases when a test reports multiple (distinct) assertion failures in a single run, vs cases when a test is retried and reports the same assertion failure (and on the same file and line) multiple times (one for each failed retry). And track the count of those instead of adding the duplicate to the array of detected failures
 - This allows to build the list of distinct assertion failures, and identify the test that each `TestFailure` is about, allowing us to distinguish the number of distinct failed tests, the number of distinct failed assertions (which can be different if a test had multiple assertion failures), and the number of times each failure was reported (for cases of retries)
 - I'm now also tracking and reporting flaky tests in a dedicated `warning` annotation, in addition to only report true failures in the `error` annotation.

@mokagio Given all the changes I've made after all that intense testing and tweaking, I'm interested in your re-review! (Note: I've updated the testing instructions in the PR description)

bin/annotate_test_failures
AliSoftware: Note: suggestion is outdated now that I've refactored the code and made different annotations for flaky vs failures with new wording, so marking this as "resolved".

bin/annotate_test_failures
mokagio: > "if set ends with success node, then whole test was ultimately marked green, and if not then it was red, whether that was due to multiple retries or multiple assertions in single try or mix of both" logic work…?

Yes, I think that rule would be correct under the assumption that the test suite either has test repetition set as _None_ or _Retry on Failure_.

It might break if we used _Until Failure_ or _Up Until Maximum Repetition_ depending on what one considers a success, give the set might end with a failure in those cases even though previous repetitions were successful. 🤔 Personally, I think both options are useful only for stress testing and given we are not doing any of that, I don't think we should worry about it too much for the moment.

bin/annotate_test_failures
```diff
@@ -14,31 +18,102 @@ unless File.exist?(junit_path)
...
+# Given a list of failures for a given step and state, update the corresponding annotation
+#
+def update_annotation(title, list, style, state)
+  annotation_context = "#{title} (#{state})"
+  if list.empty?
+    puts "No test #{state}. Removing any previous `#{annotation_context}` Buildkite annotation if any.\n\n"
+    system('buildkite-agent', 'annotation', 'remove', '--context', annotation_context)
+  else
+    tests_count = list.map(&:key).uniq.count # Count the number of tests that failed, even if some tests might have multiple assertion failures
+    puts "#{tests_count} test(s) #{state} (#{list.count} distinct assertion failures in total). Reporting them as a `#{annotation_context}` Buildkite #{style} annotation.\n\n"
```
mokagio, line 92: Looking at the CI annotations, I was going to suggest to use "test(s)" because "1 tests" reads odd. But I see you already addressed this 👍 

Screenshot 2023-03-14 at 1 09 37 pm

bin/annotate_test_failures
```diff
@@ -14,31 +18,102 @@ unless File.exist?(junit_path)
...
 
-if failures.empty?
-  puts "No test failure found. Removing any previous `#{title}` Buildkite annotation if any."
-  system('buildkite-agent', 'annotation', 'remove', '--context', title)
-  exit 0 # We don't want to fail if the `annotation remove` command failed because there was no previous annotation
-else
-  puts "Found #{failures.count} test failure(s). Reporting them as a `#{title}` Buildkite annotation."
-  markdown = "\#\#\#\# #{title}: #{failures.length} failure(s)\n\n" + failures.join("\n")
-  system('buildkite-agent', 'annotate', markdown, '--style', 'error', '--context', title)
-  exit $?.exitstatus
```
mokagio, line 43: I wondered a bit about the explicit `exit` disappearing from the code, but I guess it no longer makes sense to bubble up the exit code of the `buildkite-agent annotate` command when we may call it more than once.

The important thing is that the test task reports the correct exit code, but that's already handled [by the way we capture the tests exit code before calling the annotation step](https://github.com/wordpress-mobile/WordPress-iOS/blob/1be5d8cf6be34a2245400a2164375d123fb161b6/.buildkite/commands/run-unit-tests.sh#L15-L38).

mokagio (approved):
This is really excellent. Thank you for taking the time.

mokagio:
While commenting about reporting the return code from the test task rather than the annotation command, I realized that there is nothing platform-specific that forces us to run the annotation logic in the macOS agent. 

We _could_ move it to a cheaper agent with Ruby support, but:

1. The build time overhead of the annotation is negligible compared to the test run time
2. Having tests and annotation together is handy if we retry the tests. I'm not sure how it would work if we split them... Does Buildkite handle a retry step re-triggering steps that are downstream from it? 🤔 No sure... and given point 1, don't think it's worth finding out

bin/annotate_test_failures
AliSoftware: Exactly my reasoning for why I removed it 🙂👍

bin/annotate_test_failures
AliSoftware: Yeah the test CI jobs on my dummy WPiOS test branch didn't pick up the change even if I retried the builds because of cache (since the plugin [was referenced in the `pipeline.yml` as `automattic/a8c-ci-toolkit#fix/annotate_test_failures`, so by branch name, not commit](https://github.com/wordpress-mobile/WordPress-iOS/blob/bcfad407146ab9077fb77a5a7e085340af776ff2/.buildkite/pipeline.yml#L5), and it didn't feel worth it to go all the way to invalidate cache or reference by commit and retry all the test builds just for getting that `(s)` change and update the screenshots 😄

PR #47, opened 2023-03-22
Add note about checking for the latest release in `README`

mokagio:
This should save us the effort of updating the `README` every time we ship a new release.

I also removed the old `v` prefix from the snippet, so it won't cause issue if someone copy pastes it.


README.md
```diff
@@ -26,9 +26,11 @@ steps:
       restore_cache $(hash_file package-lock.json)
 
     plugins:
-      - automattic/a8c-ci-toolkit#v2.14.0
+      - automattic/a8c-ci-toolkit#2.15.0
 ```
 
+Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/tag/2.15.0) is and use that value instead of `2.15.0`.
```
oguzkocer: I think linking the "latest release" to the `2.15.0` would be counter-productive. Should we use the link for the releases page instead?

```suggestion
Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/) is and use that value instead of `2.15.0`.
```

AliSoftware: +1

mokagio: Good point. I'll one-up you on that and use https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest

I discovered the existence of `/latest` when working on this change. But as you can see by following the link, it redirects to the actual latest URL. So I copy pasted that URL believing it was `/latest` _and didn't notice it..._ 😅

mokagio:
```suggestion
Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `2.15.0`.
```

mokagio:
I noticed that the linter fails... I opened #48 to try addressing it, but haven't gotten to the bottom of it yet.

AliSoftware (approved):

AliSoftware:
> This should save us the effort of updating the README every time we ship a new release.

Note that Buildkite's `plugin-linter` will still enforce us to update that value in the example every time we do a new tag … so I think we'll still have to update the README on every new release anyway 🤷‍♂️

.buildkite/pipeline.yml
AliSoftware, line 1: FYI: I just updated the required checks in GitHub's branch protection settings to account for the name changes made here.

PR #48, opened 2023-03-23
Update a link that was using old bash-cache repo name

mokagio:
Noticed the linter failing with "No readme config examples found with plugin id 'automattic/a8c-ci-toolkit'. Have you configured the linter with the correct plugin id?" in #47.

I've updated the README, but the linter is still failing 🤔 . [This should be the source for the linter command](https://github.com/buildkite-plugins/buildkite-plugin-linter/blob/851ebed34bc6e38598fc530bb3160df29efd2d5e/lib/linters/example-linter.js#L64).

While trying to understand the issue, I noticed a few leftover references to `bash-cache`. This include one fix for an edge case in which `save_cache` and `restore_cache` would break.


README.md
```diff
@@ -26,14 +26,17 @@ steps:
...
 ```
 
 Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `2.15.0`.
 
 ## Configuration
 
-There are no configuration options for this plugin
+### `bucket` (Optional, string)
+
+The name of the S3 bucket to fallback to if the `CACHE_BUCKET_NAME` environment variable is not set in the CI host. Use by `save_cache` and `restore_cache`.
```
mokagio, line 39: I question the value of this option. I guess it's useful to have is something breaks, but I wonder whether we'd be better off fixing a CI agent that doesn't expose that env var, rather than updating the pipeline to pass one?

It might be different if the option allowed to _override_ the bucket name from the CI agent, but that's not the case. [See code below](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/48/files#r1146082800).

Besides, I don't know of any repo that uses it.

_Maybe a way to address that linter failure (albeit without understanding why it fails...) would be to remove the configuration._

bin/restore_cache
```diff
@@ -4,12 +4,12 @@ CACHE_KEY=$1
...
 if [ -z "$S3_BUCKET_NAME" ]; then
-	if [ -z "$BUILDKITE_PLUGIN_BASH_CACHE_BUCKET" ]; then
+	if [ -z "$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_BUCKET" ]; then
 		echo "⛔Unable to restore file from cache – no \$CACHE_BUCKET_NAME is set"
 		exit 1
 	else
-		echo "Reading bucket name from 'BUILDKITE_PLUGIN_BASH_CACHE_BUCKET'"
-		S3_BUCKET_NAME="$BUILDKITE_PLUGIN_BASH_CACHE_BUCKET"
+		echo "Reading bucket name from 'BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_BUCKET'"
+		S3_BUCKET_NAME="$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_BUCKET"
```
mokagio, line 12: Unless I'm mistaken, this code uses `$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_BUCKET` only if `$CACHE_BUCKET_NAME` is empty (`-z`).

AliSoftware:
@mokagio at that point it might be worth pinging Buildkite support via email, or via a GitHub issue on their plugin-linter, to get some help from them understanding why the linter is failing while it shouldn't? 🤔

AliSoftware:
@mokagio After a lot of investigation and a plunge into the rabbit hole of [the `plugin-linter`'s `example-linter.js`](https://github.com/buildkite-plugins/buildkite-plugin-linter/blob/master/lib/linters/example-linter.js) source code and unit tests, I found out why the plugin-linter fails on our example 😓 

The reason is because the linter [expects tag names to always start with `v`, as in `#v2.16.0`](https://github.com/buildkite-plugins/buildkite-plugin-linter/blob/master/lib/linters/example-linter.js#L110).
 - See how [adding the `v` in front of the version](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/commit/24883fb8f954013d590f4a5b842b895ceb98539e) made [the "No readme config examples found" failure disappear (and now it correctly finds that `v2.15.0` is not the latest version, which is a different test)](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/454#01884467-67b2-4bb3-90b6-6e692647d703/120-138).
 - And then [updating the example to `#v2.16.0`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/commit/c48dacc2ebf5f2ff83e55e9f863221e30829204d) (while our latest tag is `2.16.0`, without the `v`) made [the linter finally pass](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/455#0188446a-c1bd-49b9-a9cc-33726dfd8395).

The rest of Buildkite's `plugin-linter` tests (especially the one validating that the example references the latest tag) use [the `compare-version` npm library](https://www.npmjs.com/package/compare-versions) to run [`compareVersion(latestTag, hashValueInYaml)`](https://github.com/buildkite-plugins/buildkite-plugin-linter/blob/master/lib/linters/readme-version-number-linter.js#L19)… and since `compareVersion('1.2.3', 'v1.2.3')` returns `0` (aka "equal") †, that's why using `v2.16.0` still makes the test pass even if our tag name is just `2.16.0` not `v2.16.0`.

<details><summary><em>† I've tested this with the<a href="https://npm.runkit.com/compare-versions">RunKit playground</a> to confirm</em></summary>
image
</details>

The big question that remains is, would it work in reality to use `v2.16.0` if our tag name is just `2.16.0`? i.e.:
 - Would Buildkite find the plugin and the right tag if we used `automattic/a8c-ci-toolkit#v2.16.0` in the `pipeline.yml` files of our client repos by somehow automatically stripping the `v` from it? If so, then providing an example that uses the `v` in our `README.md` is valid and we'd be good to go.
 - Or would it fail because it uses a different logic when trying to find the tag to pull when using the plugin in the world than the one that the plugin-linter unit tests are using? If so, and thus the only correct usage would be to reference it using `automattic/a8c-ci-toolkit#2.16.0` without the `v`, then Buildkite will need to update their `plugin-linter` so that their testing logic supports tags without a `v` prefix. Or alternatively we'll have to start using the `v` prefix in our tags… 😒 

I guess the only way to know if referencing a plugin using `#v2.16.0` would work even if the tag name is just `2.16.0`… is to test it on one of our client repos and find out? Though it'd feel strange if Buildkite made it so we could provide a `#v` part that doesn't exactly match the tag name (what if we ended up having both a `v1.2.3` and a `1.2.3` tag in our repo then?) so my bet would be that it won't work and thus we need to (ask Buildkite to) fix the plugin-lint… or start using `v` in our tag names

AliSoftware:
> I guess the only way to know if referencing a plugin using #v2.16.0 would work even if the tag name is just 2.16.0… is to test it on one of our client repos and find out? Though it'd feel strange if Buildkite made it so we could provide a #v part that doesn't exactly match the tag name (what if we ended up having both a v1.2.3 and a 1.2.3 tag in our repo then?) so my bet would be that it won't work and thus we need to (ask Buildkite to) fix the plugin-lint… or start using v in our tag names

As expected, [confirmed on a dummy branch of WPiOS that using `#v2.16.0` to reference the plugin version in the client yaml doesn't work](https://buildkite.com/automattic/wordpress-ios/builds/14291#0188447c-a758-41fa-ac96-2705cc0c74c8)

So now we have to decide to go with one of the following routes:
1. Accept to have this README incorrectly have an example that mentions `v` in the tag name
2. Ask Buildkite to fix their `plugin-linter` code logic to accept tags not starting with `v`
3. Provide a PR to fix Buildkite's `plugin-linter` ourselves
4. Or start using `v` prefix in all our future tags for Buildkite plugins

Given how Buildkite support is quite reactive (from my recent experience at least), I'd suggest to start with option 2 and see if they're willing to fix it soon.

AliSoftware:
> So now we have to decide to go with one of the following routes:
> […]
> 2\. Ask Buildkite to fix their `plugin-linter` code logic to accept tags not starting with `v`
> […]
> Given how Buildkite support is quite reactive (from my recent experience at least), I'd suggest to start with option 2 and see if they're willing to fix it soon.

I finally took the time to open https://github.com/buildkite-plugins/buildkite-plugin-linter/issues/478 to report the issue to Buildkite.

AliSoftware:
> I finally took the time to open [buildkite-plugins/buildkite-plugin-linter#478](https://github.com/buildkite-plugins/buildkite-plugin-linter/issues/478) to report the issue to Buildkite.

And [they just replied 1h ago](https://github.com/buildkite-plugins/buildkite-plugin-linter/issues/478#issuecomment-1642436806), telling that they merged a fix for this already, and should release a new version around next week 🙂

mokagio:
🎉 Thank you @AliSoftware and Buildkite team!

AliSoftware:
@mokagio I've force-pushed some commits to update the version in the README to the latest as of today (`2.18.2`) and to fix the issue with using the latest `plugin-linter` (+ some quick `Makefile` cleanup while I was at it)… and it seems that we're now all green 🎉 

I'll let you rebase both this PR and your `mokagio/update-readme-for-version-info` branch on top of our latest `trunk` and finally merge it all once ready 🙂

.buildkite/pipeline.yml
```diff
@@ -1,9 +1,6 @@
 steps:
-  - label: ":shell: Lint (Shellcheck)"
-    command: make shellcheck
```
AliSoftware, line 3: Removed this because the `shellcheck` make target is already part of the `lint` target that we call in the other step, so we were running it twice.

README.md
```diff
@@ -26,14 +26,17 @@ steps:
       restore_cache $(hash_file package-lock.json)
 
     plugins:
-      - automattic/a8c-ci-toolkit#2.15.0
+      - automattic/a8c-ci-toolkit#2.18.2:
+          bucket: a8c-ci-cache # optional
 ```
 
 Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `2.15.0`.
```
AliSoftware:
```suggestion
Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `2.18.2`.
```

AliSoftware (approved):

PR #50, opened 2023-04-14
`annotate_test_failure`: Absorb potential failure in annotation removal

AliSoftware:
This is to avoid a misleading red error log to be output if no annotation existed when we tried to remove it.

See https://github.com/wordpress-mobile/WordPress-iOS/pull/20378

## Testing

 - This fix is being tested in [this CI build of WPiOS](https://buildkite.com/automattic/wordpress-ios/builds/13716).

## Next Steps

 - Release a new version `2.15.1`
 - Update [this WPiOS PR](https://github.com/wordpress-mobile/WordPress-iOS/pull/20378) to point to the new version/tag (in _all_ `.yml` pipelines) once its released.

## Checklist

### DIFF
--- a/tests/environment.bats
+++ b/tests/environment.bats
@@ -1,6 +1,6 @@
 #!/usr/bin/env bats
 
-load '/usr/local/lib/bats/load.bash'
+load "$BATS_PLUGIN_PATH/load.bash"
 
 fail() {
   echo "Failure: $1"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,8 +42,8 @@ _None._
 
 ### Bug Fixes
 
-_None._
+- Prevent `annotate_test_failures` to print a misleading red error message when trying to remove previous annotation. [#50]
 
 ### Internal Changes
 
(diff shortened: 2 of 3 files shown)


tests/environment.bats
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env bats
 
-load '/usr/local/lib/bats/load.bash'
+load "$BATS_PLUGIN_PATH/load.bash"
```
AliSoftware, line 3: This is not directly related to the PR itself, but was done because `make tests` started [failing on this](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/401#01878078-eff5-4a1d-a188-5187577048dd/120-148), making the PR go red.
I figured that the plugin-tester docker image might have been updated recently and broke things, so I took a look at [the plugin-tester readme](https://github.com/buildkite-plugins/buildkite-plugin-tester#usage) and noticed they were recommending to use this env var here, and that indeed fixed the CI issue.

mokagio (approved):
Nice!

PR #51, opened 2023-05-01
Re-implement `comment_on_pr` to use `curl` instead of `gh`

crazytonyli:
The existing script depends on `gh` and macOS (installing `gh` when it's missing). I've re-implemented to make it compatible to Linux too, as long as `curl` and `jq` are installed.

I recorded this as an internal change, since the interface is backwards compatible.

## Test instructions

Go to this PR's buildkite job page, select a step, and go to the "Environment" tab. Click the "Show export prefix" button and copy the environment variables. Open a terminal on your computer and paste the copied env vars. Invoking this script from your shell session should post a comment to this PR, or fail with an non-zero exist code.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,3 +47,4 @@ _None._
 ### Internal Changes
 
 - Added this changelog file. [#49]
+- Re-implement the `comment_on_pr` script. [#51]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,8 @@ _None._
 
 ### New Features
 
-_None._
+- Add a new `github_api` utlity to make API requests to GitHub.com or Github Enterprise instances. [#51]
+- Extend the `comment_on_pr` utility to update or delete the existing comment. [#51]
 
 ### Bug Fixes
 
@@ -48,4 +49,3 @@ _None._
 
 - Added this changelog file. [#49]
 - Re-implement the `comment_on_pr` utility to use `curl` instead of `gh`. [#51]
-- Extend the `comment_on_pr` utility to update or delete the existing comment. [#51]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,4 +47,4 @@ _None._
 ### Internal Changes
 
 - Added this changelog file. [#49]
-- Re-implement the `comment_on_pr` script. [#51]
+- Re-implement the `comment_on_pr` utility to use `curl` instead of `gh`. [#51]
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -48,3 +48,4 @@ _None._
 
 - Added this changelog file. [#49]
 - Re-implement the `comment_on_pr` utility to use `curl` instead of `gh`. [#51]
+- Extend the `comment_on_pr` utility to update or delete the existing comment. [#51]
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -63,14 +63,13 @@ case "$#" in
     *) echo "Error: invalid comment argument"; exit 1 ;;
 esac
 
+export GITHUB_TOKEN="$arg_github_token"
+
 github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
 github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
-issues_endpoint="https://api.github.com/repos/${github_user}/${github_repo}/issues"
+issues_endpoint="repos/${github_user}/${github_repo}/issues"
 
-curl -s --fail -o /dev/null \
-    -H "Authorization: token ${arg_github_token}" \
-    -H "Accept: application/vnd.github+json" \
-    "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
+github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail -o /dev/null \
     || (echo "The GitHub token is invalid" && exit 1)
 
 # Find the existing PR comment
@@ -82,10 +81,7 @@ if [[ -n "$opt_comment_id" ]]; then
 
     # Do not change this variable value, it's used to identify the comments posted by this utility.
     comment_body_id="<!-- DO NOT REMOVE ci-toolkit-comment-identifier: $opt_comment_id -->"
-    existing_comment_id="$(curl -s \
-        -H "Authorization: token ${arg_github_token}" \
-        -H "Accept: application/vnd.github+json" \
-        "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
+    existing_comment_id="$(github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
         | jq ".[] | select(.body | contains(\"$comment_body_id\")) | .id")"
 
     # When comment id is provided without a comment body, delete the existing comment.
@@ -100,10 +96,8 @@ fi
 # Delete the existing comment if needed
 if [[ -n "$existing_comment_id" && "$opt_if_exist" == "delete" ]]; then
     echo "Delete the existing comment containing $existing_comment_id"
-    curl -s --fail-with-body -X DELETE \
-        -H "Authorization: token ${arg_github_token}" \
-        -H "Accept: application/vnd.github+json" \
-        "$issues_endpoint/comments/$existing_comment_id"
+    github_api "$issues_endpoint/comments/$existing_comment_id" \
+        --fail-with-body -X DELETE
 fi
 
 # Construct the comment JSON
@@ -112,25 +106,22 @@ $comment_body_id
 $arg_pr_comment
 EOF
 )
-json_payload=$(jq --null-input --arg body "$comment_body" '{body: $body}')
+json_payload=$(jq -c --null-input --arg body "$comment_body" '{body: $body}')
 
 if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
     echo "Update the existing comment: $existing_comment_id"
-    curl -s --fail-with-body -X PATCH \
-        -H "Authorization: token ${arg_github_token}" \
+    github_api "$issues_endpoint/comments/$existing_comment_id" \
+        --fail-with-body -X PATCH \
         -H "Content-Type: application/json" \
-        -H "Accept: application/vnd.github+json" \
-        -d "${json_payload}" \
-        "$issues_endpoint/comments/$existing_comment_id"
+        -d "${json_payload}"
+
 elif [[ -n "$arg_pr_comment" ]]; then
     echo "Post a new comment"
-    curl -s --fail-with-body -X POST \
-        -H "Authorization: token ${arg_github_token}" \
+    github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
+        --fail-with-body -X POST \
         -H "Content-Type: application/json" \
-        -H "Accept: application/vnd.github+json" \
-        -d "${json_payload}" \
-        "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments"
+        -d "${json_payload}"
 else
-    # No comment body was given in CLI, so do nothing
-    echo "No comment to post"
+    # No comment body was given in CLI, so no new comment to post
+    echo "No new comment to post"
 fi
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -7,8 +7,19 @@
 # 2. The GitHub token to use to post the comment
 
 # Check dependencies and print their versions for diagnosis purposes
-jq --version
-curl --version
+cat <<EOF
+> jq --version
+$(jq --version)
+
+> curl --version
+$(curl --version)
+
+EOF
+
+if [[ "${BUILDKITE_PULL_REQUEST:-invalid}" == "invalid" ]]; then
+    echo "Error: this tool can only be called from a Buildkite PR job"
+    exit 1
+fi
 
 if [[ -f "$1" ]]; then
     PR_COMMENT=$(cat "$1")
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -47,6 +47,10 @@ while [[ "$#" -gt 0 ]]; do
 done
 
 case "$#" in
+    0)
+        arg_pr_comment=""
+        arg_github_token=${GITHUB_TOKEN}
+        ;;
     1)
         arg_pr_comment=""
         arg_github_token=$1
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -39,7 +39,7 @@ fi
 # Default options
 opt_comment_id=""
 opt_if_exist="update"
-opt_github_token="${GITHUB_TOKEN-:}"
+opt_github_token="${GITHUB_TOKEN:-}"
 
 # If there are two arguments and they don't start with an option, then we assume
 # this utility is called in the deprecated way: comment_on_pr <comment> <github-token>
@@ -74,7 +74,7 @@ else
                 arg_pr_comment="$1"
             fi
             ;;
-        *) echo "Error: invalid comment argument"; exit 1 ;;
+        *) echo "Error: too many arguments"; exit 1 ;;
     esac
 fi
 
@@ -129,7 +129,6 @@ if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
         --fail-with-body -X PATCH \
         -H "Content-Type: application/json" \
         -d "${json_payload}"
-
 elif [[ -n "$arg_pr_comment" ]]; then
     echo "Post a new comment"
     github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -84,10 +84,14 @@ github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
 github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
 issues_endpoint="repos/${github_user}/${github_repo}/issues"
 
-all_pr_comments=$(
-    github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail \
-    || (>&2 echo "Error: Failed to use the GitHub token to retreive PR comments. Please double check the GitHub token is valid and has permission to read and write PR comments in ${github_user}/${github_repo} repo." && exit 1)
-)
+pr_comments_file=$(mktemp)
+
+if ! github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail-with-body --output "$pr_comments_file";
+then
+    echo "Error: Failed to use the GitHub token to retreive PR comments."
+    [[ -f "$pr_comments_file" ]] && cat "$pr_comments_file"
+    exit 1
+fi
 
 # Find the existing PR comment
 if [[ -n "$opt_comment_id" ]]; then
@@ -98,8 +102,7 @@ if [[ -n "$opt_comment_id" ]]; then
 
     # Do not change this variable value, it's used to identify the comments posted by this utility.
     comment_body_id="<!-- DO NOT REMOVE ci-toolkit-comment-identifier: $opt_comment_id -->"
-    existing_comment_id="$(echo "$all_pr_comments" \
-        | jq ".[] | select(.body | contains(\"$comment_body_id\")) | .id")"
+    existing_comment_id="$(jq --arg comment_body_id "$comment_body_id" '.[] | select(.body | contains($comment_body_id)) | .id' "$pr_comments_file")"
 
     # When comment id is provided without a comment body, delete the existing comment.
     if [[ -z "$arg_pr_comment" ]]; then
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -84,8 +84,10 @@ github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
 github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
 issues_endpoint="repos/${github_user}/${github_repo}/issues"
 
-github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail -o /dev/null \
-    || (echo "The GitHub token is invalid" && exit 1)
+all_pr_comments=$(
+    github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail \
+    || (>&2 echo "Error: The GitHub token is invalid" && exit 1)
+)
 
 # Find the existing PR comment
 if [[ -n "$opt_comment_id" ]]; then
@@ -96,7 +98,7 @@ if [[ -n "$opt_comment_id" ]]; then
 
     # Do not change this variable value, it's used to identify the comments posted by this utility.
     comment_body_id="<!-- DO NOT REMOVE ci-toolkit-comment-identifier: $opt_comment_id -->"
-    existing_comment_id="$(github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
+    existing_comment_id="$(echo "$all_pr_comments" \
         | jq ".[] | select(.body | contains(\"$comment_body_id\")) | .id")"
 
     # When comment id is provided without a comment body, delete the existing comment.
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -31,7 +31,7 @@ $(curl --version)
 
 EOF
 
-if [[ "${BUILDKITE_PULL_REQUEST:-invalid}" == "invalid" ]]; then
+if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
     echo "Error: this tool can only be called from a Buildkite PR job"
     exit 1
 fi
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -7,12 +7,13 @@
 #
 # Examples:
 #
-#    # Post a new comment which can be uniqly identifiered by the given comment id,
+#    # Post a new comment which can be uniquely identified by the given comment id,
 #    # or update the existing comment that's associated with the given comment id.
-#    # You can choose to delete the existing comment if it exists by passing `--if-exist delete`.
+#    # By default, the existing comment will be updated. You can choose to delete
+#    # the existing comment by passing `--if-exist delete`.
 #    comment_on_pr --id comment-id [--if-exist update|delete] "This is a comment"
 #
-#    # Delete the comment associated with the given comment id
+#    # Delete the comment associated with the given comment id by passing an empty comment
 #    comment_on_pr --id comment-id [--if-exist delete]
 #
 #    # Deprecated: Post a new comment to the PR
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -86,7 +86,7 @@ issues_endpoint="repos/${github_user}/${github_repo}/issues"
 
 all_pr_comments=$(
     github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail \
-    || (>&2 echo "Error: The GitHub token is invalid" && exit 1)
+    || (>&2 echo "Error: Failed to use the GitHub token to retreive PR comments. Please double check the GitHub token is valid and has permission to read and write PR comments in ${github_user}/${github_repo} repo." && exit 1)
 )
 
 # Find the existing PR comment
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -43,7 +43,7 @@ opt_github_token="${GITHUB_TOKEN:-}"
 
 # If there are two arguments and they don't start with an option, then we assume
 # this utility is called in the deprecated way: comment_on_pr <comment> <github-token>
-if [[ $# == 2 && "$1" != "--" ]]; then
+if [[ $# == 2 && "$1" != "--"* ]]; then
     if [[ -f "$1" ]]; then
         arg_pr_comment=$(cat "$1")
     else
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -3,17 +3,20 @@
 # This script is used to comment on a pull request, and must be run in a buldkite PR step.
 #
 # Usage:
+#    [GITHUB_TOKEN=<token>] comment_on_pr [--id <comment-id>] [--if-exist update|delete] [--github-token <token>] <comment>
 #
-#    # Post a new comment to the PR
-#    comment_on_pr "This is a comment" <github-token>
+# Examples:
 #
 #    # Post a new comment which can be uniqly identifiered by the given comment id,
 #    # or update the existing comment that's associated with the given comment id.
 #    # You can choose to delete the existing comment if it exists by passing `--if-exist delete`.
-#    comment_on_pr --id comment-id "This is a comment" [--if-exist update|delete] <github-token>
+#    comment_on_pr --id comment-id [--if-exist update|delete] "This is a comment"
 #
 #    # Delete the comment associated with the given comment id
-#    comment_on_pr --id comment-id <github-token>
+#    comment_on_pr --id comment-id [--if-exist delete]
+#
+#    # Deprecated: Post a new comment to the PR
+#    comment_on_pr "This is a comment" <github-token>
 #
 # Please note the comment argument can be a path to a file containing the comment, or a markdown.
 
@@ -35,39 +38,46 @@ fi
 # Default options
 opt_comment_id=""
 opt_if_exist="update"
+opt_github_token="${GITHUB_TOKEN-:}"
 
-while [[ "$#" -gt 0 ]]; do
-    case $1 in
-        --id) opt_comment_id="$2"; shift ;;
-        --if-exist) opt_if_exist="$2"; shift ;;
-        --*) echo "Unknown option: $1"; exit 1 ;;
-        *) break ;;
+# If there are two arguments and they don't start with an option, then we assume
+# this utility is called in the deprecated way: comment_on_pr <comment> <github-token>
+if [[ $# == 2 && "$1" != "--" ]]; then
+    if [[ -f "$1" ]]; then
+        arg_pr_comment=$(cat "$1")
+    else
+        arg_pr_comment="$1"
+    fi
+
+    opt_github_token=$2
+else
+    while [[ "$#" -gt 0 ]]; do
+        case $1 in
+            --id) opt_comment_id="$2"; shift ;;
+            --if-exist) opt_if_exist="$2"; shift ;;
+            --github-token) opt_github_token="$2"; shift ;;
+            --*) echo "Unknown option: $1"; exit 1 ;;
+            *) break ;;
+        esac
+        shift
+    done
+
+    case "$#" in
+        0)
+            arg_pr_comment=""
+            ;;
+        1)
+            if [[ -f "$1" ]]; then
+                arg_pr_comment=$(cat "$1")
+            else
+                arg_pr_comment="$1"
+            fi
+            ;;
+        *) echo "Error: invalid comment argument"; exit 1 ;;
     esac
-    shift
-done
+fi
 
-case "$#" in
-    0)
-        arg_pr_comment=""
-        arg_github_token=${GITHUB_TOKEN}
-        ;;
-    1)
-        arg_pr_comment=""
-        arg_github_token=$1
-        ;;
-    2)
-        if [[ -f "$1" ]]; then
-            arg_pr_comment=$(cat "$1")
-        else
-            arg_pr_comment="$1"
-        fi
-
-        arg_github_token=$2
-        ;;
-    *) echo "Error: invalid comment argument"; exit 1 ;;
-esac
-
-export GITHUB_TOKEN="$arg_github_token"
+export GITHUB_TOKEN="$opt_github_token"
 
 github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
 github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -2,9 +2,20 @@
 
 # This script is used to comment on a pull request, and must be run in a buldkite PR step.
 #
-# It takes two arguments:
-# 1. The comment to post, or the path to a file containing the comment
-# 2. The GitHub token to use to post the comment
+# Usage:
+#
+#    # Post a new comment to the PR
+#    comment_on_pr "This is a comment" <github-token>
+#
+#    # Post a new comment which can be uniqly identifiered by the given comment id,
+#    # or update the existing comment that's associated with the given comment id.
+#    # You can choose to delete the existing comment if it exists by passing `--if-exist delete`.
+#    comment_on_pr --id comment-id "This is a comment" [--if-exist update|delete] <github-token>
+#
+#    # Delete the comment associated with the given comment id
+#    comment_on_pr --id comment-id <github-token>
+#
+# Please note the comment argument can be a path to a file containing the comment, or a markdown.
 
 # Check dependencies and print their versions for diagnosis purposes
 cat <<EOF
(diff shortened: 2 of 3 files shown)


crazytonyli:
test

crazytonyli:
test again

crazytonyli:
test again

CHANGELOG.md
```diff
@@ -47,3 +47,4 @@ _None._
 ### Internal Changes
 
 - Added this changelog file. [#49]
+- Re-implement the `comment_on_pr` script. [#51]
```
mokagio: What do you think about adding more detail here?

```suggestion
- Re-implement the `comment_on_pr` utility to use `curl` instead of `gh`. [#51]
```

bin/comment_on_pr
```diff
@@ -1,14 +1,33 @@
...
-GITHUB_TOKEN=$2
+# This script is used to comment on a pull request, and must be run in a buldkite PR step.
+#
+# It takes two arguments:
+# 1. The comment to post, or the path to a file containing the comment
+# 2. The GitHub token to use to post the comment
+
+# Check dependencies and print their versions for diagnosis purposes
+jq --version
+curl --version
```
mokagio: This is a handy way to check if the tools are available :) 

Compared to the `command -v <tool>...` approach, this doesn't give us the opportunity of suggesting to the user what to do to resolve. 

I think that's okay in this context, since these scripts are meant to be used in machines we control and have enough knowledge of how to address a failure if `curl` or `jq` were missing.

mokagio (commented):
I tried running this locally and hit a couple of issues because not all builds will have a `BUILDKITE_PULL_REQUEST_REPO` or `BUILDKITE_PULL_REQUEST` value.

E.g. notice how only the build at the top of the list here has a PR associated with it:

Screenshot 2023-05-02 at 1 58 57 pm

I was able to work around the missing `BUILDKITE_PULL_REQUEST_REPO` by using `BUILDKITE_REPO` instead:

```bash
# GITHUB_URL="${BUILDKITE_PULL_REQUEST_REPO%.git}"
URL_WITHOUT_SUFFIX="${BUILDKITE_REPO%.git}"
GITHUB_URL="${URL_WITHOUT_SUFFIX#git@github.com:}"
```

Only after working on that I realized my fix was pointless because if `BUILDKITE_PULL_REQUEST_REPO` is empty and `BUILDKITE_PULL_REQUEST` is `false`, then clearly we don't have a PR to comment on 🤦‍♂️ 

I'm not sure how `gh` would have dealt with this situation, but looking at the existing code my guess is that it would figured out internally whether the commit from which it was called (i.e. the state in which the repo was) belonged to a PR.

I guess we could exit early if `BUILDKITE_PULL_REQUEST` is `false` but that might resulting in missing out on some messages for rebuilds or for commits that trigger a build before the author has a chance to open a PR?

The second option could be achieved via something like

```bash
curl -s -X GET \
         --fail-with-body \
         -H "Authorization: token ${GITHUB_TOKEN}" \
         -H "Content-Type: application/json" \
         -H "Accept: application/vnd.github+json" \
         "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/commits/${BUILDKITE_COMMIT}/pulls" \
          | jq '.[0].number'
```

mokagio:
test comment from Gio

mokagio:
[My comment above](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51#issuecomment-1530849169) was posted with a modified version of this script hacked to test the suggestions I made

```diff
diff --git a/bin/comment_on_pr b/bin/comment_on_pr
index 6b01903..a957f74 100755
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -1,4 +1,4 @@
-#!/bin/bash -eu
+#!/bin/bash -eux
 
 # This script is used to comment on a pull request, and must be run in a buldkite PR step.
 #
@@ -18,16 +18,24 @@ fi
 
 GITHUB_TOKEN=$2
 
-GITHUB_URL="${BUILDKITE_PULL_REQUEST_REPO%.git}"
+URL_WITHOUT_SUFFIX="${BUILDKITE_REPO%.git}"
+GITHUB_URL="${URL_WITHOUT_SUFFIX#git@github.com:}"
 GITHUB_USER=$(basename "$(dirname "$GITHUB_URL")")
 GITHUB_REPO=$(basename "$GITHUB_URL")
 
+# TODO: What to do if there is more than one PR?
+# TODO: Maybe this could be a fallback if BUILDKITE_PULL_REQUEST is false, not the default approach?
+PR=$(curl -s -X GET --fail-with-body -H "Authorization: token ${GITHUB_TOKEN}" -H "Content-Type: application/json" -H "Accept: application/vnd.github+json" "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/commits/${BUILDKITE_COMMIT}/pulls" | jq '.[0].number')
+
+# FIXME: Fail if there is no PR
+
 JSON_PAYLOAD=$(jq --null-input --arg body "$PR_COMMENT" '{body: $body}')
 
 curl -s -X POST \
-    --fail-with-body \
-    -H "Authorization: token ${GITHUB_TOKEN}" \
-    -H "Content-Type: application/json" \
-    -H "Accept: application/vnd.github+json" \
-    -d "${JSON_PAYLOAD}" \
-    "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/issues/${BUILDKITE_PULL_REQUEST}/comments"
+  --location \
+  --fail-with-body \
+  -H "Authorization: token ${GITHUB_TOKEN}" \
+  -H "Content-Type: application/json" \
+  -H "Accept: application/vnd.github+json" \
+  -d "${JSON_PAYLOAD}" \
+  "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/issues/${PR}/comments"
```

The `--location` was a leftover from a previous attempt to follow a redirect. When using `BUILDKITE_REPO`, the value was the legacy `bash-cache...` URL. Unfortunately, the command still failed, but I now realize it was because the commit I was using had no PR. The solution I found for that was to update the URL for this pipeline in the settings so that the GitHub API would not redirect. We might want to support redirect in the future, though.

crazytonyli:
Given this script is designed to be called only in PR jobs, I think we should reply on those env vars only present in PR jobs? Because technically there are difference between a job that builds a branch and a job that builds a PR.

> E.g. notice how only the build at the top of the list here has a PR associated with it:

I think that might be related to this project's buildkite pipeline settings. Buildkite kicks off a branch job when new branch is created or new commits is pushed. If a PR is created after the branch job has completed, buildkite doesn't start an new PR build. That's probably because the "Skip pull request builds for existing commit" is turned on.

Screenshot 2023-05-02 at 8 43 50 PM

If you search for "skip_pull_request_builds_for_existing_commits" in one of our private repos (that shall not be named 😄 ), you'll see some projects (like wordpress-ios and woocommerce-ios) have this option turned off.

```
$ fgrep skip_pull_request_builds_for_existing_commits -r buildkite/pipelines --no-filename | sort | uniq -c
  30     skip_pull_request_builds_for_existing_commits = false
  84     skip_pull_request_builds_for_existing_commits = true
```

I think it would be good to review this at some point and see if it's necessary to have different settings on our projects.

AliSoftware:
> Given this script is designed to be called only in PR jobs, I think we should reply on those env vars only present in PR jobs? Because technically there are difference between a job that builds a branch and a job that builds a PR.

I agree with this.

> I think it would be good to review this at some point and see if it's necessary to have different settings on our projects.

Totally, and this is already in our backlog to review it all at some point anyway (with your project to manage Buildkite settings via terraform and all). Personally I suggest we don't create builds for commits not associated with Pull Requests, except for the long-running branches (`trunk`/`develop`/…). If one want to have CI test their work in progress even if they're not ready to have the work reviewed yet they can always create a Draft PR anyway. On the other end, letting CI create builds for commits without a PR all while having the "cancel previous builds" enabled (which we want) will often make the CI check first turn red on the PR (because the build for the commit push would have been created first, then immediately cancelled and marking the PR red) only to later finish the 2nd build (the one associated with PR creation) and running the CI steps again and hopefully finalluy turning them back green.

See also private discussions in Slack: p1678808720955539/1678807631.262069-slack-CR5E8MXLZ and p1677163098116359/1677160312.855239-slack-CR5E8MXLZ

mokagio:
> Given this script is designed to be called only in PR jobs, I think we should reply on those env vars only present in PR jobs? Because technically there are difference between a job that builds a branch and a job that builds a PR.

Sounds good to me. I'd still be keen to add an explicit check for `BUILDKITE_PULL_REQUEST` and `BUILDKITE_PULL_REQUEST_REPO` in the script, though, so that if they are missing we can let the "user" know about it. What do you think?

crazytonyli:
> I'd still be keen to add an explicit check

For sure. Added in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/1bf308501b9987300def5233379ad262579c3588

AliSoftware:
I'd love for this action to evolve so that it can include a comment id, allow for updating an existing comment with that comment id (instead of adding a new comment) if one exists already, and allow to delete an existing comment by passing an empty comment body.

Kinda similar to how [our equivalent `comment_on_pr` fastlane action is doing it in `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/lib/fastlane/plugin/wpmreleasetoolkit/actions/common/comment_on_pr.rb)

_imho it'd make more sense to have a single `comment_on_pr` command like that here that handles everything rather than having multiple commands (`comment_on_pr` and `delete_pr_comment` etc)—especially if we want `comment_on_pr` to already be able to delete/update an existing comment if one with passed commentID was found, then no need to implement the deletion logic twice._

See https://github.com/Automattic/buildkite-ci/pull/170#discussion_r1185077675 for a typical place where this could be useful and allow some nice DRYing, as well as [this comment on recently-merged `delete_pr_comment` PR](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/52#issuecomment-1535024307).

crazytonyli:
@AliSoftware I'm happy to add this additional feature. However, as I mentioned in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/52#issuecomment-1535472163, adding this feature won't be able to replace how some use cases use `delete_pr_comment` before `comment_on_pr`.

AliSoftware:
> @AliSoftware I'm happy to add this additional feature. However, as I mentioned in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/52#issuecomment-1535472163, adding this feature won't be able to replace how some use cases use `delete_pr_comment` before `comment_on_pr`.

I think it still would, given that my suggestion above also included the mention:

> […] and allow to delete an existing comment by passing an empty comment body.

So all you'd have to do is call `comment_on_pr --id terraform-plan-aws ""` (*) as an in-place replacement for the places where you currently wanted to call `delete_pr_comment`.

_(*) This was just an API suggestion amongst others. Now that I think more about it, maybe making the command API looking more like `comment_on_pr --id xyz --delete` (rather than relying on empty body param) would be better for this use case, up to you_

crazytonyli:
new comment

crazytonyli:
new comment

crazytonyli:
comment with id

crazytonyli:
second comment with id

crazytonyli:
third comment with id

crazytonyli:
comment with id

crazytonyli:
comment

crazytonyli:
new comment

crazytonyli:
@AliSoftware Thanks for the heads up. I thought about that and added a `--if-exist update|delete` option.

crazytonyli:
@mokagio @AliSoftware I've requested a review since there are quite lots a new changes. Please see the "Usage" comment in the script for some use cases. I have verified all use cases work. But please feel free to give it a go and let me know if you find any issues. Thanks!

crazytonyli:
This is a test

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
 #!/bin/bash -eu
 
-PR_MESSAGE_FILE=$1
-GITHUB_TOKEN=$2
```
crazytonyli, line 4: I used lowercase for the variable names. Using upper case, plus the fact that there are lots of variable declared in this script, makes me feel like the whole script is screaming.

crazytonyli:
This is a test

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+        if [[ -f "$1" ]]; then
+            arg_pr_comment=$(cat "$1")
+        else
+            arg_pr_comment="$1"
+        fi
+
+        arg_github_token=$2
+        ;;
+    *) echo "Error: invalid comment argument"; exit 1 ;;
+esac
```
AliSoftware: Is this implementation of accepting GH token and comment as "unnamed parameters" done to keep backwards compatibility with currently shipped implementation or something? (I don't think so because I thing the old implementation using `gh` didn't read the GitHub token from the args in the first place? but just checking)

💡  If not, I think it would be nicer to have at least the GH token parameters be a "named" params just (like `--id` and `--if-exist` above). And also, we might want to consider making it default to the `$GITHUB_TOKEN` env var, since:
 - Most of us have that value already exported in our env by our `.zshrc`, so making the script pick that env var up by default would make it easier to use / test the command
 - It would avoid having to pass the token as parameter that could potentially end up being visible if the invoked command is printed in logs.

---

```ruby
arg_github_token="$GITHUB_TOKEN" # default value
while [[ "$#" -gt 0 ]]; do
    case $1 in
        --id) opt_comment_id="$2"; shift ;;
        --if-exist) opt_if_exist="$2"; shift ;;
        --token) arg_github_token="$2"; shift ;;
        --*) echo "Unknown option: $1"; exit 1 ;;
        *) break ;;
    esac
    shift
done
```

---

💭 Also, regarding `arg_pr_comment` being parsed either as a file path or as a comment string, I think it's unnecessary to add that logic (and kinda-hidden-even-if-documented-in-comments feature) given one could achieve the same at call site using simple `<` shell input redirection (`comment_on_pr <my-comment.txt`). Which means that once we'd have moved the parsing of GH token to the `getopts`-style `while` loop, the above could be reduced to just this:
```suggestion
arg_pr_comment="$1"
```
And if one needs to pass the comment as a file, they could simply use `<myfile.txt` for `$1` and let the shell handle it all.

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+
+github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
+github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
```
AliSoftware: 💡  Super small nitpick because I'm not even sure we'd have a real use case for it but otoh it could also be nice to have the flexibility if we need to given it'd be easy to implement: what about allowing those to be passed via `--repo-slug` parameter? Like, if `--repo-slug automattic/myrepo` is passed, use that, otherwise use the above two lines to infer the slug to use? 🤷 

(But feel free to disregard this idea because while the thought passed my head, I'm not even sure it's worth it in practice)

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+issues_endpoint="https://api.github.com/repos/${github_user}/${github_repo}/issues"
```
AliSoftware: ⚠️  We should ideally make the `https://api.github.com` host / endpoint be configurable (`--github-host` or something?), so that we can use this command not only on GitHub.com but also on our private GitHub Enterprise instance for Tumblr.

Alternatively, maybe that value is already exposed or derivable from some ENV var (e.g. maybe we could extract it from `BUILDKITE_PULL_REQUEST_REPO` or similar)?

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+
+# Find the existing PR comment
+if [[ -n "$opt_comment_id" ]]; then
+    # Do not change this variable value, it's used to identify the comments posted by this utility.
+    comment_body_id="<!-- DO NOT REMOVE ci-toolkit-comment-identifier: $opt_comment_id -->"
+    existing_comment_id="$(curl -s \
+        -H "Authorization: token ${arg_github_token}" \
+        -H "Accept: application/vnd.github+json" \
+        "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
+        | jq ".[] | select(.body | contains(\"$comment_body_id\")) | .id")"
```
AliSoftware: Side note that this might cause issues if `$opt_comment_id` happens to have quotes or special characters in it.

I'm not sure this is a big deal (i.e. not sure we expect any practical call sites where that could be the case), and not sure if it would be easy to solve the limitation, but felt it was worth mentioning anyway.

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+        "$issues_endpoint/comments/$existing_comment_id"
+fi
+
+# Construct the comment JSON
+comment_body=$(cat <<EOF
+$comment_body_id
+$arg_pr_comment
+EOF
+)
+json_payload=$(jq --null-input --arg body "$comment_body" '{body: $body}')
```
AliSoftware: ~Unlike the case about just the comment ID I mentioned above, here it feels way more likely that the body of comment we'd want to post could contain quotes or special characters somewhere in its text. I'm expecting such use cases to cause trouble with this line…~

Actually I've tested that case [here](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51#issuecomment-1535933692) and it seems it doesn't cause any issue and actually works as expected, so nvm! 👌 

```
$ bin/comment_on_pr --id 'alisoftware-test-1' "@AliSoftware says: \"testing, 1,2,1,2\"... I'll add some \$special \`characters\` too" $GITHUB_TOKEN
```
image

bin/comment_on_pr
```diff
@@ -1,14 +1,125 @@
...
+    echo "Post a new comment"
+    curl -s --fail-with-body -X POST \
+        -H "Authorization: token ${arg_github_token}" \
+        -H "Content-Type: application/json" \
+        -H "Accept: application/vnd.github+json" \
+        -d "${json_payload}" \
+        "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments"
+else
+    # No comment body was given in CLI, so do nothing
+    echo "No comment to post"
```
AliSoftware: 💄  Nitpick: This "so do nothing" here is slightly misleading because if no comment body was given, then we might still have done something (earlier in the script, namely delete existing comment via L90-96) if we still passed an id and an existing comment with that id was found. So maybe this might be a more appropriate wording?

```suggestion
    # No comment body was given in CLI, so no new comment to post
    echo "No new comment to post"
```

AliSoftware:
@AliSoftware says: "testing, 1,2,1,2"... I'll add some $special `characters` too

bin/comment_on_pr
AliSoftware: Also note that when I tested this command the first time, I didn't pass the GitHub token as the second parameter to the command, and that led the script to error with an obscure message:

```
jq: error (at <stdin>:4): Cannot index string with string "body"
```
Took me a second to realize that this error was because the `curl` request returned an error message/body instead of the expected JSON due to me lacking to pass the appropriate GitHub token as parameter.

Not a blocker but would be nice to add some error handling or preliminary test about token being passed or whatnot to make such error more clear. And one more reason why it would also be nice to have `arg_github_token` pick up the existing `GITHUB_TOKEN` env var by default if not provided explicitly 😉

AliSoftware: 🛑  Confirmed the issue:

```
$ bin/comment_on_pr --id 'Olivier aka "AliSoftware" is testing' "@AliSoftware here Testing with a \"tricky\" comment ID full of spaces and quotes." $GITHUB_TOKEN
> jq --version
jq-1.6

> curl --version
curl 7.87.0 (x86_64-apple-darwin22.0) libcurl/7.87.0 (SecureTransport) LibreSSL/3.3.6 zlib/1.2.11 nghttp2/1.51.0
Release-Date: 2022-12-21
Protocols: dict file ftp ftps gopher gophers http https imap imaps ldap ldaps mqtt pop3 pop3s rtsp smb smbs smtp smtps telnet tftp
Features: alt-svc AsynchDNS GSS-API HSTS HTTP2 HTTPS-proxy IPv6 Kerberos Largefile libz MultiSSL NTLM NTLM_WB SPNEGO SSL threadsafe UnixSockets

jq: error: syntax error, unexpected IDENT, expecting ';' or ')' (Unix shell quoting issues?) at <top-level>, line 1:
.[] | select(.body | contains("<!-- DO NOT REMOVE ci-toolkit-comment-identifier: Olivier aka "AliSoftware" is testing -->")) | .id                                                                                              
jq: 1 compile error
```

Sure thing it might be quite unlikely that we'd use a tricky comment ID with quotes and all, at least when we're using explicit/hardcoded comment IDs. But I can see a case where the comment ID could be derived from something dynamic, like `terraform-plan-$PROJECT_NAME` but then `$PROJECT_NAME` might contain special characters like quotes, backslash, etc.

We had a similar case for when I implemented [`annotate_test_failures`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/annotate_test_failures) where the `--context` parameter for the annotation to add/update was derived and built from the buildkite step name amongst other things, which could contain special characters as well; so it's not like it's _totally_ unlikely to happen…

---

I haven't tried but I think that using `jq --arg body "$comment_body_id"` could be a possible solution to solve this?

AliSoftware (changes requested):
Mostly works as expected, but still found some edge cases we might want to fix:

 - ⚠️ We should allow the command to be run on GH:E too (for Tumblr), not just github.com — i.e. allow to override the `api.github.com` with `github.tumblr.net/api/v3` or whatever is needed
 - 🛑  The command fails if we pass a comment ID with special characters (which is not as unlikely as one might think when comment ID is derived from other env vars). I think we can use `jq --arg` to solve that.
 - 💡  It would be nice to make the script default to using `$GITHUB_TOKEN` env var if no token is provided as parameter. And while we're at it, use named parameters for that GH token param to make call site more readable (since we already have named parameters for the rest of them…)
 - 💭 Finally, we can avoid handling the special case of comment-param-is-a-file-path given one can already achieve the same using `<` shell input redirection at call site (`comment_on_pr <myfile.txt`); removing that custom logic will also simplify the code quite a bit.

Otherwise the logic looks good to me, thanks for working on this!

bin/comment_on_pr
crazytonyli: Yes, the token and comment parsing is for backwards compatibility. In the latest release version 2.15.1, the first argument is [a file path](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/2.15.1/bin/comment_on_pr#L3) and the second argument is [the GitHub token](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/2.15.1/bin/comment_on_pr#L4).

> Also, regarding arg_pr_comment being parsed either as a file path or as a comment string,

Accepting a file path is for backwards compatibility. Accepting a comment is a new feature, which is for caller's convenience.

bin/comment_on_pr
crazytonyli: This issue actually came to me, but I decided to ignore it. It's probably looking for trouble if the caller decides to use "unusual" identifier, like quotes as you mentioned, newline potentially, or with `-->` in it. Reason I decided to ignore it is because the comment id is a hardcoded text created by developers, rather than taken from a user input. I don't think we would go look for troubles by using special characters in the code.

However, I think it'd be a good idea to at least perform a simple validation, like only `[0-9a-zA-Z\-_]` is acceptable characters.

crazytonyli: I've added a step to valid the token by making an API request: https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/88ff81de10b3867d8d522109162a897cddd18e36

crazytonyli: ⬆️ is implemented in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/0a74452d45f0e755669830bcc3995519109f5376

crazytonyli: > it would also be nice to have arg_github_token pick up the existing GITHUB_TOKEN env var by default if not provided explicitly

Done in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/11671dd87536d4e65444eb5b843a0c70d415d1f0

bin/comment_on_pr
crazytonyli: Added another utility `github_api` in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/8e4f9c75d2ebeb04e33fd9e0ed0a99252b4e74cf which support both GHE and GitHub.com

bin/comment_on_pr
AliSoftware: > Reason I decided to ignore it is because the comment id is a hardcoded text created by developers, rather than taken from a user input

Ideally yes I agree. But in practice I can see cases like the ID being `report-test-failures-$BUILDKITE_STEP` then if the buildkite step name contains spaces it will cause issues…?

But maybe it's a non issue after all and we can keep it as a limitation/restriction for now and improve later if we really need such use case. Limiting to `[0-9a-zA-Z_-]` for now is probably a good start indeed.

bin/comment_on_pr
AliSoftware: Oh. TIL. That's unfortunate that we implemented it that was back then.

I wonder how many callers of this utility we have in the wild and if any of them really need to pass a file in practice vs if this was a limitation of the implementation detail of using `gh` while in practice all current call sites write the string they want to comment in a file just before calling the utility while they could have passed the comment directly if that old implementation  supported it… 🤔

crazytonyli: I'm not sure where this utility is called in our repos. That's why I'd like to make it backward compatible. BTW, I've simplified the argument parsing in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/d80a8724852a2054a0252c76b2d14cf13fe7b205, by having a dedicated clause for parsing the deprecated CLI: `comment_on_pr <comment> <github token>`.

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
-PR_MESSAGE_FILE=$1
-GITHUB_TOKEN=$2
+# This script is used to comment on a pull request, and must be run in a buldkite PR step.
+#
+# Usage:
+#    [GITHUB_TOKEN=<token>] comment_on_pr [--id <comment-id>] [--if-exist update|delete] [--github-token <token>] <comment>
+#
+# Examples:
+#
+#    # Post a new comment which can be uniqly identifiered by the given comment id,
```
AliSoftware: 💄 
```suggestion
#    # Post a new comment which can be uniquely identified by the given comment id,
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+#    # or update the existing comment that's associated with the given comment id.
+#    # You can choose to delete the existing comment if it exists by passing `--if-exist delete`.
```
AliSoftware: 💄 
```suggestion
#    # You can choose to delete the existing comment if it exists by passing `--if-exist delete` (defaults to `--if-exist update`).
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+#    comment_on_pr --id comment-id [--if-exist update|delete] "This is a comment"
+#
+#    # Delete the comment associated with the given comment id
```
AliSoftware: 💄 
```suggestion
#    # Delete the comment associated with the given comment id by passing an empty comment
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+cat <<EOF
+> jq --version
+$(jq --version)
+
+> curl --version
+$(curl --version)
+
+EOF
+
+if [[ "${BUILDKITE_PULL_REQUEST:-invalid}" == "invalid" ]]; then
```
AliSoftware: 🛑  When a Build is not associated to a PR, Buildkite defines this env var with the value `false`, not empty or undefined like one might think (see Environment tab in [an example build like here](https://buildkite.com/automattic/wordpress-ios/builds/14082#0187fda0-9e41-4e0c-98a5-9ecccaaaf01e).
So that's the value we want to test for this env var.

```suggestion
if [[ "${BUILDKITE_PULL_REQUEST:-false}" == "false" ]]; then
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+    echo "Error: this tool can only be called from a Buildkite PR job"
+    exit 1
+fi
+
+# Default options
+opt_comment_id=""
+opt_if_exist="update"
+opt_github_token="${GITHUB_TOKEN-:}"
```
AliSoftware: 🛑 
```suggestion
opt_github_token="${GITHUB_TOKEN:-}"
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+
+# If there are two arguments and they don't start with an option, then we assume
+# this utility is called in the deprecated way: comment_on_pr <comment> <github-token>
+if [[ $# == 2 && "$1" != "--" ]]; then
```
AliSoftware: 🛑  This only tests if `$1` is equal to `--`, not if it starts with `--`.

---

❓  Besides, could one still call this tool with `comment_on_pr "My comment" --id foo`, passing the options flags _after_ the fixed parameter?

I think the way you implemented `getopts`-like option parsing in your `while` loop below we can't, and we're expected to pass the `--` options first (unlike true `getopts` which allows this iirc), so if that's a requirement we have and agree to keep then I'm ok with the logic (as long as we fix the "equal" to a "begin with" in your test above;, but if not, then you might want to adjust this condition to also test that `$2` doesn't start with `--` either.

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+            arg_pr_comment=""
+            ;;
+        1)
+            if [[ -f "$1" ]]; then
+                arg_pr_comment=$(cat "$1")
+            else
+                arg_pr_comment="$1"
+            fi
+            ;;
+        *) echo "Error: invalid comment argument"; exit 1 ;;
```
AliSoftware: 💄  It's not so much that the $1 comment argument is invalid here, but rather that we have more arguments than we expected.
```suggestion
        *) echo "Error: too many arguments"; exit 1 ;;
```

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
 
-FILENAME="/tmp/github-token-$BUILDKITE_BUILD_ID"
-echo "$GITHUB_TOKEN" > "$FILENAME"
+export GITHUB_TOKEN="$opt_github_token"
+
+github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
+github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
+issues_endpoint="repos/${github_user}/${github_repo}/issues"
+
+github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail -o /dev/null \
```
AliSoftware: 🤔  Maybe instead of `-o /dev/null` we can store the result of this request in a variable, so that you can reuse it in line 98—instead of redoing the exact same request on line 98 only to pass it thru `jq`?

bin/comment_on_pr
```diff
@@ -1,14 +1,141 @@
...
+)
+json_payload=$(jq -c --null-input --arg body "$comment_body" '{body: $body}')
+
+if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
+    echo "Update the existing comment: $existing_comment_id"
+    github_api "$issues_endpoint/comments/$existing_comment_id" \
+        --fail-with-body -X PATCH \
+        -H "Content-Type: application/json" \
+        -d "${json_payload}"
+
```
AliSoftware: 💄 
```suggestion
```

bin/github_api
AliSoftware, line 1: 🤔  I think the name of that utility might be misleading, as I first thought that `github_api` would return the API endpoint to call (i.e. just return the `url`) while in fact it actually does the `curl` call.

Maybe rename this `call_github_api` instead to make it more explicit?

AliSoftware (changes requested):

bin/github_api
crazytonyli: I used this name because it's similar, in terms of interface and feature to `gh api <endpoint>`. How about `curl_github_api` 🥁  , to indicate that the utility accepts curl options?

bin/comment_on_pr
crazytonyli: Thanks for the suggestions. Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/96cc99f44591c8e0720d246e26141c7a053bcb9e

bin/comment_on_pr
crazytonyli: Good catch! Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/26dd39a40ee0c303ecee1f9f42dc1c325eb38988

bin/comment_on_pr
crazytonyli: 🙈 Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/d283690d67d2f709077d03a8835dcdca3e3db3db

bin/comment_on_pr
crazytonyli: 🙈 Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/d9fed6260cb02315ea076116980aa9f1a6e34f6d

> if that's a requirement we have and agree to keep

Having the options before the argument is an intention "feature". The only reason is it's easy and straight-forward to parse such cli arguments. This is not something that I explicitly brought up for discussion, because I thought it would be a nice and simple improvement to the existing way of parsing arguments which is based on the argument position. We can definitely standardize the argument parsing by using `getopt`, but I think we could do it in a separate PR if it's needed?

bin/comment_on_pr
crazytonyli: Good call! Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/a2df1e2e8c2d563c5ee9cee4dd8a3b1d67db90cb

bin/github_api
AliSoftware: Mmmh at this point this is really just bikeshedding and personal preference, so I won't force one naming over the other.

I like the explicit inclusion of `curl` to indicate that it accepts curl options, but I also feel like this makes the name of the utility less discoverable and less of a "verb-like / English-sentence-like" usage compared to something like `call_github_api` or something similar? But I'll let you decide 🙃

bin/comment_on_pr
AliSoftware: To be clear, I don't mind if we have the requirement that all `--` options have to appear first, especially if it makes the implementation simpler; given we're the users of those utilities and all, it probably isn't worth it to make them _that_ flexible around the parameters order, and I'm ok to keep the current parsing logic that expects `--` first.

I was just mentioning this in order to make sure that your test on `if [[ $# == 2 && "$1" != "--" ]]; then` didn't make any incorrect assumptions or failed to consider a case that we were supposed to support (arbitrary param order), but I'm ok with us not supporting that case in the first place.

bin/comment_on_pr
```diff
@@ -1,14 +1,143 @@
...
+export GITHUB_TOKEN="$opt_github_token"
+
+github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
+github_repo=$(basename "${BUILDKITE_PULL_REQUEST_REPO%.git}")
+issues_endpoint="repos/${github_user}/${github_repo}/issues"
+
+all_pr_comments=$(
+    github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" --fail \
+    || (>&2 echo "Error: The GitHub token is invalid" && exit 1)
```
AliSoftware: 🛑  The `|| (…)` creates a subshell, so your `exit 1` inside this only exits the subshell within `(…)` instead of exiting the main script.

I had a doubt if this was the case or not so I tested it in a small shell and confirmed:
```bash
#!/bin/bash

echo "Hello"
false || (>&2 echo "Error" && exit 1)
echo "Bye"
```
```
$ ./test.sh 
Hello
Error
Bye
```

bin/comment_on_pr
```diff
@@ -1,14 +1,143 @@
...
+# Find the existing PR comment
+if [[ -n "$opt_comment_id" ]]; then
+    if ! [[ "$opt_comment_id" =~ ^[a-zA-Z0-9_-]+$ ]]; then
+        echo "Error: comment id can only contain alphanumeric characters, dashes and underscores"
+        exit 1
+    fi
+
+    # Do not change this variable value, it's used to identify the comments posted by this utility.
+    comment_body_id="<!-- DO NOT REMOVE ci-toolkit-comment-identifier: $opt_comment_id -->"
+    existing_comment_id="$(echo "$all_pr_comments" \
```
AliSoftware: 🤔  I am not sure about this but I hope that if the `all_pr_comments` JSON contains special characters and `"…"` quotes in it, this `echo "$all_pr_comment" | jq …` won't risk getting tripped… worth testing with a concrete use case to ensure that this is not a problem

bin/comment_on_pr
AliSoftware: Also, are we sure that the only possible reason for the request to fail would always be because "The GitHub token is invalid"?

Maybe it would make more sense to move the logic of `curl` response check into the `github_api` utility instead, and make that utility check for the `$?` exit status of `curl` and if it's non-zero, print the error output of the request in `stderr`… Though… now that I think about it I'm expecting `curl … --fail` to do that (print request error message on stderr) already, so maybe all there's to do is `github_api … --fail-with-body || exit 1` here or something like that? 🤔

bin/github_api
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -eu
+
+# Similar to `gh api <endpoint>`, this utility makes API calls to GitHub.
+#
+# ENVIRONMENT VARIABLES:
+#   GITHUB_TOKEN: Required. The GitHub token to use for authentication.
+#   GH_HOST: Optional. The GitHub host to use. We'll figure out the host from env vars if not set.
```
AliSoftware: 💄 
```suggestion
#   GH_HOST: Optional. The GitHub host to use. We'll figure out the host from `BUILDKITE_*` env vars if not set.
```

bin/comment_on_pr
crazytonyli: Do you mean with a JSON like this?
```json
[{"body":"comment-id $#.% with \"quotes\"","id":1}]
```

I tested the above JSON using this shell and it worked as expected.
```
comment_body_id=comment-id; all_pr_comments='[{"body":"comment-id $#.% with \"quotes\"","id":1}]'; echo "$all_pr_comments" | jq ".[] | select(.body | contains(\"$comment_body_id\")) | .id"
```

bin/comment_on_pr
crazytonyli: This utility has the `-e` flag, which would make the `exit` statement work. I see the expected result on my end if add `-e` option to your `test.sh` script.

> so maybe all there's to do is github_api … --fail-with-body || exit 1 here or something like that?

I don't think we can do that if we'd like to capture the API response to be used later. Because curl prints "transportation errors" on stderr, but prints the HTTP response errors (4xx and 5xx for example) to stdout, which means it would be captured by the `all_pr_comments` variable.

bin/comment_on_pr
AliSoftware: 👌

bin/comment_on_pr
crazytonyli: > are we sure that the only possible reason for the request to fail would always be because "The GitHub token is invalid"?

Definitely not the only reason. Given this script is suppose to be called in PR jobs. The curl errors could be:
1. GitHub API is down
2. The GitHub token is not a token at all.
3. The GitHub token is correct, but do not have permission to look PR comments in the repo.

I have updated the error message with more details in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/b9104aa7bcfca34369c54fbbf100aceff1f459cf.

AliSoftware: > This utility has the -e flag, which would make the exit statement work. I see the expected result on my end if add -e option to your test.sh script.

Ah! Good catch! 👍

> […] but [curl] prints the HTTP response errors (4xx and 5xx for example) to stdout, which means it would be captured by the all_pr_comments variable.

🤔 Then maybe the `github_api` utility needs to handle the "capture curl's stdout, then if `$? -eq 0` then just return the captured string… but if non-zero then print the captured error on `stderr` instead, and forward the exit code"?

💡 Which would then allow the caller to just `github_api … -fail || exit 1` without having to print an error message because `github_api` would already have printed the error details to stderr at that point …

crazytonyli: ~It just occurred to me that I can use `--output /dev/stderr` to print the HTTP errors to stderr. See https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/7c45813bf76b5cfca16742d33cb24221da08dd65.~

Update: nope, that wouldn't work. The `all_pr_comments` would be empty on successful API calls. 🥲

crazytonyli: I don't want the `github_api` utility to do too much. `curl` command has loads of options to tweak its behaviour. It's probably more flexible to let the user decides what they want.

AliSoftware: > It just occurred to me that I can use --output /dev/stderr to print the HTTP errors to stderr. See 7c45813.

Does that still capture the result in `all_pr_comments` in case of success? Wouldn't that print the request's valid JSON response to `stderr` too even if no error, and thus not have it captured into `all_pr_comments` as expected?

AliSoftware: > I don't want the github_api utility to do too much. curl command has loads of options to tweak its behaviour. It's probably more flexible to let the user decides what they want.

If there's already a set of curl options we can use to tell it "print output to stdout on success but to stderr on failure" (so that valid requests' responses get captured in `$all_pr_comments` on success but error is printed to stderr, instead of captured, on error) then I'm all for it!

AliSoftware: Ah! Check [this](https://stackoverflow.com/a/18284752)! 🙂

crazytonyli: I re-worked this part a little bit in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/add969cf7cb55dfa7f016607d7ef0c233dd91106 and also addressed [your concern around escaping CLI arguments](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51#discussion_r1190923791). Let me know what you think.

AliSoftware: That looks perfect, thanks! 👌

AliSoftware (approved):
All changes seems to have been addressed, looks good to me! 👍  :shipit:

bin/comment_on_pr
```diff
@@ -1,14 +1,146 @@
...
+                arg_pr_comment="$1"
+            fi
+            ;;
+        *) echo "Error: too many arguments"; exit 1 ;;
+    esac
+fi
+
+export GITHUB_TOKEN="$opt_github_token"
+
+github_user=$(basename "$(dirname "${BUILDKITE_PULL_REQUEST_REPO%.git}")")
```
AliSoftware: 💄  nit: `%.git` is not really necessary for this case because we're taking the `dirname` anyway so don't care if the basename ends in `.git`. Thus that `%.git` is only necessary for the `github_repo` computation below, but not needed for the `github_user` 🙃

crazytonyli: Good catch! Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51/commits/250e38c97aa6591693610f978473998f16af1848

PR #52, opened 2023-05-02
Add `delete_pr_comment`

crazytonyli:
A new script to delete PR comments that contain a given content.

### Test instructions

Go to this PR's buildkite job page, select a step, and go to the "Environment" tab. Click the "Show export prefix" button and copy the environment variables. Open a terminal on your computer and paste the copied env vars. Invoking this script from your shell session should delete comments (or not) based on the `--containing` argument.

---

### DIFF
--- a/bin/delete_pr_comment
+++ b/bin/delete_pr_comment
@@ -6,8 +6,14 @@
 #   delete_pr_comment --github-token <token> --containing "part of a comment"
 
 # Check dependencies and print their versions for diagnosis purposes
-jq --version
-curl --version
+cat <<EOF
+> jq --version
+$(jq --version)
+
+> curl --version
+$(curl --version)
+
+EOF
 
 # Parse command line arguments
 while [[ "$#" -gt 0 ]]; do
--- a/bin/delete_pr_comment
+++ b/bin/delete_pr_comment
@@ -1,0 +1,48 @@
+#!/bin/bash -eu
+
+# This script is used to delete comments on a pull request, and must be run in a buldkite PR step.
+#
+# Example:
+#   delete_pr_comment --github-token <token> --containing "part of a comment"
+
+# Check dependencies and print their versions for diagnosis purposes
+jq --version
+curl --version
+
+# Parse command line arguments
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --containing) COMMENT_CONTAINING="$2"; shift ;;
+        --github-token) GITHUB_TOKEN="$2"; shift ;;
+        *) echo "Unknown option: $1"; exit 1 ;;
+    esac
+    shift
+done
+
+function require_option() {
+    if [[ -z "$2" ]]; then
+        echo "Error: $1 is required"
+        exit 1
+    fi
+}
+
+require_option '--containing' "$COMMENT_CONTAINING"
+require_option '--github-token' "$GITHUB_TOKEN"
+
+GITHUB_URL="${BUILDKITE_PULL_REQUEST_REPO%.git}"
+GITHUB_USER=$(basename "$(dirname "$GITHUB_URL")")
+GITHUB_REPO=$(basename "$GITHUB_URL")
+
+COMMENT_IDS="$(curl -s \
+    -H "Authorization: token ${GITHUB_TOKEN}" \
+    -H "Accept: application/vnd.github+json" \
+    "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/issues/${BUILDKITE_PULL_REQUEST}/comments" \
+    | jq ".[] | select(.body | contains(\"$COMMENT_CONTAINING\")) | .id")"
+
+while IFS= read -r comment_id || [[ -n $comment_id ]]; do
+    echo "Delete the comment containing $COMMENT_CONTAINING: $comment_id"
+    curl -s --fail-with-body -X DELETE \
+        -H "Authorization: token ${GITHUB_TOKEN}" \
+        -H "Accept: application/vnd.github+json" \
+        "https://api.github.com/repos/${GITHUB_USER}/${GITHUB_REPO}/issues/comments/$comment_id"
+done < <(printf '%s' "$COMMENT_IDS")
--- a/bin/delete_pr_comment
+++ b/bin/delete_pr_comment
@@ -1,6 +1,6 @@
 #!/bin/bash -eu
 
-# This script is used to delete comments on a pull request, and must be run in a buldkite PR step.
+# This script is used to delete comments on a pull request, and must be run in a Buildkite PR step.
 #
 # Example:
 #   delete_pr_comment --github-token <token> --containing "part of a comment"
--- a/bin/delete_pr_comment
+++ b/bin/delete_pr_comment
@@ -35,6 +35,11 @@ function require_option() {
 require_option '--containing' "$COMMENT_CONTAINING"
 require_option '--github-token' "$GITHUB_TOKEN"
 
+if [[ "$COMMENT_CONTAINING" != "<!--"*"-->" ]]; then
+    echo "Error: for safety reasons, the '--containing' argument must be a markdown comment: '<!-- comment -->'"
+    exit 1
+fi
+
 GITHUB_URL="${BUILDKITE_PULL_REQUEST_REPO%.git}"
 GITHUB_USER=$(basename "$(dirname "$GITHUB_URL")")
 GITHUB_REPO=$(basename "$GITHUB_URL")
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Add a new script: `delete_pr_comment --github-token <token> --containing "part of a comment"`. [#52]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-- Add a new script: `delete_pr_comment --github-token <token> --containing "part of a comment"`. [#52]
+- Add a new script: `delete_pr_comment --github-token <token> --containing '<!-- identifier -->'`. [#52]
 
 ### Bug Fixes
 


crazytonyli:
test comment

crazytonyli:
test comment again

crazytonyli:
comment

bin/delete_pr_comment
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -eu
+
+# This script is used to delete comments on a pull request, and must be run in a buldkite PR step.
```
mokagio:
```suggestion
# This script is used to delete comments on a pull request, and must be run in a Buildkite PR step.
```

mokagio:
Test comment 1

mokagio:
Test comment 2

mokagio:
Comment test 1

mokagio:
Comment test 3

mokagio:
Comment test 3

bin/delete_pr_comment
```diff
@@ -0,0 +1,48 @@
...
+
+require_option '--containing' "$COMMENT_CONTAINING"
+require_option '--github-token' "$GITHUB_TOKEN"
+
+GITHUB_URL="${BUILDKITE_PULL_REQUEST_REPO%.git}"
+GITHUB_USER=$(basename "$(dirname "$GITHUB_URL")")
+GITHUB_REPO=$(basename "$GITHUB_URL")
+
+COMMENT_IDS="$(curl -s \
+    -H "Authorization: token ${GITHUB_TOKEN}" \
```
mokagio, line 48: Just curious. Is there any advantage in interpolating with `${GITHUB_TOKEN}` over `$GITHUB_TOKEN`?

bin/delete_pr_comment
```diff
@@ -0,0 +1,48 @@
...
+
+function require_option() {
+    if [[ -z "$2" ]]; then
+        echo "Error: $1 is required"
+        exit 1
+    fi
+}
+
+require_option '--containing' "$COMMENT_CONTAINING"
+require_option '--github-token' "$GITHUB_TOKEN"
```
mokagio, line 36: I tried to break this by calling without passing the option and it worked anyway, my guess is because I have `$GITHUB_TOKEN` as an export in my environment.

Screenshot 2023-05-03 at 1 19 57 pm

I don't think this is a blocker in the context of how we can expect this script to be used.

bin/delete_pr_comment
```diff
@@ -0,0 +1,48 @@
...
+done
+
+function require_option() {
+    if [[ -z "$2" ]]; then
+        echo "Error: $1 is required"
+        exit 1
+    fi
+}
+
+require_option '--containing' "$COMMENT_CONTAINING"
```
mokagio, line 35: Because of the `u` option in the shebang, this call will fail before `require_option` gets a chance to run if `$COMMENT_CONTAINING` is empty:

```
./bin/delete_pr_comment: line 29: COMMENT_CONTAINING: unbound variable
```

The result is the same (the script fails) but the UX is not as clear.

bin/delete_pr_comment
```diff
@@ -0,0 +1,48 @@
+#!/bin/bash -eu
+
+# This script is used to delete comments on a pull request, and must be run in a buldkite PR step.
+#
+# Example:
+#   delete_pr_comment --github-token <token> --containing "part of a comment"
+
+# Check dependencies and print their versions for diagnosis purposes
+jq --version
+curl --version
```
mokagio: How do you feel about adding an empty line here so that the version diagnostics info are separated from the rest of the script output?

```suggestion
curl --version
# Print line to separate versions info from the rest of the script output
echo ''
```

mokagio (approved):
I tried it on some tests comments in this PR. Works as advertised 👍 

I'm going to approve to unblock, because nothing I said here is really a blocker.

However, there is one last thing I'd like to ask. I am a bit worried about unintended consequences of selecting any comment matching the given `containing` value. I'm probably overthinking this, because the usage of this script will be quite limited and I assume we'll also introduce our system to use unique identifiers or something (e.g. Danger leaves `<!-- comment id: 1234 -->`-style comments that do not render but are still readable to API calls). Still, I can see someone accidentally passing the wrong string to this command and deleting other people's comments 🤔

bin/delete_pr_comment
crazytonyli: Updated in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/52/commits/5ad7153d746b7a0a20287cafc284707e2784bf89

bin/delete_pr_comment
crazytonyli: Yeah, this is an intended, however undocumented, behaviour.

When invoking this script on CI, we have a few env vars defined by convention, GITHUB_TOKEN being one of them. They are used by many other tools too (`gh`, octokit etc.), it sometimes can be quite handy to just use the env var if it's not passed via CLI. 😄

bin/delete_pr_comment
crazytonyli: Yeah I agree. Considering we have so many bash script, I thought I'll add this simply argument parsing function to give the scripts a better CLI, with `cli --foo value --bar true` instead of `cli value true`. It does its job, but the error message may not be intuitive.

bin/delete_pr_comment
crazytonyli: I guess `${}` is less prone to error, since it clearly states what the variable is. Like `${GITHUB_TOKEN}_VALUE` vs `$GITHUB_TOKEN_VALUE`. I think I copied and pasted a few code here and there, and simply forgot to make them consistent. 🥲

crazytonyli:
@mokagio Regarding your concern on deleting other comment by accident. I definitely agree with you. That's on us to make sure the query criteria is unique enough, i.e. using an invisible code block like you mentioned. I extracted this script from an automation tool in one of our private repo. That tool posts and deletes comments that has certain invisible block content.

But I have added a restriction in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/52/commits/0cdfa64dc528ea9c486b6f1f1a02e1338b9af15b, to only allow query `<!-- xxx -->`, which should reduce the chances of deleting wrong comments.

AliSoftware:
Late to the party, but instead of adding this `delete_pr_comment` command, with the dangerous side effect that Gio highlighted, I'd argue to instead make the `comment_on_pr` command accept a `commentID` param, and then allow that `comment_on_pr` command to delete a comment if the body we pass to it is empty.

That way, not only will we guarantee that we can only delete comments by commentID (and not arbitrary `containing:` test), but also we'd only have one source of truth for how that magic HTML comment is formatted and how to get a GitHub comment handle from the comment id.

---

@crazytonyli  Given all that, I'd advocate to revert that PR / remove that new `delete_pr_comment` command ASAP before we ship a new `a8c-ci-toolkit` version (otherwise if we release a new version before deleting the command, and delete that command only later, we'll need to do a major version bump…), and instead move the logic to delete an existing comment into `comment_for_pr`, as part of the work being done in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/51#issuecomment-1534855103.

Doing that, the cal site would then look like this:
```sh
export GITHUB_TOKEN=…
comment_on_pr --id "terraform-plan-aws" "Go to CI and click on the button to apply the plan if it looks ok" # add a new comment
comment_on_pr --id "terraform-plan-aws" "The plan has now been applied into production" # update existing comment
comment_on_pr --id "terraform-plan-aws" "" # Delete an existing comment completely
```

crazytonyli:
@AliSoftware One side effect with this approach is, the previous comment, which potentially contains outdated and inaccurate information, would only be deleted at the time a new comment is updated or posted. Continuing with your example, let's say the `terraform plan` step failed, what you'd see on PR is a failed "terraform-plan" CI check and a comment saying "here is the terraform plan" with an outdated plan.

In the PR where this `delete_pr_comment` was originated from, I initially took the "updating the existing comment" approach you mentioned, but ended up going with deleting the comment first before creating a new one, because it can produce the most accurate information.

But I'm happy to remove this `delete_pr_comment` from this plugin though, deleting a comment probably isn't a common thing we do. It should be fine just keep the script in the repo that needs it.

AliSoftware:
That's a valid point.

And to be fair that's a limitation that we already have with the comments we do on PRs of app repos to announce where to download Prototype Build for the PR: if the first CI build succeeded in building the Prototype Build, but q subsequent commit and need CI build failed to build the Prototype Build on the new commit, the old comment pointing to the old Prototype Build would not be dismissed while in practice it would not represent the reality of "the Prototype Build containing every changes that appears on this PR" (but instead, every change on this Pr except last commit that failed to build)

That being said, nothing prevents us to implement the capability for `comment_on_pr --id 'terraform-plan-aws' ""` to delete a comment when the provided comment body is empty (or when we pass `--delete` flag or whatever), and then call that as an in-place replacement of `delete_comment_pr` (ie before running `terraform plan` to ensure old comment got deleted even if `terraform plan` failed. 

It'd just be an implementation change detail (of having this feature from this `delete_pr_comment` being now implemented as `comment_on_pr --delete` instead).
But the big advantage of doing such a change being that we'd now guarantee that the logic to add the `<!-- comment-id:${COMMENT_ID} -->` hidden HTML comment to identify a GH comment would be handled directly by `comment_on_pr` instead of having the caller to do that dance, and then have the guarantee that the logic to find which comment to delete on `comment_on_pr --id xyz --delete` would rely on then same format for the `<!-- comment-id:${COMMENT_ID} -->` bit in both the use case of adding and deleting a comment (instead of having to replicate the HTML comment format assumption in the implementation of both `comment_on_pr --id someid` and `delete_pr_comment --id someid` if we were to keep those feature in separate commands/implementation like here)

Wdyt?

crazytonyli:
That sounds good to me! I'll make the change then. It didn't occur to me that we no longer need the `delete_pr_comment`, because `comment_on_pr` with an empty comment does that too.

AliSoftware:
Closing the loop: deletion of this new `delete_pr_comment` applied in #53, while the ability to delete a PR comment has been moved into the implementation of #51.

PR #56, opened 2023-05-14
Prepare for release 2.16.0

crazytonyli:
This PR updates the change log file to prepare for v2.16.0 release, which I'll create after this PR is merged.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,20 @@ _None._
 
 ### New Features
 
+_None._
+
+### Bug Fixes
+
+_None._
+
+### Internal Changes
+
+_None._
+
+## 2.16.0
+
+### New Features
+
 - Add a new `github_api` utlity to make API requests to GitHub.com or Github Enterprise instances. [#51]
 - Extend the `comment_on_pr` utility to update or delete the existing comment. [#51]
 


AliSoftware (approved):
💡  I wonder if it would be worth creating a small `Makefile` target to handle release preparation in the future, similar to [what we use in `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/Rakefile#L36-L82).

Not that it's complicated to do it manually, but just so that we are sure not to forget any step (like updating the CHANGELOG like you did here, or not having to wonder if there is a version number to update in some file somewhere or not (there isn't)…) when one does a release.

That target could ideally:
 - Parse the first section CHANGELOG to determine what type of version bump is needed based on each "New Features" has any content (major) or "Bug Fixes" has any content (minor) or patch.
 - Do the CHANGELOG dance (replace the `## Unreleased` section with a `## $newversion` title + only the non-empty `###` sub-sections that were found during parse, then add a new empty `## Unreleased` section on top
 - Create the Pull Request, with the appropriate "Release $newversion` title and instructions to create the GitHub Release and tag once it lands.

In the long run we could even consider having the CI have a `block` step on `release/* <- trunk` PRs that would allow us to { merge the PR + create GH Release } on unblocking… 🤔 

💡  And in the meantime, a quick win would be to just document that process (however simple it is, just so that we don't have to wonder how to get it right) in a `## Release Process` section of the `README.md`.

---

Anyway, those are all just random process improvement ideas for later, not blocking this PR in any way.

PR #58, opened 2023-05-24
[annotate_test_failures] Include test cases with error nodes in the failures list

adimoldovan:
## Description

`annotate_test_failures` doesn't recognize as failures the tests that have an `error` node instead of a `failure` one.
`Pytest` for example reports failures in fixtures as errors, not as failures.

More context: p1684494232589389-slack-C02K7C5K8TD

## How to test

Tested locally with a this input:

```
<?xml version="1.0" encoding="utf-8"?>
<testsuites>
  <testsuite name="pytest" errors="2" failures="2" skipped="1" tests="6">
    <testcase classname="some.class" name="test_passed">
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="some.class" name="test_with_error">
      <error message="failed on setup with &quot;TypeError: list indices must be integers or slices, not str&quot;">
        Traceback (most recent call last):...
      </error>
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="come.class" name="test_with_failure">
      <failure message="selenium.common.exceptions.TimeoutException: Message: &#10;Stacktrace:&#10;NoSuchElementError">
        Traceback (most recent call last):
      </failure>
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="some.class" name="test_skipped">
      <skipped type="pytest.skip" message="skipped test">some skip details</skipped>
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="some.class" name="test_flaky_with_error">
      <error message="failed on setup with &quot;TypeError: list indices must be integers or slices, not str&quot;">
        Traceback (most recent call last):...
      </error>
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="some.class" name="test_flaky_with_error">
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="come.class" name="test_flaky_with_failure">
      <failure message="selenium.common.exceptions.TimeoutException: Message: &#10;Stacktrace:&#10;NoSuchElementError">
        Traceback (most recent call last):
      </failure>
      <system-err>Captured Err</system-err>
    </testcase>
    <testcase classname="come.class" name="test_flaky_with_failure">
      <system-err>Captured Err</system-err>
    </testcase>
  </testsuite>
</testsuites>
```

The result was as expected, 2 failures and 2 flaky tests, with console output:

```
2 test(s) have failed (2 distinct assertion failures in total). Reporting them as a `Tests (have failed)` Buildkite error annotation.

[...]

PR #60, opened 2023-05-25
Send Slack notification for Test Failures

jostnes:
### Description
This PR updates `annotate_test_failures` with the ability to send Slack notification when there are test failures. It is an optional feature and is used by passing a second argument when calling it from project repos (I just realized via [this PR](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/58/files) when updating to latest trunk that Tumblr also uses this and they might not want the Slack notification to be sent which was the reason I added the second argument). 

The changes made for this to happen:
1. Update `update_annotation` to return a JSON object when a failure happen
2. Create `send_slack_notification` to read the JSON object, parse and form the message payload, and send a POST call to the Slack Webhook. For now, it's set specifically to `#build-and-ship` channel

In this change, the message displays the repo (to know which app) and step name (to know if it's Unit, UI iPhone or UI iPad, etc), the number of failing test cases, and the names + test classes of failing tests.

Some notes:
1. Failure notification is only sent for merges to trunk (failures on other branches will not as we are trying to get an idea how noisy this will be for trunk first)
2. When there are no failures, a notification is not sent (no notification for flaky tests)
4. If the app crashes before the JUnit XML file is created, a notification is not sent (similar to how annotation on Buildkite is not displayed when this happens currently)

### Testing
Testing PR: https://github.com/woocommerce/woocommerce-ios/pull/9798

1. When there are multiple failures ([build](https://buildkite.com/automattic/woocommerce-ios/builds/13554#018851b9-9446-43fc-a284-2f5822501058)):
image

Slack: p1685000034603039-slack-C03Q9FTCQA0

2. When there is 1 failure ([build](https://buildkite.com/automattic/woocommerce-ios/builds/13554#018851da-e85e-4e0b-9152-3d21a8161cf4)):
image

Slack: p1685001564964369-slack-C03Q9FTCQA0

3. When all test pass (with flaky tests) or one without flaky test [here](https://buildkite.com/automattic/woocommerce-ios/builds/13554#018851b9-945f-4a9b-825d-42e2949209b3):
image

5. When failure not on trunk branch:
image

6. When `should_send_slack_notification` is "false":
image

---


AliSoftware:
@jostnes FYI, given https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/58 was recently merged and also updated `annotate_test_failures` script, you might want to update your branch to make sure your changes fit nicely on top of those (I don't forsee any conflict, but just to make things clean and avoid any surprise) 🙃 

image

CHANGELOG.md
```diff
@@ -18,7 +18,7 @@ _None._
 
 ### New Features
 
-_None._
+- Update `annotate_test_failures` to be able to send Slack Notification when there are failures. [#60]
```
AliSoftware: This line was added inside the "template" (that is inside `<!-- … -->` comments) for creating a new empty changelog section after a new release has been made.

You want to add your entry to like 41 instead.

That being said, given a couple of PRs have been merged in `trunk` since you opened yours (unlucky timing!), maybe update your branch first to get the latest CHANGELOG entries, to avoid your new entry to conflict with recently-added ones.

bin/annotate_test_failures
```diff
@@ -96,16 +113,104 @@ def update_annotation(title, list, style, state)
...
+      end
+
+      test_failure_json_data = {
+        'title' => title,
+        'tests_count' => tests_count,
+        'assertion_failures_count' => assertion_failures_count,
+        'tests' => tests
+      }
+      return JSON.generate(test_failure_json_data)
+    end
```
AliSoftware: It feels odd to me that this code, which is to create a JSON, is part of the `update_annotation` method.

✋ 🛑  Besides, that added code ends up with a `return`… before the end of the method, which means that the line of code below calling `system('buildkite-agent', 'annotate', …)`… will in practice never be reached, and that would break the initial feature of annotating the Buildkite build.

---

I'd instead suggest to move that whole part of the code into its own `def generate_json(list)` method, to not mix that logic with the logic related to annotations.

Or even better: avoid that intermediate step of generating an intermediate JSON here altogether, and instead make the `send_slack_notification` method take the initial `failures` array directly and parse it there:
```ruby
update_annotation(title, failures, 'error', 'have failed')
update_annotation(title, flakies, 'warning', 'were flaky')

if should_send_slack_notification == 'true'
  send_slack_notification(title, failures)
else
  puts '⏩ should_send_slack_notification is false, skipping Slack notification.'
end
```

And then the implementation of `send_slack_notification` won't have to use some conditions on `state == 'have failed'` to only include failures—since you'll already pass it only the `failures` list when calling it, not the `flakies` list.

bin/annotate_test_failures
```diff
@@ -96,16 +113,104 @@ def update_annotation(title, list, style, state)
...
+  else
+    data = JSON.parse(failures)
+    failing_tests = data['tests'].map { |test| "#{test['name']} in #{test['classname']}" }
+    assertion_failures_count = data['assertion_failures_count']
+
+    header_text = (assertion_failures_count == 1) ? "Test" : "Tests"
+    content_text = "*Failing #{header_text}:*"
+
+    slack_message_payload = {
+      "channel": "#build-and-ship",
```
AliSoftware: It would be nice to make the target channel be customizable.

The easiest way to do that I think is to trade the 2nd parameter of `annotate_test_failures` (`ARGV[1]`) from being `true`/`false`, to instead be the name of the slack channel:
```ruby
slack_channel = ARGV[1]
if slack_channel.nil? || slack_channel.empty?
  puts '⏩ no slack channel provided, skipping Slack notification.'
else
  send_slack_notification("##{slack_channel.delete_prefix('#')}", title, failures)
end
````
_(PS: the `"##{slack_channel.delete_prefix('#')}"` trick is so that if the caller didn't include the `#` character as part of the channel name in the 2nd parameter, we make sure to add it. Given `#` is interpreted as a comment in bash, I feel it'll be more common for callers to just use `annotate_test_failures report.junit build-and-ship` rather than having to escape it like `annotate_test_failures report.junit '#build-and-ship'` or `annotate_test_failures report.junit \#build-and-ship`)_

---

The main use case I am thinking of for using a different channel is mostly Tumblr btw, since (at least as of now) Tumblr folks don't really use `#build-and-ship` but instead their own channels (`#tumblr-qa`, `#tumblr-ios-pr-build-ui-tests`… and the likes) to report failures in Slack. So it would be quite useful in practice already.

File: bin/annotate_test_failures [hunk 4]
```diff
@@ -1,10 +1,27 @@
...
+require 'json'
+
+BUILDKITE_ORGANIZATION_SLUG = ENV.fetch('BUILDKITE_ORGANIZATION_SLUG')
+BUILDKITE_PIPELINE_SLUG = ENV.fetch('BUILDKITE_PIPELINE_SLUG')
+BUILDKITE_BUILD_NUMBER = ENV.fetch('BUILDKITE_BUILD_NUMBER')
+BUILDKITE_BUILD_AUTHOR = ENV.fetch('BUILDKITE_BUILD_AUTHOR')
+BUILDKITE_PIPELINE_NAME = ENV.fetch('BUILDKITE_PIPELINE_NAME')
+BUILDKITE_JOB_ID = ENV.fetch('BUILDKITE_JOB_ID')
+BUILDKITE_BRANCH = ENV.fetch('BUILDKITE_BRANCH')
+SLACK_WEBHOOK = ENV.fetch('SLACK_WEBHOOK')
```
AliSoftware: `ENV.fetch(…)` will fail if the env var doesn't exist. Given that not all callers will want to use your new feature of sending messages to slack, they might not all have the `SLACK_WEBHOOK` env var declared in the first place if they don't really need it. But reading that value unconditionally at the start of the script will crash early anyway.

In theory you could use `ENV.fetch('SLACK_WEBHOOK', nil)` to provide a default value of `nil` instead of crashing when the env var is undefined. `ENV['SLACK_WEBHOOK']` syntax also works and also returns `nil` if undefined, so both syntaxes would be acceptable here.

In practice, I'd instead suggest to:
 - Not bother declaring those constants, and direclty use `ENV['BUILDKITE_ORGANIZATION_SLUG']` and the likes in your `send_slack_notification` method.
    - Defining those as constants doesn't feel a big win to me given all it wins you is saving 5 characters (`ENV[`+`]`)
    - It's ok to assume that all the `BUILDKITE_*` env vars are defined since this `annotate_test_failures` is made to only be called from Buildkite scripts (it doesn't even make sense to try creating a Buildkite annotation outside of the context of a Buildkite build anyway)
 - As for the `SLACK_WEBHOOK` one:
    - I'd suggest you to either only read it from inside your `send_slack_notification`, to not crash if the user didn't even intent to send a Slack notification in the first place
    - Or you could use `webhook_url = ENV['SLACK_WEBHOOK']` (it being potentially `nil`), then report that value to be missing in a more user-friendly way(e.g. `puts "Can't send the Slack notification because 'SLACK_WEBHOOK' env var is not defined." if webhook_url.nil?`) before trying to use it or pass it to `send_slack_notification`
    - But I think an even better alternative would be to allow passing that webhook URL as a parameter to the script — see [my suggestion with `OptionParser` above](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60#discussion_r1205932793), so that it can allow other projects that might declare the ENV var with a different name (e.g. some use `SLACK_BUILD_WEBHOOK` or `SLACK_URL` in their `.mobile-secrets/*/env`) to tell what to use. (We can still default to `ENV['SLACK_WEBHOOK']` if not provided explicitly though, as long as we don't make it crash if that env var is undefined)

bin/annotate_test_failures
AliSoftware: (†) At that point we might actually consider [using `OptionParser`](https://docs.ruby-lang.org/en/2.1.0/OptionParser.html) to make the parsing of the arguments and the call site not only nicer but also more flexible and explicit. Something like:

```ruby
###################
# Parse arguments
###################

slack_channel = nil
slack_webhook = ENV['SLACK_WEBHOOK'] # default value inferred from env var if not provided explicitly
args = OptionParser.new do |opts|
  opts.banner = <<~HELP
    Usage: annotate_test_failures [options] [junit-report-file-path]

    Annotates the Buildkite build with a summary of failed and flaky tests based on a JUnit report file.
    Optionally also posts the same info to a Slack channel.
  HELP
  opts.on('--slack CHANNEL_NAME', 'The name of the Slack channel to post the failure summary to') { |v| slack_channel = '#' + v.delete_prefix('#') }
  opts.on('--slack-webhook URL', 'The Slack Webhook URL to use if --slack is used. Defaults to the value of the `SLACK_WEBOOK` env var') { |v| slack_webhook = v }
  opts.on_tail("-h", "--help", "Show this help message") { puts opts; exit }
end.parse!
junit_path = args.first || File.join('build', 'results', 'report.junit')
```
Then when calling the methods at end of the script:
```ruby
if slack_channel.nil? || slack_webhook.nil?
  puts '⏩ No `--slack` channel name and/or `--slack-webhook` URL provided; skipping Slack notification.'
else
  send_slack_notification(slack_webhook, slack_channel, title, failures) unless failures.empty?
end
```

---

The nice thing about `OptionParser` is that it allows using `--` options like other typical UNIX/POSIX commands, so the call site can look like:
```bash
# No slack notification sent:
annotate_test_failures report.junit

# Use default 'build/results/report.junit' path for the JUnit report
# While still providing a channel to send a Slack notification to. Will use `ENV['SLACK_WEBHOOK']` for the webhook URL
annotate_test_failures --slack 'build-and-ship'

# Provide both a JUnit report path and a slack channel, relying on SLACK_WEBHOOK env var to be defined
annotate_test_failures build/test-report.junit --slack 'build-and-ship'

# Use default `build/results/report.junit' path for the JUnit report,
# And also provide all the info for Slack notification, including a custom webhook URL defined in a different env var
annotate_test_failures --slack 'build-and-ship' --slack-webhook "$SLACK_BUILD_URL"

# Print the help/documentation
annotate_test_failures -h
```
_Note: I kept the optional path to the JUnit report as a separate argument rather than making it a `--report PATH` option like the rest, just for backwards compatibility with the existing way people call `annotate_test_failures`._


wdyt?

AliSoftware (changes requested):
My main concern is that your current implementation `return` from the `update_annotation` method before reaching `system('buildkite-agent', 'annotate', …)`, so that will break the existing behavior. Another concern is the code crashing if `SLACK_WEBHOOK` env var is not defined (which will likely be the case for many pipelines other than Tumblr-Android and WCiOS).
Hence why the "Request changes" review.

The rest of the comments are mostly suggestions to make the code much nicer (less convoluted and hopefully easier to read) and more flexible. ✨ 😉

CHANGELOG.md
jostnes: hah very unlucky! thanks for pointing this out, should be updated in 4fb1c76f19868e6b66c8c36c937a39652415a93c

bin/annotate_test_failures
jostnes: good points, i've decided to use your second suggestion to send the failures list value directly to `send_slack_notification()` and parse it from there. i realized the change also made the code simpler. changes made in 2a4c6a7bd7351bd0747ccd255ee427ee84a50173

bin/annotate_test_failures
jostnes: thanks for sharing about `OptionParser`! i had a look and tried it out and yeah it's much more flexible as the slack channel can now be any channel and not just the one defined in the payload, how it's called in the command line is also clearer as slack is explicitly mentioned. tested out different scenarios (like in your example, and with the `#` in the channel name) and looks to be working as expected. 

changes made in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/351c7348152a3ddad0aa0c0d97ffc05a79b0734a

bin/annotate_test_failures
jostnes: yeah makes sense, updated the new env vars to this format `ENV['ENV_VAR_NAME']` as part of the same commit for the OptionParser change in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/351c7348152a3ddad0aa0c0d97ffc05a79b0734a

jostnes:
thanks for the review @AliSoftware ! i made the suggested changes and tested with [this CI build](https://buildkite.com/automattic/woocommerce-ios/builds/13579#01885643-67af-41c0-abf6-5b5b8631f3f8) and locally, it's ready for another review.

though i did notice one thing - the main branches name are different on Tumblr repos. i made this update https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/f89ed065dfb7e31cd51d5713fa1555ddc55fdc32 to also check if the build is on develop or master. wdyt?

bin/annotate_test_failures
```diff
@@ -100,6 +115,76 @@ def update_annotation(title, list, style, state)
...
+#
+def send_slack_notification(slack_webhook, slack_channel, title, list)
+  tests = list.map do |item|
+    {
+      'name' => item.name,
+      'classname' => item.classname,
+      'message' => item.message,
+      'details' => item.details
+    }
+  end
```
AliSoftware: I'm not sure I'm seeing the point of converting the array of `TestFailure` objects into an array of Hashes/Dictionaries like that, only to have to use `test['name']` and the likes to access each Hash's key in the code below… instead of directly looping on `list` on the code below then using `test.name` to access each TestFailure's property directly?

I'm guessing this idea is more of a leftover from the fact that your previous implementation was using an intermediate json? But now that it's not and directly receives the array of `TestFailure` objects I think you can get rid of those lines, and have the rest of the code directly use the `list` variable instead of `tests`

bin/annotate_test_failures
```diff
@@ -100,6 +115,76 @@ def update_annotation(title, list, style, state)
...
+
+  if !["trunk", "develop", "master"].include?(ENV['BUILDKITE_BRANCH']) # Don't send notification if build is not on trunk/develop/master branch
+    puts "⏭  Failures not on trunk/develop/master branch. Slack notification not sent."
+  else
```
AliSoftware: This definitively shouldn't be hardcoded here. This tool is a general-purpose utility that might be called by any BuildKite project that generates Junit results. We can't assume which repos will be using it and which branches the users of that script will want to enable or disable Slack notifications for.

For example, maybe at some point some of our pods repos like WPAuthenticator and friends, or even a brand new repo, will like to use this utility. There's no guarantee that those other repos will use those branch names… nor that we'd want the Slack notification not to be sent for them on trunk (in fact, I can totally see some repos willing to send the notification on test failures even for builds on trunk)

Therefore that decision to not send the Slack notification on certain conditions (like branch betting trunk or not) should be made by the callers of this script, not hardcoded in the scripts itself. For example WCiOS could decide to use:
```bash
if [[ "$BUILDKITE_BRANCH" =~ ^(develop|trunk)$ ]]; then
  annotate_test_failures
else
  annotate_test_failures --slack 'build-and-ship'
fi
```

bin/annotate_test_failures
jostnes: oops yeah was a copypasta from the previous implementation without considering rewriting, thanks for catching this, removed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/b36024e676bd41372c1134f30c31875fb3ba3812#

bin/annotate_test_failures
jostnes: thanks for the input! i wasn't sure how to go about this and was too fixated on having the change in this repo i did not consider having it decided on the caller repo. this makes sense and is a better way. condition to check the branch removed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/b36024e676bd41372c1134f30c31875fb3ba3812#diff-93aca9db104a198497cb12c9b10d00804f941c566994d7683548ed14cc113610L130-L131

jostnes (commented):
Removed the branch condition and tested using two WCiOS testing builds:
1. [Slack notification sent](https://buildkite.com/automattic/woocommerce-ios/builds/13582#01885706-ec09-42ae-9555-a7b95f10e914/6097-6114)
2. [Slack notification not sent](https://buildkite.com/automattic/woocommerce-ios/builds/13586#0188574b-3c91-4a3a-98ff-23c1c7fee600/6173-6190)

jostnes (commented):
Removed the branch condition and tested using two WCiOS testing builds:
1. [Slack notification sent](https://buildkite.com/automattic/woocommerce-ios/builds/13582#01885706-ec09-42ae-9555-a7b95f10e914/6097-6114)
2. [Slack notification not sent](https://buildkite.com/automattic/woocommerce-ios/builds/13586#0188574b-3c91-4a3a-98ff-23c1c7fee600/6173-6190)

jostnes (commented):
Removed the branch condition and tested using two WCiOS testing builds:
1. [Slack notification sent](https://buildkite.com/automattic/woocommerce-ios/builds/13582#01885706-ec09-42ae-9555-a7b95f10e914/6097-6114)
2. [Slack notification not sent](https://buildkite.com/automattic/woocommerce-ios/builds/13586#0188574b-3c91-4a3a-98ff-23c1c7fee600/6173-6190)

bin/annotate_test_failures
```diff
@@ -100,6 +115,63 @@ def update_annotation(title, list, style, state)
...
+          }
+        },
+        {
+          "type": "divider"
+        },
+        {
+          "type": "section",
+          "text": {
+            "type": "mrkdwn",
+            "text": "*Failing #{test_text}:* #{failing_tests.to_s[1..-2]}\n" # To remove the bracket from the notification and only list the tests
```
AliSoftware: Instead of using `.to_s` to convert the array into its debug representation (that will contain the brackets), I'd suggest you to use `.join(',')` to join the items of the array into a nice-looking list.

You could even decide to join the list of failing tests using `.join("\n")` instead if you prefer them to be rendered each on their own line, In which case I'd advocate to also add `-` in front of each entry to make it look like a bullet-point list in Slack 🙂 And while we're at prettifying all this, you can also use backticks around the test and class name too 🙃 

```ruby
failing_tests = list.map { |item| " - `#{item.name}` in `#{item.classname}`" }
```
```ruby
   "text": "*Failing #{test_text}:*\n#{failing_tests.join("\n")}"
```

bin/annotate_test_failures
```diff
@@ -100,6 +115,63 @@ def update_annotation(title, list, style, state)
...
+          }
+        }
+      ]
+    }
+
+    json_payload = JSON.generate(slack_message_payload)
+
+    # Send message to Slack
+    uri = URI("#{slack_webhook}")
```
AliSoftware: Nitpick: no need to use string interpolation to interpolate… a variable that is already the whole string you need 😃 

```suggestion
    uri = URI(slack_webhook)
```

bin/annotate_test_failures
```diff
@@ -100,6 +115,63 @@ def update_annotation(title, list, style, state)
...
+          },
+          "accessory": {
+            "type": "button",
+            "text": {
+              "type": "plain_text",
+              "text": "Build",
+              "emoji": true
+            },
+            "value": "build",
+            "url": "https://buildkite.com/#{ENV['BUILDKITE_ORGANIZATION_SLUG']}/#{ENV['BUILDKITE_PIPELINE_SLUG']}/builds/#{ENV['BUILDKITE_BUILD_NUMBER']}##{ENV['BUILDKITE_JOB_ID']}",
```
AliSoftware: The whole first part of that URL is [already also exposed by Buildkite as `BUILDKITE_BUILD_URL` env var](https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_BUILD_URL), so you might as well use this to make this more readable 😉 

```suggestion
            "url": "#{ENV['BUILDKITE_BUILD_URL']}##{ENV['BUILDKITE_JOB_ID']}",
```

bin/annotate_test_failures
```diff
@@ -1,16 +1,31 @@
...
 ###################
-
-junit_path = ARGV.first || File.join('build', 'results', 'report.junit')
+slack_channel = nil
+slack_webhook = ENV['SLACK_WEBHOOK'] # default value inferred from env var if not provided explicitly
+args = OptionParser.new do |opts|
+  opts.banner = <<~HELP
+    Usage: annotate_test_failures [options] [junit-report-file-path]
+
+    Annotates the Buildkite build with a summary of failed and flaky tests based on a JUnit report file.
```
AliSoftware:
```suggestion
    Annotates the Buildkite build with a summary of failed and flaky tests based on a JUnit report file (defaults to using `build/results/report.junit`).
```

bin/annotate_test_failures
```diff
@@ -1,16 +1,31 @@
 #!/usr/bin/env ruby
 #
 # Usage:
-#   annotate_test_failures [junit-file-path]
+#   annotate_test_failures [options] [junit-report-file-path]
```
AliSoftware, line 4:
```suggestion
#   annotate_test_failures [junit-report-file-path] [--slack CHANNEL] [--slack-webhook URL] 
```

AliSoftware (approved):
Left some final nitpick touches/suggestions, otherwise this looks really nice and ready now 🎉 

I'll approve it already to unblock, and trust you to properly apply the final suggestions and test them one last time before merging if you like them, without needing another approval round from me by now 🙂

bin/annotate_test_failures
jostnes: thanks for the re-review! 

agreed and made all the changes in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/60/commits/32ef08fe69d5fd0c2f5b0de90ddac3fcd0c08856 except for displaying them as a bullet point list. found that Slack message, when sent by the block API, only covers half the screen ([looks like a known issue](https://stackoverflow.com/questions/57438097/how-to-make-slack-api-block-kit-take-up-the-entire-width-of-the-slack-window)) and because some test names can be pretty long, for those, the message doesn't look that great (screenshot below from a local test, which is why the default label "Tests" is displayed):

image

jostnes:
Merging this, the latest changes were tested using this WCiOS build ([notification sent](https://buildkite.com/automattic/woocommerce-ios/builds/13605#01886553-e754-435a-a5ad-d4c3d5c4b230/10172-10198), [notification not sent](https://buildkite.com/automattic/woocommerce-ios/builds/13605#01886553-e760-42fb-865d-81d1fd3ee7fa/6114-6131)) and works as expected.

PR #59, opened 2023-05-25
Add `git diff` to `Podfile.lock` diverged annotation

mokagio:
I was baffled today by a test build failing because the `Podfile.lock` changed in CI even though running `bundle exec pod install` locally didn't make it change for me.

I found it useful to print the diff that triggered the failure, which revealed the code I wrote was printing absolute paths in the `Podfile.lock`.

[In action](https://buildkite.com/automattic/wordpress-ios/builds/14343#0188516d-6cf8-4f71-b295-9e0cfe440414):
 
image

image

---

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #61, opened 2023-05-29
Add `diff` language specification to `Podfile.lock` diff annotation

mokagio:
In #59, @AliSoftware suggested adding `diff` to the code block in the `Podfile.lock` diverged failure annotation, to see if Buildkite also adds syntax highlighting to the rendered Markdown.

I implemented it here and made a test PR [on Woo iOS](https://github.com/woocommerce/woocommerce-ios/pull/9832). It doesn't look like there's any syntax highlighting in the annotation 😞 

image

On the other hand, the `diff` in the code block is not rendered in the code, which makes me hopeful. Or, at least, it makes it look like an harmless change to add.

---

### DIFF
--- a/bin/install_cocoapods
+++ b/bin/install_cocoapods
@@ -37,7 +37,7 @@ function lockfile_error () {
 
   <details><summary>See diff</summary>
 
-  \`\`\`
+  \`\`\`diff
   $(git diff -- Podfile.lock)
   \`\`\`
   </details>


AliSoftware (approved):

mokagio:
@AliSoftware super fast review. Thanks! So fast, in fact, that I think it arrived before the edits I made in the description in regards to the lack of syntax highlighting.

I think you'll agree with my conclusion that it's okay to merge anyway, but leaving a note about it, just in case.

AliSoftware:
Unfortunate that there's no syntax highlighting after all, but doesn't hurt to merge anyway 👍

PR #62, opened 2023-05-29
Update `CHANGELOG.md` for version 2.17.0

mokagio:
Just a breadcrumb PR to track the release of a new version. cc @jostnes this should be useful for you, right?

Note that I haven't pushed the matching tag yet, so we're still in time to cancel this and add more changes if necessary.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. 😅

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,16 +38,26 @@ _None._
 
 ### New Features
 
+_None._
+
+### Bug Fixes
+
+_None._
+
+### Internal Changes
+
+_None._
+
+## 2.17.0
+
+### New Features
+
 - When `install_cocoapods` fails because `Podfile.lock` changed in CI, it now prints a diff of the changes [#59]
 - Update `annotate_test_failures` to be able to send Slack Notification when there are failures. [#60]
 
 ### Bug Fixes
 
 - Fix the `annotate_test_failures` to include test cases with error nodes in the failures list. [#58]
-
-### Internal Changes
-
-_None._
 
 ## 2.16.0
 


crazytonyli (approved):

AliSoftware:
👍 

> Note that I haven't pushed the matching tag yet, so we're still in time to cancel this and add more changes if necessary.

I think it's still worth doing the tag and thus proper release. If we want to add more stuff, we can do a `2.18.0`, it's not costing us anything much to do smaller, more frequent releases rather than waiting too long to fit more stuff.

The only drawback of doing more releases is that people will have to update their `.buildkite/*.yml` files more frequently to get the latest tag if they want it. But if one wants to wait for more stuff and only update from `2.16.0` to `2.18.0` once this one will go out, skipping `2.17.0`, that's all OK too, so better give the option to the client repos of this plugin (to adopt more granular releases or not and skip some intermediate versions), than have them being forced to wait?

mokagio:
Thanks @AliSoftware 

I'm not sure about the timeline of the events, but just wanted to say that after merging I did indeed tag and ship even though I didn't see your message.

image

_Not sure why the merge events is rendered before the comment, even though their timestamps suggest the opposite order._

I wanted to clear that out for reference and because your rationale resonates with me.

AliSoftware:
It's also very possible that the timeline of events is correct, eg I opened your PR in a tab, then got distracted, meanwhile Tony approved it and you merged it, and only then did I get back to my open tab and submitted my review without realizing that the PR have already been reviewed and merged since?

Or that I saw that the PR was already merged but still decided to comment to suggest you to do the tag and release (because that part in your PR suggestion was mentioning that you were not sure you'd create the tag as soon as the PR got merged vs if you'd do it later in case we wanted other changes to go in first)…

jostnes:
thank you for wrangling this release @mokagio 🙇

PR #64, opened 2023-06-08
Example `rspec` tests

mokagio:


Gemfile
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+source "https://rubygems.org"
+
+gem "rexml"
```
mokagio, line 5: We'd probably want to use nokogiri if we pushed this to "production" as it's the more commonly used, but this does the job alright in this context.

Makefile
```diff
@@ -1,14 +1,17 @@
...
 
 lint: buildkite-plugin-lint rubocop shellcheck
-test: buildkite-plugin-test buildkite-plugin-test-commands-are-executable
+test: unit-tests buildkite-plugin-test buildkite-plugin-test-commands-are-executable
 
 buildkite-plugin-test:
 	docker run -t --rm -v "${PWD}":/plugin buildkite/plugin-tester
 
 buildkite-plugin-test-commands-are-executable:
 	docker run -t --rm -v "${PWD}":/plugin -w /plugin ruby:2.7.4 /bin/bash -c "gem install --silent rspec && rspec tests/test-that-all-files-are-executable.rb"
```
mokagio, line 10: I didn't change this Ruby value to the one from the new `.ruby-version` to keep this PR focused.

.ruby-version
```diff
@@ -0,0 +1 @@
+3.2.2
```
mokagio, line 1: Given this is an experiment, might as well try with the latest Ruby :)

mokagio:
Closing as stale.

PR #70, opened 2023-07-18
Fix `install_gems`'s `CACHE_KEY` when using `RBENV_VERSION`

AliSoftware:
## Why

When the ruby version is forced by a mean _other_ than the local `.ruby-version` file—especially when it is set via `export RBENV_VERSION` instead—then the `RUBY_VERSION` variable that is used to build the `CACHE_KEY` picks up the wrong ruby version value.

This is in particular the case [in the `release-toolkit` pipeline](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/.buildkite/pipeline.yml#L30), which uses `export RBENV_VERSION={{ matrix.ruby }}` to test our codebase against multiple ruby versions as part of a build matrix. In such cases, the `.ruby-version` file is unchanged and still contains `2.7.4` (as the ruby version we expect developers to use locally when they work on the `release-toolkit`), but the CI step dynamically change it via the env var instead.

## How

Make `RUBY_VERSION` use `rbenv version-name` to determine the ruby version instead. We still fallback to `cat .ruby-version` as a precaution, in case `rbenv` were not installed on the machine running that code.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -43,6 +43,7 @@ _None._
 ### Bug Fixes
 
 - Ensure that `install_gems` doesn't restore a macOS cache on Linux build machines. [#69]
+- Ensute that `install_gems` uses the correct ruby version in its `CACHE_KEY` in all cases [#70]
 
 ### Internal Changes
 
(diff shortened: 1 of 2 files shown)


jkmassel (approved):
Seem like a sensible change, and AFAICT it'll work fine with `rvm` and other ruby version managers

PR #72, opened 2023-08-16
Pass --no-use to postpone using nvmrc node

crazytonyli:
See pdnsEh-1g5-p2#comment-2567 for context.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,7 +42,7 @@ _None._
 
 ### Bug Fixes
 
-_None._
+* Do not use the node version specified in `.nvmrc` until it's installed. [#72]
 
 ### Internal Changes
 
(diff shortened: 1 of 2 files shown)


crazytonyli:
I just realised it doesn't seem to make sense to call `nvm install` as part of a executable shell script. It has to be sourced. Otherwise, nvm can't change the environment of the project's CI job process.

I think we'll need to use job lifecycle hooks (`environment` or `pre-command`) to make nvm work.

AliSoftware:
> I just realised it doesn't seem to make sense to call `nvm install` as part of a executable shell script. It has to be sourced. Otherwise, nvm can't change the environment of the project's CI job process.
> 
> 
> 
> I think we'll need to use job lifecycle hooks (`environment` or `pre-command`) to make nvm work.

The way this has been solved in the past iinm is by relying on [our `install_nvm` utility from `a8c-ci-toolkit`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/nvm_install).

PR #73, opened 2023-08-23
Remove nvm_install and recommend using nvm-buildkite-plugin

crazytonyli:
The `nvm install|use` command achieves switching node versions by updating its running process's `PATH` environment. Calling `nvm_install` would only change the node version used by the `nvm_install` shell process, not the pipeline's process.

In order to change the node.js version used by pipeline's process, we'll need to call the `nvm install|use` by `source`. And that's what https://github.com/Automattic/nvm-buildkite-plugin/pull/3 does.

**Update**: Instead of error with "command not found" when the utility is called, it's probably better to print a message about migration. So I added the script back, which now fails with an more helpful message.

---


bin/nvm_install
```diff
@@ -1,12 +1,6 @@
-#!/bin/bash -eu
+#!/bin/bash
 
-# shellcheck source=/dev/null
-source "$HOME"/.nvm/nvm.sh
+echo 'The nvm_install utility is removed.'
```
AliSoftware:
```suggestion
echo 'The nvm_install utility has been removed and replaced by our nvm Buildkite plugin.'
```

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- `nvm_install` utility is removed. See [MIGRATION.md](./MIGRATION.md#from-200-to-300) for details.
```
AliSoftware, line 37: I'm not an native English speaker, but I don't know why, the "is removed" wording sounds odd to me? 🤷 
```suggestion
- `nvm_install` utility has been removed. See [MIGRATION.md](./MIGRATION.md#from-200-to-300) for details.
```

AliSoftware (approved):

CHANGELOG.md
crazytonyli: Thanks for the suggestions. Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/73/commits/679f919cb64e8a8ec02aadb442d88c3c96104c51

PR #74, opened 2023-09-06
Refine message in `annotate_test_failures` when optional config missing

mokagio:
I configured Slack reporting in https://github.com/wordpress-mobile/gutenberg-mobile/pull/6134 based on what I saw in WordPress iOS and was surprised by the info message I got. _Why did that string work in WP but not on my project_? I turns out that WP set the webhook via an env var. Admittedly, It didn't take long to find that out by looking at the code. Still, I decided to update the "Slack skipped" message to something I think would have been clearer for me at the time.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-_None._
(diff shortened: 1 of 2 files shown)


AliSoftware (approved):

mokagio:
Thanks @AliSoftware. That was _fast_! 😄

PR #75, opened 2023-09-29
install_swiftpm_dependencies: prevent Package.resolved from being updated

AliSoftware:
## Why?

This came up in a Slack discussion (p1696008080699129/1696006524.465569-slack-C02KLTL3MKM) and also [in the Swift forums](https://forums.swift.org/t/xcode-ignoring-package-resolved-file/65031/6).

When calling `xcodebuild -resolvePackageDependencies`, Xcode might happily update the `Package.resolved` file instead of treating it like a lockfile (like other dependency managers do).

Basically, to my understanding, `xcodebuild -resolvePackageDependencies` is akin to `pod update`, in the sense that it might not just resolve the dependencies but also update them… while if you only want the equivalent of `pod install` aka install exactly what's in the `Package.resolved` without trying to _update_ any dependency, you have to add this `-onlyUsePackageVersionsFromResolvedFile` flag explicitly?

This has caused issue in our CI builds on Tumblr-iOS after recently introducing a call to `install_swiftpm_dependencies` in our pipeline, because some of our lanes also check `ensure_git_repo_clean` and they were thus failing due to that utility causing a change on the `Package.resolved` file.

## Testing

I believe, based on what I've been reading in the Swift forums and my understanding of the problem, that this change would fix that issue and prevent `Package.resolved` from being updated when we don't want to, but haven't actually tested it.

Just making that PR so we don't forget about it, but we should probably first test it by pointing Tumblr-iOS's `a8c-ci-toolkit` plugin to the branch of this PR, then restore our calls to `ensure_git_repo_clean` in our CI lanes, and check if that it indeed prevents the `Package.resolved` from being changed and failing the check.

---

### DIFF
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -40,7 +40,7 @@ add_host_to_ssh_known_hosts github.com
 # Based on the project type, resolve the packages using the correct build type
 if [[ "$BUILD_TYPE" == "XCODEBUILD" ]]; then
   echo "	Resolving packages with \`xcodebuild\`"
-  xcodebuild -resolvePackageDependencies
+  xcodebuild -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
 elif [[ "$BUILD_TYPE" == "SWIFT" ]]; then
   echo "	Resolving packages with \`swift\`"
   swift package resolve


crazytonyli (approved):

mokagio (approved):
Thank you!

AliSoftware:
@crazytonyli @mokagio Just curious, did any of you tested this?

I opened this PR mainly to not forget to fix this but the solution I adopted in this PR was made directly via the GitHub web UI and only based on what I read in the forums, so I didn't have the opportunity to test if that would properly solve the issue and/or if this is how we're supposed to use the flag (e.g. should we remove the `-resolvePackageDependencies` flag when we use this new `-onlyUsePackageVersionsFromResolvedFile`, or are they supposed to be used together as I did in the PR?)

Also, I wonder if this issue also applies to `swift package resolve` (for non-Xcode project using `Package.swift` and `swift` directly), and if so what's the corresponding flag to use for `swift package`… or if the "bug"/unexpected behavior _only_ applies to calls to `xcodebuild`?

AliSoftware:
Reading `swift package -h` help page, I see this in the options:
```
  --force-resolved-versions, --disable-automatic-resolution, --only-use-versions-from-resolved-file
                          Only use versions from the Package.resolved file and fail resolution if it is out-of-date
  --skip-update           Skip updating dependencies from their remote during a resolution
```

Which makes me think that for `swift package resolve` we probably don't want `--only-use-versions-from-resolved-file` (which is the one that most resembles `xcodebuild`'s `-onlyUsePackageVersionsFromResolvedFile` flag) if it's gonna fail if out of date, and instead we might want `--skip-update` instead?

Though on the other hand, this is the help for `swift package` as a whole. That command has two separate subcommands `swift package resolve` and `swift package update`, which from their name alone I'd assume the former is akin to `pod install` and shouldn't update the versions nor the `.resolved` file while the latter would be like `pod update` and update the `.resolved` file? Also, `swift package resolve -h` doesn't show flags like `--skip-update` or `--only-use-versions-from-resolved-file` in its help.
But I don't really trust that 100%, given that `xcodebuild -resolvePackageDependencies`, despite the name of the flag, does more than just resolving but also updates… so that makes me wonder if `swift package …` subcommands don't have a similar behavior… like maybe the term "resolve" is used loosely in both cases not just in `xcodebuild`'s case? 🤔

AliSoftware:
[This part of the SPM documentation](https://github.com/apple/swift-package-manager/blob/main/Documentation/Usage.md#resolving-versions-packageresolved-file) is encouraging regarding `swift package resolve` vs `swift package update` though

AliSoftware:
### Trying to test and repro the behavior

Ok I tried testing the behavior in https://github.tumblr.net/TumblrMobile/orangina/pull/26229 but couldn't make `xcodebuild` update the `Package.resolved` file during `xcodebuild -resolvePackageDependencies`, even after [artificially reverting one of the package (`test-collector-swift`) from `0.4.0` (latest available version) to `0.3.0`](https://github.tumblr.net/TumblrMobile/orangina/pull/26229/files#diff-9755293ffc62d2f15be77bf4d9763a694df8776b17f394fe0406fc11e314fa74), then [set the project's packages dependency to `upToNextMajorVersion 0.3.0`](https://github.tumblr.net/TumblrMobile/orangina/pull/26229/files#diff-cb5f3529c6b992b95dd7d75289d5113eeac734a3f1b57a2362846ec135b57901).

Running `xcodebuild -resolvePackageDependencies`, **without** the `-onlyUsePackageVersionsFromResolvedFile` flag, all while the package was back at `0.3.0`, both locally and on CI, didn't make it update the package to `0.4.0`.

Either this is because this issue has been solved and behavior changed in Xcode 15.0 (though I then tried with `DEVELOPER_DIR=/Applications/Xcode-14.3.1.app xcodebuild -resolvePackageDependencies` and didn't manage to make it update the `Package.resolved` either), or there's something else that makes this auto-update a bit random with some rules to trigger it that I quite don't get.

### `onlyUsePackageVersionsFromResolvedFile` will fail in case of incompatible constraints

One thing to note is that when using `xcodebuild -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile`, if the `Package.resolved` file cannot match the dependencies constraints set in the Xcode project (e.g. `Package.resolved` points to a version `0.3.0` of the package but the Xcode project is set to `upToNextMajorVersion 0.4.0`, meaning that the `Package.resolved` doesn't fit the constraints anymore, then the `xcodebuild -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile` will error with the following message:

```
xcodebuild: error: Could not resolve package dependencies:
  an out-of-date resolved file was detected at /Users/olivierhalligon/Documents/Dev/apps/Tumblr-iOS/Orangina.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved, which is not allowed when automatic dependency resolution is disabled; please make sure to update the file to reflect the changes in dependencies. Running resolver because requirements have changed.
  fatalError
```

But I think that's actually a good thing, because that means we have committed an inconsistent state between the `Package.resolved` and the Xcode project and this should indeed be fixed.

### Multiple `Package.resolved` files (workspace vs project)

One gotcha about this is that I noticed that sometimes there might be 2 `Package.resolved` — one in `App.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` and the other in `App.xcworkspace/xcshareddata/swiftpm/Package.resolved` and they might not always be in sync (I think `xcodebuild -resolvePackageDependencies` might update one while Xcode's IDE might update the other, depending on what was changed and what project/workspace was opened?) — so that could lead to inconsistencies there.

For example, in Tumblr those two `Package.resolved` were out of sync with one another. Which might have been the source of the original issue after all (where the `Package.resolved` manipulated by Xcode IDE was up-to-date, but the one manipulated by `xcodebuild -resolvePackageDependencies` was different and out of date, leading to it being updated as part of running this utility, then leading to `ensure_git_repo_clean` failing…

@mokagio @crazytonyli @spencertransier What do you think about that last point?

 - I wonder if it's worth investigating deeper which `Package.resolved` file (the workspace one vs the project one) Xcode IDE is updating (since developers are more likely to interact with the IDE to add or update package version constraints) and if `xcodebuild -resolvePackageDependencies` might accidentally be based on the other one.
 - Or maybe we should make this `install_swiftpm_dependencies` utility more resilient to this kind of possible discrepancy between workspace and project resolved files… though I'm guessing there might be cases where the resolved file of the workspace might include the union of all the packages used in the pbxproj projects that the workspace is referencing, making it being different be a correct behavior (even though in most of our cases our workspaces just have the main project + Pods.xcodeproj, the latter never having swiftpm dependencies… so we might never have hit that case… but still should handle the possibility)

AliSoftware:
Merging because there has been no activity in the last week and I figure the changes made in that PR can't hurt, and the separate issue/open question of which `Package.resolved` is being updated by the command (workspace vs project) can be investigated and handled in a separate issue and PR.

PR #76, opened 2023-10-20
annotate_test_failures: fix nil crash

AliSoftware:
I encountered a crash on `@details` being `nil` in [this build](https://buildkite.com/automattic/tumblr-android/builds/20404#018b4c6a-3553-45bc-af43-f75dc07031ed/351-352) so I figured I'd make a quick fix for it.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-- `nvm_install` utility is removed. See [MIGRATION.md](./MIGRATION.md#from-200-to-300) for details.
+- `nvm_install` utility is removed. See [MIGRATION.md](./MIGRATION.md#from-200-to-300) for details. [#73]
 
 ### New Features
 
@@ -42,18 +42,19 @@ _None._
 
 ### Bug Fixes
 
-_None._
+ - Fix issue in `annotate_test_failures` where the error details' XML node could sometimes be `nil`. [#76]
+ - Make `install_swiftpm_dependencies` use `-onlyUsePackageVersionsFromResolvedFile` flag, to prevent `Package.resolved` from being modified on CI. [#75]
 
 ### Internal Changes
 
-- Refine message in `annotate_test_failures` when configuration for optional Slack reporting is missing [#74]
+- Refine message in `annotate_test_failures` when configuration for optional Slack reporting is missing. [#74]
 
 ## 2.18.2
 
 ### Bug Fixes
 
 - Ensure that `install_gems` doesn't restore a macOS cache on Linux build machines. [#69]
-- Ensute that `install_gems` uses the correct ruby version in its `CACHE_KEY` in all cases [#70]
+- Ensure that `install_gems` uses the correct ruby version in its `CACHE_KEY` in all cases. [#70]
 
 ## 2.18.1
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -34,26 +34,27 @@ _None._
...
 
 ### New Features
 
 _None._
 
 ### Bug Fixes
 
-_None._
+ - Fix issue in `annotate_test_failures` where the error details' XML node could sometimes be `nil`. [#76]
+ - Make `install_swiftpm_dependencies` use `-onlyUsePackageVersionsFromResolvedFile` flag, to prevent `Package.resolved` from being modified on CI. [#75]
```
AliSoftware, line 46: Note: I forgot to add the CHANGELOG entry in #75 so I took the opportunity of this PR to add it here as well.

oguzkocer (approved):

Twinsen81:
thanks for fixing that @AliSoftware! When will this change become effective? I [still see the error](https://buildkite.com/automattic/tumblr-android/builds/20487#018b6182-aec7-429b-9fa3-f99a26312dd6/346-347) in the latest builds.

AliSoftware:
@Twinsen81 I can make a new release, though note that because the [we also landed an (unrelated) breaking change](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/CHANGELOG.md#breaking-changes) in `trunk` the next version will have to be a major version bump `3.0.0`. This means that if you want to adopt that new version and your pipeline were using `nvm_install` in any CI step, you will have to make sure to [follow the migration instructions](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/MIGRATION.md#from-200-to-300) to migrate to the new way of using nvm as you adopt `a8c-ci-toolkit 3.0.0` even if all you wanted was to get that `annotate_test_failures` fix.

Twinsen81:
I didn't find any `nvm_install` calls in our repo. Does that mean I just need to do [this](https://github.com/Automattic/nvm-buildkite-plugin#example) to complete the migration?
```
plugins:
      - automattic/nvm#0.2.1:
          version: 'v18'
```

AliSoftware:
I see that  `nvm_install` is indeed not used in Tumblr-Android repo, and that it seems you already migrated it to use `automattic/nvm` plugin instead 👍 

Which means that the fact that `nvm_install` command will be removed as part of `a8c-ci-toolkit#3.0.0` won't affect Tumblr-Android, as it doesn't use it anymore anyway.

So all you'll have to do once I release `a8c-ci-toolkit#3.0.0` is to update all the `.buildkite/*.yml` files to use `a8c-ci-toolkit#3.0.0` everywhere it was referencing `a8c-ci-toolkit#2.18.1`, and that should be enough to get the fix for `annotate_test_failures` without breaking anything in the repo.

That being said, it won't hurt to also update the other plugins while we'll be at it, like indeed update our current references to `automattic/nvm#0.1.0` in Tumblr-Android to the latest version `0.2.1` of the nvm plugin instead (We already use a `.nvmrc` file at the root of Tumblr-Android repo, so no need to add that `version: 'v18'` parameter to the nvm plugin declaration though)

AliSoftware:
[Release 3.0.0 has been created](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/tag/3.0.0) and includes this fix, so you can start referencing it in your repos 👍

Twinsen81:
Awesome! Thanks!

PR #78, opened 2023-11-28
Add a hack to patch cocoapods 

crazytonyli:
This hack unblocks us from releasing libraries from CI. Hopefully CocoaPods can fix the issue properly and we can remove this hack soon.

This change is tested in [this build](https://buildkite.com/automattic/wordpresskit-ios/builds/1151).

---

### DIFF
--- a/bin/patch-cocoapods
+++ b/bin/patch-cocoapods
@@ -44,4 +44,4 @@ fi
 
 echo "Failed to patch cocoapods. Try to re-apply the patch"
 echo "$patch_content" | patch --reverse --force --directory "$cocoapods_gem_path" --strip 0
-echo "$patch_content" | patch --forward --force --directory "$cocoapods_gem_path" --strip 0;
+echo "$patch_content" | patch --forward --force --directory "$cocoapods_gem_path" --strip 0
--- a/bin/patch-cocoapods
+++ b/bin/patch-cocoapods
@@ -1,0 +1,47 @@
+#!/bin/bash
+
+# This is a hack to workaround https://github.com/CocoaPods/CocoaPods/issues/12033.
+# We can remove this script once the issue is fixed.
+#
+# This script updates the cocoapods source code to change Xcode project targets'
+# minimal deployment target to 13.0, which is a requirement in newer Xcode versions.
+
+set -euo pipefail
+
+patch_content=$(cat <<'EOF'
+--- lib/cocoapods/installer/analyzer.rb
++++ lib/cocoapods/installer/analyzer.rb
+@@ -857,7 +857,7 @@
+           Version.new(library_spec.deployment_target(platform_name) || default)
+         end.max
+         if platform_name == :ios && build_type.framework?
+-          minimum = Version.new('8.0')
++          minimum = Version.new('13.0')
+           deployment_target = [deployment_target, minimum].max
+         end
+         Platform.new(platform_name, deployment_target)
+--- lib/cocoapods/validator.rb
++++ lib/cocoapods/validator.rb
+@@ -566,7 +566,7 @@
+     def deployment_target
+       deployment_target = spec.subspec_by_name(subspec_name).deployment_target(consumer.platform_name)
+       if consumer.platform_name == :ios && use_frameworks
+-        minimum = Version.new('8.0')
++        minimum = Version.new('13.0')
+         deployment_target = [Version.new(deployment_target), minimum].max.to_s
+       end
+       deployment_target
+EOF
+)
+
+cocoapods_gem_path=$(bundle info cocoapods --path)
+echo "Path to the cocoapods gem: $cocoapods_gem_path"
+
+if echo "$patch_content" | patch --forward --force --directory "$cocoapods_gem_path" --strip 0; then
+    echo "cocoapods patched successfully"
+    exit 0
+fi
+
+echo "Failed to patch cocoapods. Try to re-apply the patch"
+echo "$patch_content" | patch --reverse --force --directory "$cocoapods_gem_path" --strip 0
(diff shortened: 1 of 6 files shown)


bin/publish_pod
```diff
@@ -10,6 +10,9 @@ if [ -n "$BUILDKITE_TAG" ] && [ "$BUILDKITE_TAG" != "$POD_VERSION" ]; then
 	exit 1
 fi
 
+echo "⚠️ Remove the patch-cocoapods step once this issue is fixed: https://github.com/CocoaPods/CocoaPods/issues/12033"
+patch-cocoapods
```
AliSoftware: I wonder if we shouldn't instead let the caller projects call `patch-cocoapods` before they call `validate_pod` and `publish_pod` in their own `.buildkite/commands/*.sh` scripts.

The reason I suggest separating those is because the patch depends on the version of CocoaPods being installed. If/when CocoaPods releases a new version with this issue fixed, some projects will adopt the new version quickly while others might take more time.

At least by letting the caller script decide, it will be easier once the new CP version arrives to just update CP in the project's `Gemfile.lock` then stop calling `patch-cocoapods` on the buildkite scripts in the same PR updating CP, rather than having to release a new version of a8c-ci-toolkit that will remove that patch's call… but then only work with projects that updated CocoaPods.

If we strongly couple the version of a8c-ci-toolkit with the version of CocoaPods used, if some projects want to update CocoaPods but are not ready to update a8c-ci-toolkit (eg say there was a major version of the toolkit since for unrelated changes, and the project isn't ready to adopt those yet) they'd be stuck.

AliSoftware (changes requested):

bin/publish_pod
crazytonyli: The main reason for putting this script here is: `install_gems` overwrites the local gems, so `patch-cocoapods` needs to run straight after `install_gems`. Also, it'd be nice to avoid copying scripts among libraries repos.

But I agree with you it's not ideally that this script may break pipelines.

Initially I tried to add an optional plugin configuration to the plugin, `patch-cocoapods: true`, but I can't get the environment variable `BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_PATCH_COCOAPODS` to work: it doesn't exist even though [the new option is set in pipeline](https://github.com/wordpress-mobile/WordPressKit-iOS/commit/0a960998625c3c2c4b6488c24aab0564d0ea3a73).

What do you think adding a new option to the existing shell script, like `validate_podspec --patch_cocoapods`?

AliSoftware: > The main reason for putting this script here is: install_gems overwrites the local gems, so patch-cocoapods needs to run straight after install_gems. Also, it'd be nice to avoid copying scripts among libraries repos.

Ah. Damn.

> Initially I tried to add an optional plugin configuration to the plugin, `patch-cocoapods: true`, but I can't get the environment variable `BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_PATCH_COCOAPODS` to work: it doesn't exist even though [the new option is set in pipeline](https://github.com/wordpress-mobile/WordPressKit-iOS/commit/0a960998625c3c2c4b6488c24aab0564d0ea3a73).

Just checking: did you declare the additional plugin config parameter in `plugin.yml` [here](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/plugin.yml#L12-L15) (then pointed to the commit that added it in your WPKiOS pipeline) when you attempted that?

iinm, the plugin has `additionalProperties: false` in its `plugin.yml` config, so all additional properties have to be declared explicitly in `plugin.yml` for them to be exposed as env var and then accessible within the plugin.

That being said…

> What do you think adding a new option to the existing shell script, like validate_podspec --patch_cocoapods?

I actually think that's probably a better approach 👍  so that the option is tied to the commands that need it instead of to the whole plugin (especially wouldn't make sense to expose that option to app / non-pod repos, which use the a8c-ci-toolkit plugin… but never call `validate_pod` nor `publish_pod` in their context)

crazytonyli: > did you declare the additional plugin config parameter in plugin.yml

Yes, I did. I tried naming it as `patch-cocoapods`, `patch_cocoapods`. I tried declaring it as string, boolean. None worked...

----

I have implemented the new `--patch_cocoapods` argument in the new commits. And [this test job](https://buildkite.com/automattic/wordpresskit-ios/builds/1152#018c18f7-6f8d-46f8-8012-94234f930b27) passed.

bin/patch-cocoapods
```diff
@@ -0,0 +1,47 @@
...
+echo "Path to the cocoapods gem: $cocoapods_gem_path"
+
+if echo "$patch_content" | patch --forward --force --directory "$cocoapods_gem_path" --strip 0; then
+    echo "cocoapods patched successfully"
+    exit 0
+fi
+
+echo "Failed to patch cocoapods. Try to re-apply the patch"
+echo "$patch_content" | patch --reverse --force --directory "$cocoapods_gem_path" --strip 0
+echo "$patch_content" | patch --forward --force --directory "$cocoapods_gem_path" --strip 0;
```
AliSoftware: Funny how one line has the `;` but not the other 😄

AliSoftware (approved):

bin/patch-cocoapods
crazytonyli: 🙈 It's an copy-paste error. Removed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/78/commits/6e154745852805ebe39703de4dbd69b0d4f0dc8f

crazytonyli:
I will publish a 3.0.1 release after this PR is merged.

PR #82, opened 2024-04-02
Support additional flags for cocoapods helpers

AliSoftware:
This allows our `validate_podspec` and `publish_pod` helpers to support additional flags to forward to the `pod` commands they wrap:

### `validate_podspec`

 - Add support for `--allow-warnings`, `--sources=…`, `--private`, `--include-podspecs=…`, `--external-podspecs=…`
 - The command now automatically adds `--include-podspecs="*.podspec"` to the call if not provided explicitly, in order to support repos that contains multiple co-dependant `.podspec` where one needs the other(s) to be able to validate(†).

### `publish_pod`

 - Add support for `--allow-warnings` and `--synchronous`

## Why?

The main goal is to support repos containing multiple co-dependant `podspecs`.
Examples include [the Gravatar SDK having both `Gravatar` & `GravatarUI`](https://a8c.slack.com/archives/CC7L49W13/p1712044283064279), or [`WordPress-Aztec-iOS` and `WordPress-Editor-iOS`](https://a8c.slack.com/archives/CC7L49W13/p1711030206956679?thread_ts=1711026617.605489&cid=CC7L49W13).

In those cases:
 - The validation of one pod might require to include the other pod too in order for the linting to not fail compilation (hence the need for `--include-podspecs=…`)
 - The `pod trunk push` requires `--synchronous` otherwise the push of the first pod will succeed, but when trying to push the second pod which has the first one as a dependency, it won't be able to find the newly-pushed first pod it depends on, because it would not have propagated to the CocoaPods' CDNs yet

## Testing

I've [tested those changes by doing a dummy `1.0.1-alpha` tag / version on the `Gravatar-SDK-iOS` repo after making it point to this branch of the CI_TOOLKIT plugin](https://github.com/Automattic/Gravatar-SDK-iOS/pull/147#issuecomment-2032367149).

---

### DIFF
--- a/bin/publish_pod
+++ b/bin/publish_pod
@@ -56,7 +56,7 @@ fi
 # https://github.com/Automattic/buildkite-ci/issues/7
 xcrun simctl list >> /dev/null
 
-bundle exec pod trunk push ${COCOAPODS_FLAGS[*]} "$PODSPEC_PATH"
+bundle exec pod trunk push "${COCOAPODS_FLAGS[@]}" "$PODSPEC_PATH"
 
 if [[ ! "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
   cache_cocoapods_specs_repo
--- a/bin/publish_pod
+++ b/bin/publish_pod
@@ -44,7 +44,7 @@ if [[ "${PATCH_COCOAPODS}" ==  'true' ]]; then
 	patch-cocoapods
 fi
 
-if [[ ! "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
+if [[ "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
 	# When using --synchronous to support pushing co-dependant podspecs one after the other,
 	# Cocoapods will use the master spec repo to avoid being dependant on CDN propagation time
 	# when validating a podspec against another, recently-deployed pod.
@@ -58,6 +58,6 @@ xcrun simctl list >> /dev/null
 
 bundle exec pod trunk push "${COCOAPODS_FLAGS[@]}" "$PODSPEC_PATH"
 
-if [[ ! "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
+if [[ "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
   cache_cocoapods_specs_repo
 fi
--- a/bin/publish_pod
+++ b/bin/publish_pod
@@ -1,10 +1,29 @@
 #!/bin/bash -eu
 
+# Usage: publish_pod [OPTIONS] PODSPEC_PATH
+#
+# OPTIONS:
+#  `--patch-cocoapods`:
+#     Apply a patch to work around issues with older deployment targets — see https://github.com/CocoaPods/CocoaPods/issues/12033
+#  `--allow-warnings`, `--synchronous`:
+#     Those options are passed to `pod trunk push` verbatim.
+#
+# Note: Use `--synchronous` if you have co-dependant podspecs in your repo and need to publish multiple pods at the same time.
+# Without this option, since the first pod you push will take time to propagate thru the CocoaPods CDNs, attempting to push
+# the other dependant pod(s) in your repo might fail to find the first pushed pod until it has propagated thru CDNs.
+#
+
 PATCH_COCOAPODS="false"
+COCOAPODS_FLAGS=(--verbose)
 
 while [[ "$#" -gt 0 ]]; do
 	case $1 in
-		--patch-cocoapods) PATCH_COCOAPODS="true" ;;
+		--patch-cocoapods)
+			PATCH_COCOAPODS="true"
+			;;
+		--allow-warnings | --synchronous)
+			COCOAPODS_FLAGS+=("$1")
+			;;
 		*) break ;;
 	esac
 	shift
@@ -25,8 +44,20 @@ if [[ "${PATCH_COCOAPODS}" ==  'true' ]]; then
 	patch-cocoapods
 fi
 
+if [[ ! "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
+	# When using --synchronous to support pushing co-dependant podspecs one after the other,
+	# Cocoapods will use the master spec repo to avoid being dependant on CDN propagation time
+	# when validating a podspec against another, recently-deployed pod.
+	# The master spec repo is pretty large to clone though, so using cache for it is important
+	restore_cache "$BUILDKITE_PIPELINE_SLUG-specs-repos"
+fi
+
 # For some reason this fixes a failure in `lib lint`
 # https://github.com/Automattic/buildkite-ci/issues/7
 xcrun simctl list >> /dev/null
 
-bundle exec pod trunk push --verbose "$PODSPEC_PATH"
+bundle exec pod trunk push ${COCOAPODS_FLAGS[*]} "$PODSPEC_PATH"
+
+if [[ ! "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
+  cache_cocoapods_specs_repo
+fi
--- a/bin/validate_podspec
+++ b/bin/validate_podspec
@@ -1,14 +1,45 @@
 #!/bin/bash -eu
 
+# Usage: validate_podspec [OPTIONS] [PODSPEC(S)_PATH(S)...] 
+#
+#  - If no `PODSPEC_PATH` provided, will lint all `*.podspec` files found
+#  - By default, the linting of each podspec will `--include-podspecs=*.podspec` any other podspec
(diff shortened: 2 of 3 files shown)


bin/publish_pod
```diff
@@ -25,8 +44,20 @@ if [[ "${PATCH_COCOAPODS}" ==  'true' ]]; then
 	patch-cocoapods
 fi
 
+if [[ "${COCOAPODS_FLAGS[*]}" =~ --synchronous ]]; then
+	# When using --synchronous to support pushing co-dependant podspecs one after the other,
+	# Cocoapods will use the master spec repo to avoid being dependant on CDN propagation time
+	# when validating a podspec against another, recently-deployed pod.
+	# The master spec repo is pretty large to clone though, so using cache for it is important
+	restore_cache "$BUILDKITE_PIPELINE_SLUG-specs-repos"
```
iangmaia, line 52: 👍

bin/publish_pod
```diff
@@ -1,10 +1,29 @@
...
 
+# Usage: publish_pod [OPTIONS] PODSPEC_PATH
+#
+# OPTIONS:
+#  `--patch-cocoapods`:
+#     Apply a patch to work around issues with older deployment targets — see https://github.com/CocoaPods/CocoaPods/issues/12033
+#  `--allow-warnings`, `--synchronous`:
+#     Those options are passed to `pod trunk push` verbatim.
+#
+# Note: Use `--synchronous` if you have co-dependant podspecs in your repo and need to publish multiple pods at the same time.
```
iangmaia, line 11: I wonder if it wouldn't make sense to automatically add `--synchronous` when multiple pod specs are used 🤔 

I can imagine a hypothetical scenario of pod specs that aren't dependent being pushed together, but I guess `--synchronous` wouldn't hurt either on that case?

bin/validate_podspec
```diff
@@ -29,4 +60,4 @@ echo "--- :microscope: Validate Podspec"
 # https://github.com/Automattic/buildkite-ci/issues/7
 xcrun simctl list >> /dev/null
 
-bundle exec pod lib lint --verbose --fail-fast
+bundle exec pod lib lint "${COCOAPODS_FLAGS[@]}" -- "$@"
```
iangmaia, line 63: What is the use of the extra `--` before the last parameter?

bin/publish_pod
AliSoftware: It would hurt the build time, because `--synchronous` implies that CocoaPods will clone the master Spec repo (which is **_huge_** —like multiple GBs—and quite long to clone) so that when it validates the second pod it can see the first and newly published pod immediately in the master spec repo instead of checking the CDN.

While without `--synchronous` it only checks the CDN (aka `trunk`) and doesn't need to clone (or restore cache of) the huge master spec repo nor update it via `git pull`; hence using the CDN is much faster.

---

One could argue that we don't have this level of refinement in our other helpers—for example our `bin/install_cocoapods` command unconditionally calls `restore_cache "$BUILDKITE_PIPELINE_SLUG-specs-repos"` and thus downloads the cache'd version of the CocoaPods master spec repo… even if in 95% of cases we don't need it at all (The `Podfile` of most repos should have removed the legacy `source 'https://github.com/CocoaPods/Specs.git'` line and migrated to use the CDN by now—but even if it has been years and years since CDN was introduced… very old app projects which were created even before that might never have taken the time to remove that legacy line though).

So in the same vein, maybe we should also modernize our `bin/install_cocoapods` to **only** call `restore_cache "$BUILDKITE_PIPELINE_SLUG-specs-repos"` if the `SPEC REPOS` section of the `Podfile.lock` contains the master spec repo URL (`https://github.com/CocoaPods/Specs.git`) explicitly (and then maybe emit a `buildkite-agent annotate --style warning` to suggest to migrate to CDN in those cases? 🤔 ), and skip the `restore_cache` if not? But that's an optimization for another PR, touching a different `bin/` helper than the ones at hand for solving the co-dependent pods validation issue.

bin/validate_podspec
AliSoftware: Security, by separating flags from arguments and avoid allowing any user-provided parameter (`"$@"`) to be interpreted as a flag… and allowing injection.

The idea was to only allow a specific and explicit subset of flags of the `pod lib lint` command (for example `--allow-warnings` or `--include-podspecs`) but not blindly permit any other random flag (like `--skip-tests` or `--allow-root`).

Otherwise I could have made the code much simpler by just using `bundle exec pod lib lint "$@"` on this line and not parse explicit flags in `COCOAPODS_FLAGS` individually in the switch case at the start… and as a result, when calling `validate_podspec --allow-warnings MyPod.podspec` the `"$@"` would have been expanded by the shell to `"--allow-warnings" "MyPod.podspec`—and that would have worked the same as the current code in this PR… BUT that would **also** have allowed anyone to call `validate_podspec --allow-root MyPod.podspec` and thus calling `pod lib lint --allow-root MyPod.podpsec`, while I didn't want that option to be allowed.

With the `--`, if one tries to call `validate_podspec --allow-warnings --allow-root MyPod.podspec`, this will result in `bundle exec pod lib lint "--allow-warnings" -- "--allow-root" "MyPod.podspec"` once this line is expanded—with `COCOAPODS_FLAGS` containing `-allow-warnings` and `$@` containing `--allow-root MyPod.podspec`. But thanks to the `--` I inserted in between, since `--allow-root` appears after that `--`, it will be interpreted as an argument—instead of risking of being interpreted as a flag because it starts with `--…`— and `pod lib lint` will interpret it as "I want to lint the pods named `--allow-root` and `MyPod.podspec`"—and will then fail because there's no file named `--allow-root`, successfully preventing using a flag we don't want to allow—instead of interpret it as "I want to lint the pod named `MyPod.podspec` but with the `--allow-root` option enabled"—and thus succeed… but also be open to potential root injections.

AliSoftware: See https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html

> The argument -- terminates all options; any following arguments are treated as non-option arguments, even if they begin with a hyphen.

iangmaia: > The idea was to only allow a specific and explicit subset of flags

I got that from the code, though I didn't realize the use of `--`. And actually... what would be the use case of other parameters at the end on `"$@"`?

> See https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html
> 
> > The argument -- terminates all options; any following arguments are treated as non-option arguments, even if they begin with a hyphen.

TIL! Makes sense, thanks. That makes me wonder about other cases we have in `a8c-ci-toolkit-buildkite-plugin` with that potential breach that may be good to revisit.

bin/publish_pod
iangmaia: > It would hurt the build time

Do we have a case of non dependent pod specs being pushed together, though?
One could also argue that, if there's such need, you should call `publish_pod` separately anyway, while calling it for multiple pod specs would always imply on having them pushed with `--synchronous` 🤔 

> So in the same vein, maybe we should also modernize our bin/install_cocoapods

That sounds like a good thing to do anyway 👍

bin/validate_podspec
AliSoftware: > And actually... what would be the use case of other parameters at the end on "$@"?

The name of the podspec to validate, if one wants to validate each podspec individually instead of keeping the default of `pod lib lint` linting every `*.podspec` it finds. This could potentially be useful if one wants to call `validate_podspec --allow-warnings A.podspec` but `validate_podspec B.podspec` without allowing warnings for `B`, for example.

bin/publish_pod
AliSoftware: Except that one might want to push one pod with `--allow-warnings` and the other without `--allow-warnings`, so in such case they wouldn't be able to call `publish_pod A.podspec B.podspec` as a single call.

That being said, I guess we could still consider making `publish_pod` automatically add `--synchronous` if `$# > 1` after consuming known flags… to at least make this common case automatic, all while still allowing to also provide it explicitly for cases where we'd want to call `publish_pod` on each pod separately (eg with different option for each).

---

The only drawback with that approach is that there wouldn't be an easy way to disable `--synchronous` if we wanted to publish multiple non-dépendant pods via a single call. In contrast, for the case of `--include-podspecs` (on `validate_podspec`) which is added implicitly if not provided, at least we can pass `--include-podspecs=''` to disable it, while there's no equivalent possibility if we were to make `--synchronous` implicit 🤔

iangmaia: > to at least make this common case automatic

Yeah, that was my main point -- I see it more as the rule rather than the exception.

> Except that one might want to push one pod with --allow-warnings and the other without --allow-warnings, so in such case they wouldn't be able to call publish_pod A.podspec B.podspec as a single call.

True. It's hard to anticipate all these uses while being somewhat opinionated on how the commands should be used.
Perhaps it's best then to keep the options that can easily cover most cases and not block potential uses 🤷‍♂️

iangmaia (approved):

PR #86, opened 2024-04-08
Add allow warnings flag to publish_private_pod

crazytonyli:
Similar to https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/82, this PR adds the `--allow-warnings` flag to another cocoapod command `publish_private_pod`, which we may need to use in some libraries.

---

### DIFF
--- a/bin/publish_private_pod
+++ b/bin/publish_private_pod
@@ -1,10 +1,15 @@
 #!/bin/bash -eu
 
 PATCH_COCOAPODS="false"
+COCOAPODS_FLAGS=(--verbose)
 
 while [[ "$#" -gt 0 ]]; do
 	case $1 in
-		--patch-cocoapods) PATCH_COCOAPODS="true" ;;
+		--patch-cocoapods)
+			PATCH_COCOAPODS="true" ;;
+		--allow-warnings)
+			COCOAPODS_FLAGS+=("$1")
+			;;
 		*) break ;;
 	esac
 	shift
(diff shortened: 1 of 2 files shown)


bin/publish_private_pod
```diff
@@ -1,10 +1,15 @@
 #!/bin/bash -eu
 
 PATCH_COCOAPODS="false"
+COCOAPODS_FLAGS=(--verbose)
 
 while [[ "$#" -gt 0 ]]; do
 	case $1 in
-		--patch-cocoapods) PATCH_COCOAPODS="true" ;;
+		--patch-cocoapods)
+			PATCH_COCOAPODS="true" ;;
```
AliSoftware: For stylistic consistency with rest of the code:
```suggestion
			PATCH_COCOAPODS="true"
			;;
```

AliSoftware (approved):

bin/publish_private_pod
crazytonyli: Good catch! Updated in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/86/commits/a8e3224672945c82b326033156c577077ba5816d.

AliSoftware:
Btw when I added similar changes to `publish_pod` I also added [some doc comments at the top of the scripts](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/publish_pod#L3-L14)… maybe worth doing the same here?

bin/publish_private_pod
AliSoftware: Mmh seems that commit mixed tabs on line 9 vs spaces in line 10, which made the statements indentation misaligned (depending on your tab width settings)

crazytonyli:
@AliSoftware Sure thing. Here is the follow up PR: https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/87

PR #90, opened 2024-04-25
Limit the SwiftLint arguments to "--strict" and "--lenient"

iangmaia:
For security, this PR restricts the possible arguments to the `run_swiftlint` command to `--strict` and `--lenient`, which are the ones currently in use by the apps. More arguments can be added if the need comes up in the future.

---

### DIFF
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -4,10 +4,15 @@ echo "--- :swift: Running SwiftLint"
 
 SWIFTLINT_ARGUMENTS=(--quiet --reporter relative-path)
 
-if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
-  SWIFTLINT_ARGUMENTS+=("$1")
-  shift
-fi
+while [[ "$#" -gt 0 ]]; do
+  case $1 in
+    --strict | --lenient)
+      SWIFTLINT_ARGUMENTS+=("$1")
+      shift
+      ;;
+    *) break ;;
+  esac
+done
 
 if [[ -n $1 ]]; then
   echo "Error: invalid arguments."
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -1,30 +1,25 @@
 #!/bin/bash -eu
 
-validation_error() {
+echo "--- :swift: Running SwiftLint"
+
+SWIFTLINT_ARGUMENTS=(--quiet --reporter relative-path)
+
+if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
+  SWIFTLINT_ARGUMENTS+=("$1")
+  shift
+fi
+
+if [[ -n $1 ]]; then
   echo "Error: invalid arguments."
   echo "Usage: $0 [--strict | --lenient]"
   exit 1
-}
-
-echo "--- :swift: Running SwiftLint"
-
-SWIFTLINT_ARGUMENT=""
-
-if [ "$#" -eq 1 ]; then
-  if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
-    SWIFTLINT_ARGUMENT=$1
-  else
-    validation_error
-  fi
-elif [ "$#" -gt 1 ]; then
-  validation_error
 fi
 
-SWIFTLINT_VERSION=$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')
-SWIFTLINT_CMD="docker run --rm -v $PWD:/workspace -w /workspace ghcr.io/realm/swiftlint:$SWIFTLINT_VERSION swiftlint"
+SWIFTLINT_VERSION="$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')"
+SWIFTLINT_DOCKER_CMD=(docker run --rm -v "$PWD":/workspace -w /workspace ghcr.io/realm/swiftlint:"$SWIFTLINT_VERSION" swiftlint)
 
 set +e
-SWIFTLINT_OUTPUT=$($SWIFTLINT_CMD lint --quiet "$SWIFTLINT_ARGUMENT" --reporter relative-path)
+SWIFTLINT_OUTPUT=$("${SWIFTLINT_DOCKER_CMD[@]}" lint "${SWIFTLINT_ARGUMENTS[@]}")
 SWIFTLINT_EXIT_STATUS=$?
 set -e
 
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -1,12 +1,30 @@
 #!/bin/bash -eu
 
+validation_error() {
+  echo "Error: invalid arguments."
+  echo "Usage: $0 [--strict | --lenient]"
+  exit 1
+}
+
 echo "--- :swift: Running SwiftLint"
+
+SWIFTLINT_ARGUMENT=""
+
+if [ "$#" -eq 1 ]; then
+  if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
+    SWIFTLINT_ARGUMENT=$1
+  else
+    validation_error
+  fi
+elif [ "$#" -gt 1 ]; then
+  validation_error
+fi
 
 SWIFTLINT_VERSION=$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')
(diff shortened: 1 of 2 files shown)


bin/run_swiftlint
```diff
@@ -1,12 +1,30 @@
...
+elif [ "$#" -gt 1 ]; then
+  validation_error
+fi
+
 SWIFTLINT_VERSION=$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')
 SWIFTLINT_CMD="docker run --rm -v $PWD:/workspace -w /workspace ghcr.io/realm/swiftlint:$SWIFTLINT_VERSION swiftlint"
 
 set +e
-SWIFTLINT_OUTPUT=$($SWIFTLINT_CMD lint --quiet "$@" --reporter relative-path)
+SWIFTLINT_OUTPUT=$($SWIFTLINT_CMD lint --quiet "$SWIFTLINT_ARGUMENT" --reporter relative-path)
```
AliSoftware: 💭❔ Worth checking if the case of passing `""` (in the case when `$SWIFTLINT_ARGUMENT` is empty) won't cause issues by it being seen as passed as an explicit positional argument 🤔 

I mean, I think that's ok, but never hurts to double-check 😅

---

<details><summary>Note that the <code>"$@"</code> syntax doesn't have this limitation because…</summary>

…the expansion of `"$@"`—or any `"${SOMEARRAY[@]}"` similar syntax—is handled specially by `bash`, where it internally iterates over each item of the `$@` bash array and wrap _**each individual element**_ around quotes
This is unlike `"$*"` for example, which expands the array as a string joined by `$IFS` and only then wraps that whole expanded string in a single pair of quotes.

As a result, the `"$@"` syntax creates as many arguments as there are items in the `$@` array… thus if the `$@` array is empty it doesn't even produce any argument at all. While `"$*"` generates always one, even if it ends up being the empty string.
</details>

mokagio: I tried it on my end with `swiftlint` installed via Homebrew. Note it was version 0.53.0, not 0.54.0 because [that version had issues shipping](https://github.com/Homebrew/homebrew-core/pull/153931).

image

The command above succeeded. I tried it without `--quiet`, too, just to see some output.

image

mokagio (approved):

bin/run_swiftlint
AliSoftware: Alternative approach would be to use the same approach we did in `publish_pod` and friends, where we manipulate an array variable to which we append arguments and then use `"${SWIFTLINT_ARGS[@]}"` in the final command being built.

AliSoftware: Confirmed the difference in bash interpretation really quickly this morning:

<details><summary>Test Scripts</summary>

```bash
$ cat log-args.sh 
#!/bin/bash

echo "== Called with: $* =="
for arg in "$@"; do
  echo "- Argument: \`${arg}\`"
done
echo "== Done =="
echo
```
```bash
$ cat test-args.sh 
#!/bin/bash

SARGS=""
if [[ $1 == "--strict" ]]; then
  SARGS="--strict"
fi
./log-args.sh string start "$SARGS" end

AARGS=()
if [[ $1 == "--strict" ]]; then
  AARGS+=("--strict")
fi
./log-args.sh array-star start "${AARGS[*]}" end
./log-args.sh array-at start "${AARGS[@]}" end
```
</details>

### Calling test script with an arg (`--strict`)

```
$ ./test-args.sh --strict
== Called with: string start --strict end ==
- Argument: `string`
- Argument: `start`
- Argument: `--strict`
- Argument: `end`
== Done ==

== Called with: array-star start --strict end ==
- Argument: `array-star`
- Argument: `start`
- Argument: `--strict`
- Argument: `end`
== Done ==

== Called with: array-at start --strict end ==
- Argument: `array-at`
- Argument: `start`
- Argument: `--strict`
- Argument: `end`
== Done ==
```

### Calling test script without arg

```       
✓ $ ./test-args.sh 
== Called with: string start  end ==
- Argument: `string`
- Argument: `start`
- Argument: ``
- Argument: `end`
== Done ==

== Called with: array-star start  end ==
- Argument: `array-star`
- Argument: `start`
- Argument: ``
- Argument: `end`
== Done ==

== Called with: array-at start end ==
- Argument: `array-at`
- Argument: `start`
- Argument: `end`
== Done ==
```

---

💄 ⛏️ So even if passing an empty argument doesn't break the `swiftlint` (as Gio tested), maybe this approach of using an array is more idiomatic and sematically correct (and would also align with how our other helpers in this repo do it)? 🤷 
```bash
SWIFTLINT_ARGUMENTS=(--quiet --reporter relative-path)
if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
  SWIFTLINT_ARGUMENTS+=("$1")
  shift
fi
if [[ -n $1 ]]; then
  echo "Error: invalid arguments."
  echo "Usage: $0 [--strict | --lenient]"
  exit 1
fi

SWIFTLINT_VERSION="$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')"
SWIFTLINT_DOCKER_CMD=(docker run --rm -v "$PWD":/workspace -w /workspace ghcr.io/realm/swiftlint:"$SWIFTLINT_VERSION" swiftlint)

SWIFTLINT_OUTPUT=$("${SWIFTLINT_DOCKER_CMD[@]}" lint "${SWIFTLINT_ARGUMENTS[@]}")
```

iangmaia: > the expansion of "$@"—or any "${SOMEARRAY[@]}" similar syntax—is handled specially by bash, where it internally iterates over each item of the $@ bash array and wrap each individual element around quotes

Good point. I agree using an array is more idiomatic (and leaves less room for this kind of wondering 😅 ). Updated it on 9238c0788fb2f0d926bfadd6c400d97ca5badfd2, thanks!

bin/run_swiftlint
```diff
@@ -2,11 +2,24 @@
...
 echo "--- :swift: Running SwiftLint"
 
-SWIFTLINT_VERSION=$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')
-SWIFTLINT_CMD="docker run --rm -v $PWD:/workspace -w /workspace ghcr.io/realm/swiftlint:$SWIFTLINT_VERSION swiftlint"
+SWIFTLINT_ARGUMENTS=(--quiet --reporter relative-path)
+
+if [ "$1" == "--strict" ] || [ "$1" == "--lenient" ]; then
+  SWIFTLINT_ARGUMENTS+=("$1")
+  shift
+fi
```
AliSoftware: Only drawback of this approach with a single `if` is that once it consumed `$1` it doesn't continue searching for subsequent arguments.

[For example, in our other commands](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/validate_podspec#L20), instead of a single `if`, we do a `while` loop to parse all potential flags, not just the first one.

Granted, in the specific case of this `run_swiftlint` command, since we only expect one argument at most (either `--strict` or `--lenient` or neigher of them) and never a second one… this is not super impactful for the moment. Except that:
 - If someone decides to run `run_swiftlint --lenient --strict` (probably not literally, but because one of their `.buildkite/commands/myscript.sh` is not written as well as it should for example?), then this logic will see the `--lenient` and totally ignore the `--strict`
 - This is not super future-proof, as if one day we want to support more arguments we'll have to remember to transform that into a `while` loop as opposed to being tempted to do some copy/paste of the current parsing code and risk imposing a constraint on the additional arguments' order.

This is not a blocker; but would still be nice to make the way we parse arguments consistent across all our `bin/*` commands of the plugin—not only for consistency but for resilience too in the long term 🙃

iangmaia: +1 for consistency and extensibility. Updated to a `while` loop on 3689d57e4ff3cc5795fe5f25eb417295b8c6b329.

bin/run_swiftlint
```diff
@@ -2,11 +2,29 @@
...
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --strict | --lenient)
+            SWIFTLINT_ARGUMENTS+=("$1")
+            shift
+            ;;
+        *) break ;;
+    esac
+done
```
AliSoftware: Indentation consistency: rest of this file uses 2 spaces not 4 🙃

iangmaia: 🦅 👀 
Fixed on e3a97f68b51d472840e910fb9854fc67e4f38f64.

AliSoftware (approved):

PR #92, opened 2024-05-15
Fix for the `unbound variable` error seen on `run_swiftlint` version `3.4.0`

iangmaia:
This is a fix for the error
```bash
run_swiftlint: line 17: $1: unbound variable
```
This error was seen when running `run_swiftlint` from Ruby's `system()` call [on CI](https://buildkite.com/automattic/wordpress-ios/builds/22327#018f7871-c7a1-449a-b3cf-c9bf2051f2cc/88-89).

---

### DIFF
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -14,7 +14,7 @@ while [[ "$#" -gt 0 ]]; do
   esac
 done
 
-if [[ -n ${1+x} ]]; then
+if [[ $# -gt 0 ]]; then
   echo "Error: invalid arguments."
   echo "Usage: $0 [--strict | --lenient]"
   exit 1
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -14,7 +14,7 @@ while [[ "$#" -gt 0 ]]; do
   esac
 done
 
-if [[ -n $1 ]]; then
+if [[ -n ${1+x} ]]; then
   echo "Error: invalid arguments."
   echo "Usage: $0 [--strict | --lenient]"
   exit 1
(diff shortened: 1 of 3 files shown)


bin/run_swiftlint
```diff
@@ -14,7 +14,7 @@ while [[ "$#" -gt 0 ]]; do
   esac
 done
 
-if [[ -n $1 ]]; then
+if [[ -n ${1+x} ]]; then
```
AliSoftware: That `+x` feels really odd. At first I thought it was some special bash syntax but I'm not seeing it in https://devhints.io/bash and I think I confused it with `${1:-}` which is what I'd use to get an empty string instead of undefined variable error:

```suggestion
if [[ -n ${1:-} ]]; then
```
Or alternatively you can check for the number of arguments instead of checking if `$1` is empty/undefined:
```suggestion
if [[ $# -gt 0 ]]; then
```

AliSoftware (changes requested):

bin/run_swiftlint
iangmaia: Ah, I like checking for the number of arguments (matching also the `while` on top) better! Nice suggestion, thanks!

AliSoftware: Btw worth double checking if it's supposed to be `0` or `1`: I always forget if `$0` (aka the program name) is counted in `$#` or not, and didn't test my suggestion

iangmaia: I tested to double check a bit before commiting: `$0` doesn't count in `$#`.

AliSoftware (approved):

PR #95, opened 2024-05-29
Fix `hash_directory` helper to be portable

### LINKED ISSUE #93: Make hash_directory portable and working on other system than Mac
From: https://github.com/Automattic/studio/pull/164#issuecomment-2130009902

> We should fix that in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/hash_directory#L14 so that it calls our hash_file wrapper in place of calling sha256sum directly in that find … command.

**Done is:**
- [x] `hash_directory` works on other systems than Mac

AliSoftware:
Fixes https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/issues/93

## Testing Instructions

In a Terminal on a Mac, `cd` to the root of the working directory of this repo, then:
 - Test `hash_directory`:
    - Run `bin/hash_directory bin/`, and verify that it outputs the hash `24f38e4d7a8db6137b69807fa86e70ea6ba54b78e3441e6b52d55e22cedf5b3f`
    - Run `bin/hash_directory bin` (i.e. without the ending `/`) and verify that it outputs the exact same hash
    - Run `docker run --rm -it -w /workdir -v "$(pwd)":/workdir ubuntu bin/hash_directory bin/` and verify it works and still outputs the same hash
    - Run `docker run --rm -it -w /workdir -v "$(pwd)":/workdir ubuntu bin/hash_directory bin` (i.e. without the trailing `/` in dir name) and verify it works and still outputs the same hash
 - Test `hash_file`:
    - Run `bin/hash_file bin/annotate_test_failures` and verify it outputs the hash `1c312fe4a57e5f6085fef2c75f2778b528d61ea8129ce9960b8bc10bdc0f8501`
    - Run `docker run --rm -it -w /workdir -v "$(pwd)":/workdir ubuntu bin/hash_file bin/annotate_test_failures` and verify it outputs the same hash

### DIFF
--- a/bin/hash_directory
+++ b/bin/hash_directory
@@ -18,4 +18,4 @@ fi
 # - Run `sha256sum` on each file found – the `+` flag does it in parallel for a huge speed boost.
 # - Sort the files by filename for deterministic hashing
 # - Take the hash of all of the output hashes (and file paths)
-find "$DIRECTORY_PATH" -type f -exec $sha_command "{}" \+ | sort -k 2 | $sha_command | cut -f1 -d " "
+find "${DIRECTORY_PATH%/}" -type f -exec $sha_command "{}" \+ | sort -k 2 | $sha_command | cut -f1 -d " "
--- a/bin/hash_directory
+++ b/bin/hash_directory
@@ -9,13 +9,13 @@ fi
 
 # `shasum` is available on only macOS
 if command -v shasum &> /dev/null; then
-  sha_command='shasum -a 256'
+  sha_command=(shasum -a 256)
 else
-  sha_command='sha256sum'
+  sha_command=(sha256sum)
 fi
 
 # - Find all files in the given directory
 # - Run `sha256sum` on each file found – the `+` flag does it in parallel for a huge speed boost.
 # - Sort the files by filename for deterministic hashing
 # - Take the hash of all of the output hashes (and file paths)
-find "${DIRECTORY_PATH%/}" -type f -exec $sha_command "{}" \+ | sort -k 2 | $sha_command | cut -f1 -d " "
+find "${DIRECTORY_PATH%/}" -type f -exec "${sha_command[@]}" "{}" \+ | sort -k 2 | "${sha_command[@]}" | cut -f1 -d " "
--- a/bin/hash_directory
+++ b/bin/hash_directory
@@ -7,8 +7,15 @@ if [ -z "$1" ]; then
 	exit 1
 fi
 
+# `shasum` is available on only macOS
+if command -v shasum &> /dev/null; then
+  sha_command='shasum -a 256'
+else
+  sha_command='sha256sum'
+fi
+
 # - Find all files in the given directory
 # - Run `sha256sum` on each file found – the `+` flag does it in parallel for a huge speed boost.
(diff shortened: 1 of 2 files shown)


bin/hash_directory
```diff
@@ -7,8 +7,15 @@ if [ -z "$1" ]; then
...
+else
+  sha_command=(sha256sum)
+fi
+
 # - Find all files in the given directory
 # - Run `sha256sum` on each file found – the `+` flag does it in parallel for a huge speed boost.
 # - Sort the files by filename for deterministic hashing
 # - Take the hash of all of the output hashes (and file paths)
-find "$DIRECTORY_PATH" -type f -exec sha256sum "{}" \+ | sort -k 2 | shasum -a 256 | cut -f1 -d " "
+find "${DIRECTORY_PATH%/}" -type f -exec "${sha_command[@]}" "{}" \+ | sort -k 2 | "${sha_command[@]}" | cut -f1 -d " "
```
AliSoftware, line 21: The `%/` is to eliminate the `/` suffix if present, because while on Linux it doesn't make a difference to `find` if we include a `/` suffix or not, on Mac running `find bin/` will result in the output looking like `bin//annotate_test_failures` with double slashes (the slash from the input dir + the slash added by `find` itself when building the full path of the files it found) 🤦 . Which thus generates a different final overall sha.

So in order to make sure the hash are the exact same on Mac and Linux I thus ensure to remove the `/` suffix if present.

AliSoftware: PS: I considered having the `find` command below calling our `hash_file` helper instead of replicating the `if…else…fi` in `hash_directory` here… but it didn't work because our `hash_file` helper requires a parameter to be passed (`$1`), and thus is not currently designed to accept taking input from STDIN.
And as we are using it that way in the second part of this command (`sort -k 2 | hash_file | cut -f1 -d " "`) it would end up returning `You` (i.e. the first word—because of `cut -f1`—of the "You must pass a filename to hash" error message) instead of the expected output 😅 

I figured that allowing `hash_file` to support `STDIN` as input just to fix this was not in scope of this PR and that it was simpler to just replicate the `if …` test here instead.

bin/hash_file
```diff
@@ -7,11 +7,11 @@ fi
...
+  sha_command=(shasum -a 256)
 else
-  sha_command='sha256sum'
+  sha_command=(sha256sum)
 fi
 
 # Both `shasum` and `sha256sum` will print the hash and the file name (`$1`).
 # We only care about the hash, so we use `cut` to extract it.
-$sha_command "$1" | cut -f1 -d " "
+"${sha_command[@]}" "$1" | cut -f1 -d " "
```
AliSoftware, line 17: This was updated to use a bash Array and then `"${array[@]}"` quoted array expansion syntax especially to fix the `shellcheck` violation `SC2086` we had in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/625#018fc58c-b850-4838-bc26-8cf8868c8289/163-174

crazytonyli (approved):

wojtekn:
Thanks for the fix @AliSoftware !

PR #96, opened 2024-05-30
Add alternative command to run `npm ci` with cache

### LINKED ISSUE #94: Improve `install_npm_packages` based on Studio build process
From: https://github.com/Automattic/studio/pull/166/files#r1613715724

The `install_npm_packages` command installs NPM packages. It was a bit limited for Studio, so Studio used a custom shell script, with the notable differences:

- use `patches/` directory to build a cache key
- use `npm ci` vs `npm install`

Let's review if it makes sense to bring those improvements upstream and use the main command in Studio.

**Done is:**
- [ ] We review if it's with porting Studio NPM install script to upstream

mokagio:
_WIP_

---


bin/install_npm_packages
```diff
@@ -2,13 +2,13 @@
 
 # Ensure package.json is present
 if [ ! -f package.json ]; then
-    echo "No valid package.json file found"
+    echo "No package.json found!"
```
mokagio, line 5: Me being pedantic:

- No validation made on `package.json`
- We know that `package.json` is a file

bin/install_npm_packages_clean
```diff
@@ -0,0 +1,71 @@
+#!/bin/bash -eu
+
+# Ensure package.json is present
+if [ ! -f package.json ]; then
+    echo "No package.json found!"
+    exit 1
+fi
+
+# Ensure package-lock.json is present
+# TODO: What to do with yarn-based projects? Make the hash on lockfile optional?
```
mokagio: I guess a project using `yarn` would not call a command with `npm` in the name?

bin/install_npm_packages_clean
```diff
@@ -0,0 +1,71 @@
...
+ARCHITECTURE=$(uname -m)
+NODE_VERSION=$(node --version)
+PACKAGE_HASH=$(hash_file package-lock.json)
+
+# TODO: This comes from Studio. Figure out if it's standard in Node or not.
+if [ -d patches ]; then
+  PATCHES_HASH=$(hash_directory patches/)
+else
+  PATCHES_HASH=nopatch
+fi
```
mokagio: Even if it isn't a standard approach, it might be worth leaving it here as it adds basically no overhead

bin/install_npm_packages_clean
```diff
@@ -0,0 +1,71 @@
...
+
+CACHEKEY="$BUILDKITE_PIPELINE_SLUG-npm-$PLATFORM-$ARCHITECTURE-node-$NODE_VERSION-$PACKAGE_HASH-$PATCHES_HASH"
+
+LOCAL_NPM_CACHE=./vendor/npm
+mkdir -p $LOCAL_NPM_CACHE
+echo "--- :npm: Set npm to use $LOCAL_NPM_CACHE for cache"
+npm set cache $LOCAL_NPM_CACHE
+echo "npm cache set to $(npm get cache)"
```
mokagio, line 33: Might be worth making this configurable?

bin/install_npm_packages_clean
```diff
@@ -0,0 +1,71 @@
...
+# To avoid constant ECONNRESET errors a limit is set for Linux,
+# as this is not happening with the Mac jobs.
+# This issue is being tracked here:
+# https://github.com/npm/cli/issues/4652
+if [ "$PLATFORM" = "Linux" ]; then
+  MAX_SOCKETS=1
+else
+  # Default value from npm.
+  # TODO: Link to source
+  MAX_SOCKETS=15
```
mokagio, line 49: The code this script was extracted from had the comment about 15 being the default. I had a _quick_ look and couldn't find a reference. Need to look more and document.

bin/install_npm_packages_clean
AliSoftware: FYI the implementation was inspired by this recommendation: https://www.npmjs.com/package/patch-package#circleci ; so I guess we could consider that "standard" in that regard?

mokagio: Oh, nice. 

I'll add a link to it and remove the todo. Thanks!

mokagio:
Mmmm.... 🤔 

`environment.bats` [fails in CI](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/636#018fdbb0-945b-4fbc-b00f-c565ce5c5521) but passes for me locally. Usually when a test behaves like that there are differences between CI and local environment, but in this case the tests run through Docker.

image

image

mokagio:
Note: I moved the Windows stuff to https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/100 because there's additional work to be done on that front and I wanted to keep the two streams separated. 

Sorry for the noise that the rebase created.

AliSoftware (commented):
Nitpick / Question: is the goal to keep this new script being called `install_npm_packages_clean`? Or is this just for ease of testing the new setup using a temp separate name, but once the PR is ready for review you'll end up removing `install_npm_packages` and renaming `install_npm_packages_clean` to `install_npm_packages`, or alternatively (in case we want to keep the old implementation around for some time, consolidate the two scripts in one and add a `--clean` flag or something to allow to switch between the two behaviors?

Asking mostly because I'm not sure I like the naming of `…_clean` in that new command… especially since one of its role is to use `npm ci` + `save_cache` so us still using the cache doesn't feel like the term "clean" (as in "clean install") is appropriate? Or is it because, while we do use some caching, we only cache the root npm folder with the package downloads… but not the `node_modules`, which in a way _are_ clean-installed? Though even if that's the case, it's not completely a clean install since if the cached downloads gets corrupted for some reason that cache will still poison the installed modules and in such case the naming of the helper might be misleadingly make us believe that it couldn't possibly be a case of cache poisoning because the command is named `…_clean`…

That being said, I'm not sure I have a suggestion for a better naming that would indicate "clean install of the modules, but still using cache for the downloaded packages"… yet, idk why but even if we keep the "clean" wording, it feels it'd be more logical to have it as a `--clean` flag than as a separate command? 🤷

mokagio:
At the moment, I wrote this as a dedicated script to avoid breaking the `npm install` one.

But like you noticed, it's highly likely the only difference between the two will be whether we call `install` or `ci` ([`clean-install`](https://docs.npmjs.com/cli/v10/commands/npm-ci)). As such, once this is proven to work (and to make a difference in terms of performance, which I'm a bit unsure about still) I was thinking about something along the lines of the `--clean` option (or the opposite, since `ci` is the recommended command for automated environment, we could opt-in using `install` via an option like `--cache-node-modules` to signify we cache `node_modules`, too)

> Asking mostly because I'm not sure I like the naming of …_clean in that new command… especially since one of its role is to use npm ci + save_cache so us still using the cache

I wonder if this applies more broadly to our existing naming convention. `install_gems`, `install_npm_packages`, `install_cocoapods`... these all come with a cache but it's not clear from their name.

`install_npm_packages_from_cache` maybe? (`install_npm_packages_loading_the_cache_if_any_first` would be even clearer but it's too big of a stretch 🙃 )

mokagio:
See also https://github.com/bloom/DayOne-Desktop/pull/66#discussion_r1980616778

mokagio:
If/when we'll go for this consolidate implementation, we might also want to allow passing additional flags down to the `npm install|ci` or `yarn` call. See for example how Simplenote Electron has to use the `--legacy-peer-deps` flag. https://github.com/Automattic/simplenote-electron/issues/3301

ParaskP7:
RTO: @mokagio I am seeing this PR being included on the `Apps Infra Roadmap` project, but without a status, causing it show at the bottom of the `Apps Infra Roadmap` board within the `No Status` section, is that intentional? 🤔

mokagio:
> RTO: @mokagio I am seeing this PR being included on the Apps Infra Roadmap project, but without a status, causing it show at the bottom of the Apps Infra Roadmap board within the No Status section, is that intentional? 🤔

It was not intentional. I must have forgot to set a status when adding this PR to the Apps Infra Roadmap project.

bin/install_npm_packages_clean
AliSoftware: Found [the source of that default in the `npm/make-fetch-happen` library](https://github.com/npm/make-fetch-happen/blob/main/README.md#opts-max-sockets), which is [used by `npm/cli`](https://github.com/npm/cli/blob/f6028e67aa1b2696e60e115d870182a026bae07f/package.json#L89)

```suggestion
  # Default value from npm. (See https://github.com/npm/make-fetch-happen/blob/main/README.md#opts-max-sockets)
  MAX_SOCKETS=15
```

AliSoftware:
@mokagio How do you feel about undrafting this PR and pushing it to the finish line? IMHO:
 - We can go replacing the `install_npm_dependencies` completely, instead of keeping 2 separate `install_npm_dependencies` vs `install_npm_dependencies_clean`, including because using `npm ci` is the official recommended approach by Nodejs when running on CI as it doesn't risk re-resolving dependencies (but instead only relies on the lockfile), so it's really the right command to use, not `npm install`.
 - Then you can add a "breaking" entry in the `CHANGELOG.md`, and bump the major version, just to be on the safe side to mark the change in behavior of that command. If we ever notice that one of our client repos that uses this command doesn't work well with the new implementation that uses `npm ci`, we can always decide to not update it to the latest major version of the plugin while we iterate to adjust as necessary.
 - We can start adopting it on repos that have already tested that `npm ci` works as expected ([DayOne-Desktop](https://github.com/bloom/DayOne-Desktop/pull/66#discussion_r1980616778), [Simplenote-Electron](https://github.com/Automattic/simplenote-electron/issues/3301), [experimental-wp-dev-env](https://github.com/Automattic/experimental-wp-dev-env/blob/trunk/.buildkite/commands/install_node_dependencies.sh), …)

> If/when we'll go for this consolidate implementation, we might also want to allow passing additional flags down to the `npm install|ci` or `yarn` call. See for example how Simplenote Electron has to use the `--legacy-peer-deps` flag. [Automattic/simplenote-electron#3301](https://github.com/Automattic/simplenote-electron/issues/3301)

I think the passing of unfiltered `"$@"` extra arguments on [this line](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/96/files#diff-93feaee8246a0898cbaff26b563c4ecd35634a01a221707f166fe1e99e2f5ca3R58) is good enough for a first iteration tbh. We can always iterate with a later PR if we want to filter allowed parameters.

PR #103, opened 2024-08-02
Introduce `merge_junit` script

wzieba:
To improve performance of Android unit tests collection. Originated it https://github.com/woocommerce/woocommerce-android/pull/12064

---


bin/merge_junit
AliSoftware, line 1: Nitpick: I'd recommend naming this script `merge_junit_reports` instead, to make it clearer that we're not trying to merge the Unit Tests _themselves_ somehow (?), but just the _report files_ 😛

bin/merge_junit
```diff
@@ -0,0 +1,43 @@
...
+
+touch "$output_file"
+{
+  echo '<?xml version="1.0" encoding="UTF-8"?>'
+  # Write XML header to the output file
+  echo '<testsuites>'
+
+  # Merge the content of all input JUnit files in the directory.
+  # (Note that in the case of Unit Tests, the JUnit XML files produced by Gradle
+  # don't have a parent `<testsuites>` root tag, so there's no need to try and remove it)
```
AliSoftware: While I was ok with this assumption being held for WooCommerce where you initially implemented this, now that this is going to become a broader utility for more repositories to use, I feel like we should be more careful in our implementation and assumptions here, especially to make sure we support cases and other projects where such assumption might not be always true 🤔

wzieba: What do you suggest? I think this is a standard behavior of JUnit working with Android Gradle Plugin - there's nothing specific in WooCommerce Android.

AliSoftware: My only fear is that someone would be tempted to use this `merge_junit_report` utility also on JUnit reports that don't come from Unit Tests created by Android Gradle Plugin, like JUnit reports created by fastlane, or by other wrappers (e.g. maybe `electron-builder` when building the Android app of an Electron codebase?), or for test reports that are also in the JUnit format but for tests other than Unit tests (e.g. UI Tests / E2E Tests?).
If it happens that those other reports, despite being in the JUnit format, don't fit the assumption that this `merge_junit_reports` utility makes of them not being wrapped within `<testsuites>` root node because they are generated by another tool than Android Gradle Plugin for Unit Tests, then this will break the merged file that it generates.

By contrast, just adjusting the `sed` pattern to remove those `<testsuites>` root elements if they happen to be present in the input file doesn't seem like something too complex to add (iirc I even suggested a pattern for it in the WCAndroid PR where you introduced that script initially) while making this `merge_junit_reports` utility more resilient and less likely to create invalid XML files.

wzieba: Gotcha 👍 I've added a logic to ignore `<testsuites>` tag in reports that are being merged. WDYT? 1913f43ae9ceecbef370414154264f0f95373539

I've tried to add a test for this with `bats`, but surprisingly it happened to be quite unstable even with a simple test - occasionally it was ending in a state, that was simply not finishing a test. After killing the process and running test again all worked fine. I didn't want to introduce flakiness though.

AliSoftware (approved):

CHANGELOG.md
```diff
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Introduce `merge_junit` script [#103]
```
```suggestion
- Introduce `merge_junit_reports` script [#103]
```

PR #104, opened 2024-08-05
Prepare release 3.5.0

wzieba:
---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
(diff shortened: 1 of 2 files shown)


wzieba:
I wanted to remove version in `README.md` as I don't think this manual task is really necessary, while we have the recent version on the sidebar. But I see we have a linter for this, so I'm guessing this is very intentional - leaving as it is then. Maybe we can think of it in a separate PR?

AliSoftware (approved):

PR #105, opened 2024-08-06
Refactor `install_swiftpm_dependencies` to allow specifying workspace/scheme/project

AliSoftware:
## Why?

To solve issues like the one reported in https://github.com/Automattic/pocket-casts-ios/issues/1990

TL;DR: the `install_swiftpm_dependencies` utility was calling `xcodebuild -resolvePackageDependencies` without the `-workspace` flag, thus relying on the `Package.resolved` from the `.xcodeproj` instead of the `.xcworkspace` by default (at least that's how that `xcodebuild` command seems to behave).
This led to us having multiple `Package.resolved` files (one in `xcodeproj`, one in `xcworkspace`, and two sources of truth that were disagreeing with each other, depending on which method we used to resolve the dependencies last (Xcode UI or `xcodebuild` or this a8c-ci-toolkit utility).

## What?

This PR makes some changes to our `install_swiftpm_dependencies` utility so that:
 - The caller can specify `--workspace PATH --scheme NAME` explicitly to resolve dependencies based on an Xcode `.xcworkspace` workspace
    - The `xcodebuild --help` documentation suggests that we should be able to call `xcodebuild -workspace -resolvePackageDependencies` without providing a `-scheme` parameter… but in practice doing so makes `xcodebuild` complain about that parameter missing and exit. Hence why this initial refactoring will expect a `--scheme` to be passed too, even if in practice I'm pretty sure the scheme doesn't have any effect on the packages resolution itself
    - There [might actually be a way to avoid this issue / `xcodebuild` bug and not need to pass `-scheme` when passing `-workspace` after all, by instead passing `-list` instead](#discussion_r1705963616). This might require some more testing to validate it works as expected though (and especially still honors the `-onlyUsePackageVersionsFromResolvedFile` option in those cases too)
 - Or alternatively specify `--project PATH` explicitly to resolve dependencies based on an Xcode lone `.xcodeproj` project
 - Or alternatively explicitly specify `--use-spm` to use `swift package resolve` instead of `xcodebuild` (typically for library repos which don't have an Xcode project and workspace and fully rely on a `Package.swift` file at the root)
 - Improve the logic at attempting to auto-detect the right thing to do when none of the above parameters are provided
    - If there's a single `xcworkspace` at the root of the repo, and no `Package.swift` at the root, use `xcodebuild -workspace`, using (arbitrarily) the first scheme found in that workspace for the `-scheme` parameter of `xcodebuild` (that is documented not to be needed but in practice errors out if not provided—there [might be room for improvement here / hope for not needing this after all though](#discussion_r1705963616)).
    - Otherwise if there's no `xcworkspace` but there's a `Package.swift` file, use `swift package resolve`
 - Ensure that we use `-workspace` when resolving swift dependencies using `xcodebuild` and have the cache key and the dependency resolution both based on the `.xcworkspace/xcshareddata/swiftpm/Package.resolved` file—as opposed to incorrectly using one from a `*.xcodeproj/**/Package.resolved`.

## WIP / Draft

> [!IMPORTANT]
> The changes in this PR and bash script have not been really tested. In practice I've written the logic mostly blind (i.e. without trying to run it on any repo to test it), and I'm only opening the PR as Draft so soon so I can get feedback on the code and the logic.

We should definitively take some extra time to test those changes on various repos in various setups:
 - lib repo with `Package.swift`, repo with `.xcworkspace` at repo root but `.xcodeproj` in subdir, repo with both `.xcworkspace` and `.xcodeproj` files at repo root, repo with multiple `.xcworkspace` and/or `.xcodeproj` files and multiple `Package.resolved` files, repos with no `Package.resolved` file at all…;
 - calling the utility with explicit `--workspace`/`--scheme`/`--project`/`--use-spm` parameters, with only `--workspace` and letting it guess the scheme, with no parameter, …

## Compatibility / Breaking Change

Note that this is technically a breaking change for the `install_swiftpm_dependencies` utility, since it previously accepted 2 optional extra arguments (full custom path to `Package.resolved`, and build type), while the new implementation doesn't accept those optional extra arguments anymore.

This is intentional, as:
 - It would be complex (and for little benefit) to keep backwards compatibility while also making it consistent with the new flags
 - But more importantly, I think those optional arguments were not working as expected. In particular, providing a custom path for the `Package.resolved` as the first argument made this path used for the computation of the cache key during cache restoration and save… but that custom file was not the one used by `swift package resolve` or `xcodebuild -resolvePackageDependencies` during dependency resolution, so that would lead to inconsistencies and odd behaviors if used

Tbh though, I'm not even sure any of our repos currently calling `install_swiftpm_dependencies` ever passed extra parameters to it though, so those were probably never ever used in practice anyway?

## Suggestion for adopting repos

Repos like [PocketCasts-iOS were having issues with that utility](https://github.com/Automattic/pocket-casts-ios/issues/1990) because it was inconsistently using either `podcasts.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` when calling `xcodebuild -resolvePackageDependencies`(without a `-workspace`, thus making it fallback to the `.xcodeproj`) during resolution while the cache was relying on `podcasts.xcworkspace/xcshareddata/swiftpm/Package.resolved` for the cache (or when developers resolved dependencies via Xcode and commited that file), and those two different `Package.resolved` files were not in sync.

I'd suggest for repos in such situations to:
 - Delete the `*.xcodeproj/xcshareddata/swiftpm/Package.resolved` file and remove it from git, so that only the one in the `*.xcworkspace` is kept around
 - Update the calls to `install_swiftpm_dependencies` in the `.buildkite/**` scripts to provide the `--workspace` (and `--scheme`) explicitly

---

### DIFF
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -48,10 +48,10 @@ if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "tru
   FOUND_ROOT_WORKSPACE_PATHS=(*.xcworkspace)
   if  [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 1 && -d "${FOUND_ROOT_WORKSPACE_PATHS[0]}" && ! -f "Package.swift" ]]; then
     XCWORKSPACE_PATH="${FOUND_ROOT_WORKSPACE_PATHS[0]}"
-    echo " --> Found a single \`.xcworkspace\` file and no \`Package.swift\` in the root of the repo."
+    echo " --> Found a single \`.xcworkspace\` file, and no \`Package.swift\` in the root of the repo."
     echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
   elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.swift" ]]; then
-    echo " --> Found a \`Package.swift\` and no root \`.xcworkspace\` in the root of the repo."
+    echo " --> Found a \`Package.swift\`, and no \`.xcworkspace\` in the root of the repo."
     echo "     Defaulting to \`--use-spm\`"
     USE_SPM=true
   else
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -103,7 +103,7 @@ if [[ -d "${XCWORKSPACE_PATH}" ]]; then
   echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
   if [[ -z "${SCHEME_NAME}" ]]; then
     SCHEME_NAME=$(xcodebuild -workspace "${XCWORKSPACE_PATH}" -onlyUsePackageVersionsFromResolvedFile -skipPackageUpdates -list | sed -n '/Schemes:/{n;s/^ *//p;}')
-    echo "No \`--scheme\` parameter provided; assuming the first one found (as it shouldn't matter in practice?): \`${SCHEME_NAME}\`"
+    echo "No \`--scheme\` parameter provided. Using the first one found in the list: \`${SCHEME_NAME}\`"
   fi
   echo "Using -workspace \"${XCWORKSPACE_PATH}\" -scheme \"${SCHEME_NAME}\""
   xcodebuild -workspace "${XCWORKSPACE_PATH}" -scheme "${SCHEME_NAME}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -3,7 +3,7 @@
 print_usage() {
   echo "Usage:"
   echo "  Using the \`Package.resolved\` of an Xcode \`.xworkspace\`:"
-  echo "    $0 --workspace PATH --scheme NAME"
+  echo "    $0 --workspace PATH"
   echo "  Using the \`Package.resolved\` of an Xcode \`.xcodeproj\`:"
   echo "    $0 --project PATH"
   echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` instead of Xcode:"
@@ -12,38 +12,33 @@ print_usage() {
 
 # Parse input parameters
 XCWORKSPACE_PATH=
-SCHEME_NAME=
 XCODEPROJ_PATH=
 USE_SPM=false
-while [[ "$#" -gt 0 ]]; do
-  case $1 in
-    --workspace)
-      XCWORKSPACE_PATH="$2"
-      shift
-      ;;
-    --scheme)
-      SCHEME_NAME=$2
-      shift
-      ;;
-    --project)
-      XCODEPROJ_PATH=$2
-      shift
-      ;;
-    --use-spm)
-      USE_SPM=true
-      ;;
-    -h|--help)
-      print_usage
-      exit 0
-      ;;
-    *)
-      echo "Unknown option or unexpected argument: $1"
-      print_usage
-      exit 2
-      ;;
-  esac
-  shift
-done
+
+case ${1:-} in
+  --workspace)
+    XCWORKSPACE_PATH="$2"
+    shift; shift
+    ;;
+  --project)
+    XCODEPROJ_PATH=$2
+    shift; shift
+    ;;
+  --use-spm)
+    USE_SPM=true
+    shift
+    ;;
+  -h|--help)
+    print_usage
+    exit 0
+    ;;
+esac
+
+if [[ "$#" -gt 0 ]]; then
+  echo "Unexpected extra arguments: $@" >&2
+  print_usage
+  exit 2
+fi
 
 SPM_CACHE_LOCATION="$HOME/Library/Caches/org.swift.swiftpm"
 
@@ -56,12 +51,12 @@ if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "tru
     echo " --> Found a single \`.xcworkspace\` file and no \`Package.swift\` in the root of the repo."
     echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
   elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.swift" ]]; then
-    echo " --> Found a \`Package.swift\` at the root of the repo (and no root \`.xcworkspace\`)"
-    echo "     so we are assuming \`--use-spm\`"
+    echo " --> Found a \`Package.swift\` and no root \`.xcworkspace\` in the root of the repo."
+    echo "     Defaulting to \`--use-spm\`"
     USE_SPM=true
   else
-    echo " -!- No valid --workspace & --scheme, or --project, or --use-spm flag provided, and cannot guess which one to use either, so aborting."
-    echo "     Please call $0 with an explicit \`--workspace PATH\` and \`--scheme NAME\`, or \`--project PATH\`, or \`--use-spm\` flag to disambiguate."
+    echo " -!- No valid --workspace, --project or --use-spm flag provided, and cannot guess which one to use either, so aborting."
+    echo "     Please call $0 with an explicit \`--workspace PATH\`, \`--project PATH\` or \`--use-spm\` flag to disambiguate."
     exit 1
   fi
 fi
@@ -101,12 +96,12 @@ add_host_to_ssh_known_hosts github.com
 # Resolve the packages using the correct method
 if [[ -d "${XCWORKSPACE_PATH}" ]]; then
   echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
-  if [[ -z "${SCHEME_NAME}" ]]; then
-    SCHEME_NAME=$(xcodebuild -workspace "${XCWORKSPACE_PATH}" -onlyUsePackageVersionsFromResolvedFile -skipPackageUpdates -list | sed -n '/Schemes:/{n;s/^ *//p;}')
-    echo "No \`--scheme\` parameter provided. Using the first one found in the list: \`${SCHEME_NAME}\`"
-  fi
-  echo "Using -workspace \"${XCWORKSPACE_PATH}\" -scheme \"${SCHEME_NAME}\""
-  xcodebuild -workspace "${XCWORKSPACE_PATH}" -scheme "${SCHEME_NAME}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
+  echo "Using -workspace \"${XCWORKSPACE_PATH}\""
+  # Note: we use `-list` as a workaround because otherwise `xcodebuild` complains that using `-workspace` requires to also provide the `-scheme` option
+  # (despite the help page of `xcodebuild` suggesting that it should work without `-scheme`). Since the dependency resolution doesn't really depend on the scheme
+  # and we don't want to have to provide or guess it, using `-list` instead stops making `xcodebuild` complain about `-workspace` not being used in conjunction
+  # with `-scheme` (even if in practice we don't care about the scheme list it returns)
+  xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list
 elif [[ -d "${XCODEPROJ_PATH}" ]]; then
   echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
   echo "Using -project \"${XCODEPROJ_PATH}\""
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -35,12 +35,12 @@ case ${1:-} in
 esac
 
 if [[ "$#" -gt 0 ]]; then
-  echo "Unexpected extra arguments: $@" >&2
+  echo "Unexpected extra arguments: $*" >&2
   print_usage
   exit 2
 fi
 
-SPM_CACHE_LOCATION="$HOME/Library/Caches/org.swift.swiftpm"
+SPM_CACHE_LOCATION="${HOME}/Library/Caches/org.swift.swiftpm"
 
 # Try to guess what to do if no option provided explicitly
 if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "true" ]]; then
@@ -78,12 +78,12 @@ fi
 
 # Restore SPM cache if it's available
 echo "~~~ Restoring SPM cache if available"
-PACKAGE_RESOLVED_HASH=$(hash_file "$PACKAGE_RESOLVED_LOCATION")
+PACKAGE_RESOLVED_HASH=$(hash_file "${PACKAGE_RESOLVED_LOCATION}")
 CACHE_KEY="${BUILDKITE_PIPELINE_SLUG}-spm-cache-${PACKAGE_RESOLVED_HASH}"
 
-mkdir -p "$SPM_CACHE_LOCATION"
-cd "$SPM_CACHE_LOCATION"
-restore_cache "$CACHE_KEY"
+mkdir -p "${SPM_CACHE_LOCATION}"
+cd "${SPM_CACHE_LOCATION}"
+restore_cache "${CACHE_KEY}"
 cd -
 
 # This will let Xcode use the system SSH config for downloading packages
@@ -119,8 +119,8 @@ fi
 # certain packages to have the artifacts already present after extracting 
 # cache
 echo "~~~ Cleaning up cache files before saving cache"
-rm -rf "$SPM_CACHE_LOCATION/checkouts" "$SPM_CACHE_LOCATION/artifacts"
+rm -rf "${SPM_CACHE_LOCATION}/checkouts" "${SPM_CACHE_LOCATION}/artifacts"
 
 # If this is the first time we've seen this particular cache key, save it for the future
 echo "~~~ Saving SPM Cache"
-save_cache "$SPM_CACHE_LOCATION" "$CACHE_KEY" false --use_relative_path_in_tar
+save_cache "${SPM_CACHE_LOCATION}" "${CACHE_KEY}" false --use_relative_path_in_tar
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -6,7 +6,7 @@ print_usage() {
   echo "    $0 --workspace PATH --scheme NAME"
   echo "  Using the \`Package.resolved\` of an Xcode \`.xcodeproj\`:"
   echo "    $0 --project PATH"
-  echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` (and not Xcode)"
+  echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` instead of Xcode:"
   echo "    $0 --use-spm"
 }
 
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -3,11 +3,11 @@
 print_usage() {
   echo "Usage:"
   echo "  Using the \`Package.resolved\` of an Xcode \`.xworkspace\`:"
-  echo "    $0 --workspace PATH"
+  echo "    ${0##*/} --workspace PATH"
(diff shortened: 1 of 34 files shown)


bin/install_swiftpm_dependencies
```diff
@@ -1,30 +1,91 @@
...
+
 SPM_CACHE_LOCATION="$HOME/Library/Caches/org.swift.swiftpm"
 
-WORKSPACE_PACKAGE_RESOLVED_PATH=(*.xcworkspace/xcshareddata/swiftpm/Package.resolved)
-ROOT_PACKAGE_RESOLVED_PATH=Package.resolved
+# Try to guess what to do if no option provided explicitly
+if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "true" ]]; then
+  echo "No \`--workspace\`, \`--project\` or \`--use-spm\` flag provided. Trying to guess the correct one to use..."
+  FOUND_ROOT_WORKSPACE_PATHS=(*.xcworkspace)
+  if  [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 1 && -d "${FOUND_ROOT_WORKSPACE_PATHS[0]}" && ! -f "Package.resolved" ]]; then
```
AliSoftware: 💭  Wondering if it would semantically make more sense to check the presence/absence of `Package.swift` to test if the repo is full-SPM-and-no-xcode-project, instead of testing `Package.resolved` (here + on line 56 + on line 58) 🤔

bin/install_swiftpm_dependencies
```diff
@@ -37,29 +98,34 @@ sudo defaults write com.apple.dt.Xcode IDEPackageSupportUseBuiltinSCM YES
...
-  echo "	Resolving packages with \`swift\`"
+# Resolve the packages using the correct method
+if [[ -d "${XCWORKSPACE_PATH}" ]]; then
+  echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
+  if [[ -z "${SCHEME_NAME}" ]]; then
+    SCHEME_NAME=$(xcodebuild -workspace "${XCWORKSPACE_PATH}" -onlyUsePackageVersionsFromResolvedFile -skipPackageUpdates -list | sed -n '/Schemes:/{n;s/^ *//p;}')
+    echo "No \`--scheme\` parameter provided; assuming the first one found (as it shouldn't matter in practice?): \`${SCHEME_NAME}\`"
+  fi
+  echo "Using -workspace \"${XCWORKSPACE_PATH}\" -scheme \"${SCHEME_NAME}\""
+  xcodebuild -workspace "${XCWORKSPACE_PATH}" -scheme "${SCHEME_NAME}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
```
AliSoftware: 💡 🤞   I'll have to do some more tests(†) to validate this, but from my quick experiments it seems that when I ran `xcodebuild -workspace "${XCWORKSPACE_PATH}" -list` (initially to guess a scheme to use)… `xcodebuild` printed "Resolve Package Graph", suggesting that it was resolving dependencies first before finally printing the list of schemes…

Which might mean that we could use this to our advantage, to just pass `-list` instead of having to pass an explicit `-scheme NAME` (i.e. `xcodebuild -workspace -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list`)… which would completely avoid the need to pass a `-scheme` to the command in the first place (just because of the bug of `xcodebuild` requiring `-scheme` [or `-list`] to be provided when `-workspace` is also provided… even in cases where the scheme doesn't matter, like iinm in the case of resolving dependencies)

---

(†) we should test this theory thoroughly though, especially with some edge cases—like deliberately have an outdated `Package.resolved` pointing to older versions, and/or ensuring the packages are not present yet locally, before running those commands—to check especially if the `-onlyUsePackageVersionsFromResolvedFile` is honored by `xcodebuild` (as opposed to being ignored when we use `-list` instead of an explicit `-scheme NAME`).

---

**But if this ends up behaving like I hope, this means that we could simplify the whole code of this bash script by**:
 - Removing the `--scheme` input parameter parsing from this utility
 - Directly invoke `xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list`—in order to resolve the Swift Dependencies with `xcodebuild` while **still ensuring it'd use the `.xcworkspace`** and not `.xcodeproj`, yet **without having to provide a `-scheme`** option

bin/install_swiftpm_dependencies
```diff
@@ -1,30 +1,91 @@
...
 
-CUSTOM_PACKAGE_RESOLVED_PATH=${1-}
-BUILD_TYPE=${2-}
+print_usage() {
+  echo "Usage:"
+  echo "  Using the \`Package.resolved\` of an Xcode \`.xworkspace\`:"
+  echo "    $0 --workspace PATH --scheme NAME"
+  echo "  Using the \`Package.resolved\` of an Xcode \`.xcodeproj\`:"
+  echo "    $0 --project PATH"
+  echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` (and not Xcode)"
```
mokagio: Nitpick. I feel this version "looks better"

```suggestion
  echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` instead of Xcode:"
```

Notice the `:` at the end. If the reword is not of your liking, we'd still "need" the `:` for consistency with the other lines.

```suggestion
  echo "  Using the \`Package.resolved\` at the root, managed by \`swift package\` (and not Xcode):"
```

bin/install_swiftpm_dependencies
```diff
@@ -1,30 +1,91 @@
...
+      exit 0
+      ;;
+    *)
+      echo "Unknown option or unexpected argument: $1"
+      print_usage
+      exit 2
+      ;;
+	esac
+	shift
+done
```
mokagio: Looks like the indentation here uses a mix of tabs and spaces, where the rest of the diff uses spaces.

bin/install_swiftpm_dependencies
mokagio: Using `Package.swift` would also make this work for projects that have forgotten to track `Package.resolved`. Granted, that should be a rare occurrence, but it's nice to know it would work in that case, too.

bin/install_swiftpm_dependencies
```diff
@@ -1,30 +1,91 @@
...
-WORKSPACE_PACKAGE_RESOLVED_PATH=(*.xcworkspace/xcshareddata/swiftpm/Package.resolved)
-ROOT_PACKAGE_RESOLVED_PATH=Package.resolved
+# Try to guess what to do if no option provided explicitly
+if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "true" ]]; then
+  echo "No \`--workspace\`, \`--project\` or \`--use-spm\` flag provided. Trying to guess the correct one to use..."
+  FOUND_ROOT_WORKSPACE_PATHS=(*.xcworkspace)
+  if  [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 1 && -d "${FOUND_ROOT_WORKSPACE_PATHS[0]}" && ! -f "Package.resolved" ]]; then
+    XCWORKSPACE_PATH="${FOUND_ROOT_WORKSPACE_PATHS[0]}"
+    echo " --> Found a single \`.xcworkspace\` file at the root of the repo (and no root \`Package.resolved\`)"
+    echo "     so we are assuming \`--workspace \"${XCWORKSPACE_PATH}\"\`."
```
mokagio: What do you think of:

```suggestion
    echo "--> Found a single \`.xcworkspace\` file and no \`Package.resolved\` in the root of the repo."
    echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
```

The pattern I'm suggesting here: splits the sentence in two, removing the "so", and move the "in the root of the repo" at the end of the sentence. _Just a nitpick_

bin/install_swiftpm_dependencies
```diff
@@ -1,30 +1,91 @@
...
+  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.resolved" ]]; then
+    echo " --> Found a \`Package.resolved\` at the root of the repo (and no root \`.xcworkspace\`)"
+    echo "     so we are assuming \`--use-spm\`"
+    USE_SPM=true
+  else
+    echo " -!- No valid --workspace & --scheme, or --project, or --use-spm flag provided, and cannot guess which one to use either, so aborting."
```
mokagio: `-!-` reminded me of the Reek

bin/install_swiftpm_dependencies
```diff
@@ -37,29 +98,34 @@ sudo defaults write com.apple.dt.Xcode IDEPackageSupportUseBuiltinSCM YES
...
-  echo "	Resolving packages with \`xcodebuild\`"
-  xcodebuild -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
-elif [[ "$BUILD_TYPE" == "SWIFT" ]]; then
-  echo "	Resolving packages with \`swift\`"
+# Resolve the packages using the correct method
+if [[ -d "${XCWORKSPACE_PATH}" ]]; then
+  echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
+  if [[ -z "${SCHEME_NAME}" ]]; then
+    SCHEME_NAME=$(xcodebuild -workspace "${XCWORKSPACE_PATH}" -onlyUsePackageVersionsFromResolvedFile -skipPackageUpdates -list | sed -n '/Schemes:/{n;s/^ *//p;}')
+    echo "No \`--scheme\` parameter provided; assuming the first one found (as it shouldn't matter in practice?): \`${SCHEME_NAME}\`"
```
mokagio: Suggesting to remove the "as it shouldn't matter in practice?" parenthetic because, while it makes sense as an observation in the implementation (and as far as I have seen it is correct) I don't feel it adds value to the logs. The caller didn't pass a scheme, so they'll have to deal with whichever one we default to for them.

```suggestion
    echo "No \`--scheme\` parameter provided. Using the first one found in the list: \`${SCHEME_NAME}\`"
```

bin/install_swiftpm_dependencies
mokagio: >  but from my quick experiments it seems that when I ran xcodebuild -workspace "${XCWORKSPACE_PATH}" -list (initially to guess a scheme to use)… xcodebuild printed "Resolve Package Graph", suggesting that it was resolving dependencies first before finally printing the list of schemes…

That's what I observe on my end, too. In Pocket Casts iOS:

```
➜ xcodebuild -workspace "podcasts.xcworkspace" -list
Command line invocation:
    /Applications/Xcode-16.0.0-Beta.3.app/Contents/Developer/usr/bin/xcodebuild -workspace podcasts.xcworkspace -list

User defaults from command line:
    IDEPackageSupportUseBuiltinSCM = YES

Resolve Package Graph


Resolved source packages:
  AutomatticTracksiOS: https://github.com/Automattic/Automattic-Tracks-iOS @ 3.4.2
  SwiftProtobuf: https://github.com/apple/swift-protobuf.git @ 1.22.1
  GTMAppAuth: https://github.com/google/GTMAppAuth.git @ 4.1.1
  Kingfisher: https://github.com/onevcat/Kingfisher @ 7.10.2
  TinyCSV: https://github.com/dagronf/TinyCSV @ 1.0.0
  JLRoutes: https://github.com/joeldev/JLRoutes @ 2.1.1
  Agrume: https://github.com/Automattic/Agrume @ 5.6.13
  Sentry: https://github.com/getsentry/sentry-cocoa @ 8.29.1
  SwiftSubtitles: https://github.com/dagronf/SwiftSubtitles @ 1.5.2
  DSFRegex: https://github.com/dagronf/DSFRegex @ 3.4.0
  Firebase: https://github.com/firebase/firebase-ios-sdk.git @ 10.25.0
  GoogleSignIn: https://github.com/google/GoogleSignIn-iOS @ 7.1.0
  abseil: https://github.com/google/abseil-cpp-binary.git @ 1.2024011601.1
  AppAuth: https://github.com/openid/AppAuth-iOS.git @ 1.7.5
  DifferenceKit: https://github.com/ra1028/DifferenceKit @ 1.2.0
  UIDeviceIdentifier: https://github.com/squarefrog/UIDeviceIdentifier @ 2.1.0
  AppCheck: https://github.com/google/app-check.git @ 10.19.0
  leveldb: https://github.com/firebase/leveldb.git @ 1.22.2
  GoogleAppMeasurement: https://github.com/google/GoogleAppMeasurement.git @ 10.25.0
  InteropForGoogle: https://github.com/google/interop-ios-for-google-sdks.git @ 100.0.0
  BuildkiteTestCollector: https://github.com/buildkite/test-collector-swift @ 0.1.1
  GoogleUtilities: https://github.com/google/GoogleUtilities.git @ 7.12.1
  Lottie: https://github.com/airbnb/lottie-ios.git @ 4.4.0
  Promises: https://github.com/google/promises.git @ 2.3.1
  PocketCastsServer: /Users/gio/Developer/a8c/pcios/Modules/Server
  GoogleDataTransport: https://github.com/google/GoogleDataTransport.git @ 9.3.0
  PocketCastsDataModel: /Users/gio/Developer/a8c/pcios/Modules/DataModel
  Sodium: https://github.com/jedisct1/swift-sodium @ 0.9.1
  gRPC: https://github.com/google/grpc-binary.git @ 1.62.2
  nanopb: https://github.com/firebase/nanopb.git @ 2.30909.0
  Fuse: https://github.com/krisk/fuse-swift @ 1.4.0
  FMDB: https://github.com/ccgus/fmdb.git @ 2.7.8
  GTMSessionFetcher: https://github.com/google/gtm-session-fetcher.git @ 3.4.1
  PocketCastsUtils: /Users/gio/Developer/a8c/pcios/Modules/Utils
  SwipeCellKit: https://github.com/shiftyjelly/SwipeCellKit @ 2.7.7

Information about workspace "podcasts":
    Schemes:
        BuildkiteTestCollector
        DSFRegex
        google-cast-sdk-no-bluetooth
        NotificationContent
        NotificationExtension
        Pocket Casts Prototype Build
        Pocket Casts Staging
        Pocket Casts Watch App
        pocketcasts
        PocketCastsDataModel
        PocketCastsServer
        PocketCastsUtils
        PodcastsIntents
        PodcastsIntentsUI
        Pods-PocketCastsTests
        Pods-podcasts
        PreBuildActions
        PulseCore
        PulseUI
        Screenshot Automation
        Screenshot Automation Watch
        Share Extension
        SwiftGen
        SwiftLint
        WidgetExtension
```

mokagio (commented):
Leaving some comments give this is a draft, still

mokagio:
I took this branch for a swing in Pocket Casts iOS – for other readers reference: that's the repo that originally gave us problems because of the double `Package.resolved` – and the automatic resolver worked smoothly:

https://buildkite.com/automattic/pocket-casts-ios/builds/7680#01912b6b-222e-4f57-bfe5-09cdc9038bff

bin/install_swiftpm_dependencies
AliSoftware: Fixed in 72b73ae

bin/install_swiftpm_dependencies
AliSoftware: Fixed in e220c485bfa5fe22e861469399a4f1686ae5f371

bin/install_swiftpm_dependencies
AliSoftware: Nice catch! Fixed in 407c0b814f44967d7dc181e98a36fa17f800e350

AliSoftware:
@mokagio Ok I've updated the PR to address all your feedback & remove the need for `--scheme`, to instead directly use `-list` to workaround `xcodebuild -workspace` complaining about `-scheme` missing otherwise. This seems to work as expected.

What I'm unsure about now, is if not providing a `-clonedSourcePackagesDirPath` explicitly will work well in concert with us building the app with a custom `DerivedData` on CI (especially via our fastlane's `build_app` calls).
For example, if we call `install_swiftpm_dependencies` and it doesn't provide a `-clonedSourcePackagesDirPath`. it will clone the checkout in `~/Library/Developer/Xcode/DerivedData/{projectname}-{somestring}/SourcePackages/checkouts`, the same that `Xcode.app` uses during "Resolve Package Dependencies"
 - But then if we call `build_app` from `fastlane` with a custom `DerivedData` (which ends up using `xcodebuild -derivedDataPath <custompath>`)… will this find the packages that were prefetched by `install_swiftpm_dependencies` given it's using a different `DerivedData` after all?
 - I haven't tested if passing `-derivedDataPath` to the `xcodebuild -resolvePackageDependencies` call of `install_swiftpm_dependencies` would magically make the `-clonedSourcePackagesDirPath` be derived from it (to `<provided-DerivedDataPath>/SourcePackages`?) if not provided explicitly, or if they are independent, or if one always have to explicitly provide `-clonedSourcePackagesDirPath` or whatnot… but wondering if all that warrants more testing in the end.

Besides, what `install_swiftpm_dependencies` currently zips and saves in our cache is always the content of `#{SPM_CACHE_LOCATION}` in practice. But I wonder if that's the right folder to cache and if it's working as expected when we end up using `xcodebuild -resolvePackageDependencies` as opposed to `swift package resolve` 🤔  Or maybe we should pass `-clonedSourcePackagesDirPath "${SPM_CACHE_LOCATION}"` to our call to `xcodebuild`? Or provide an extra option to `install_swiftpm_dependencies` to specify a custom `--derivedDataPath` (and use `-clonedSourcePackagesDirPath ${DERIVED_DATA_PATH}/SourcePackages` in our call to `xcodebuild` and zip that folder instead of `SPM_CACHE_LOCATION`)?!

AliSoftware:
Ok, looking a bit more into it, I think the `checkouts/` folder was misleading me (that's what you get when trying to make your brain work very late in the day 😅 ), and what's really important and the right source of truth is probably the **_bare_** repo (i.e. what looks like the result of a `git clone --bare`, with the direct content of the `.git/` repo with `refs`, `objects`, etc) that is cloned (in `~/Library/Developer/Xcode/DerivedData/…/SourcePackages/repositories` and/or in `~/Library/Caches/org.swift.swiftpm`).

Indeed I think it's likely that both `swift package` and `xcodebuild -resolvePackageDependencies` use that bare repo as source of truth and to regenerate the `checkouts` folders—as the original comment from @spencertransier suggests at the end of the `install_swiftpm_dependencies` script when he initially wrote it, where we delete `checkouts` and `artifacts` folders as they can apparently be recreated from `repositories`.

Doing a `rm -rf ~/Library/Caches/org.swift.swiftpm` + a `rm -rf ~/Library/Developer/Xcode/DerivedData/pocketcasts-hlevzbwzwxgpqbfpbknqqyqkauze/` to remove both and then calling `xcodebuild -workspace podcasts.xcworkspace -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list` ends up recreating both `~/Library/Caches/org.swift.swiftpm/repositories` and `~/Library/Developer/Xcode/DerivedData/pocketcasts-hlevzbwzwxgpqbfpbknqqyqkauze/SourcePackages/respositories`, both with the exact same content (verified with `opendiff`).

Similarly, `rm -rf` it all again then doing `xcodebuild -workspace podcasts.xcworkspace -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -clonedSourcePackagesDirPath ./MySourcePackages -list` and comparing `~/Library/Caches/org.swift.swiftpm/respositories` and `MySourcePackages/repositories` with `opendiff` confirms they are identical.

Finally only `rm -rf MySourcePackages` but keeping `~/Library/Caches/org.swift.swiftpm` untouched, and then re-running `xcodebuild -workspace podcasts.xcworkspace -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -clonedSourcePackagesDirPath ./MySourcePackages -list` I can confirm that this doesn't fetch the repos from the net but uses the local cache, as proven by the fact that the operation is quite faster and that the logs don't contain `Fetching from https://github.com/{org}/{repo}.git` logs like they did when I removed `~/Library/Caches/org.swift.swiftpm`, but only show `Creating working copy of package ‘{packagename}’` and `Checking out {version} of package ‘{packagename}’ in those logs 🎉 

---

So TL;DR I guess I just got confused by it being late here and not knowing how it was working with the difference between `checkouts` vs [bare] `repositories`, but in the end I think we're all good after all.
After all this investigation I think the caching per-se (and its goal to not have the CI fetch repositories from the web but from local cache instead) should work as expected regardless of the project using a custom `DerivedData`, and that we're still always zipping and storing the right `$SPM_CACHE_LOCATION` even for when we're using `xcodebuild -resolvePackages` after all 🎉 

(Phew, what an adventure.)

AliSoftware:
Now that the Integration Tests from #109 have been merged into this PR, this is now ready for a final review before auto-merge! cc @mokagio @spencertransier

mokagio (approved):
:shipit:

PR #108, opened 2024-08-09
Add integration tests for `install_swift_dependencies`

mokagio:
See the `README.md` in the new `tests/spm_caching` folder

> [!NOTE]
> This PR builds on top of https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/105

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #109, opened 2024-08-29
Add integration tests for `install_swiftpm_dependencies` 

mokagio:
See the `README.md` in the new `tests/spm_caching` folder

> [!NOTE]
> Builds on top of #105 and supersedes #108.

I was not happy with how much noise there was in the diff with the `project.pbxproj` files, so I decided to use code generation for them.

I chose `xcodegen` because we use it in at least one other (not-yet-open-source) app. Unfortunately, by the time I realized `xcodegen` only generates projects and not workspaces (see https://github.com/yonaskolb/XcodeGen/issues/168) I had already invested enough time in the set up to not justify starting from scratch. The bulk of the noise has been removed anyway thanks to not having to track the `.xcodeproj` folders.

## Where to go from here

- These tests are just for the happy paths. We could add more simulations, such as where the command runs in an empty folder and fails. 
- We could also consider using [Tuist](https://docs.tuist.io/guides/develop/projects#how-does-it-work) which supports creating a workspace to further DRY

---

At the time of writing, the lint and tests tasks failed, but because of rate liminting

image

Reading the `xcodegen` for this PR made me notice a docs typo, for which I opened an easy-fix PR  https://github.com/yonaskolb/XcodeGen/pull/1499

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. – N.A.


.buildkite/pipeline.yml
```diff
@@ -1,6 +1,42 @@
...
 steps:
   - label: "🕵️ Lint"
     command: make lint
+    agents:
+      queue: default
 
   - label: "🔬 Test"
     command: make test
+    agents:
+      queue: default
```
mokagio, line 16: By the way, these have been running on the `upload` queue all this time because of how the pipeline was configured in the IaC setup.

bin/install_swiftpm_dependencies
```diff
@@ -45,13 +45,22 @@ SPM_CACHE_LOCATION="${HOME}/Library/Caches/org.swift.swiftpm"
...
     XCWORKSPACE_PATH="${FOUND_ROOT_WORKSPACE_PATHS[0]}"
     echo " --> Found a single \`.xcworkspace\` file, and no \`Package.swift\` in the root of the repo."
     echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
-  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.swift" ]]; then
-    echo " --> Found a \`Package.swift\`, and no \`.xcworkspace\` in the root of the repo."
+  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && ${#FOUND_ROOT_PROJECT_PATHS[@] } -eq 1 && ! -f "Package.swift" ]]; then
+    XCODEPROJ_PATH="${FOUND_ROOT_PROJECT_PATHS[0]}"
+    echo " --> Found an \`.xcodeproj\`, and no \`Package.swift\` or \`.xcworkspace\` or \`.xcodeproj\` in the root of the repo."
+    echo "     Defaulting to \`--project \"${XCODEPROJ_PATH}\`"
+    USE_SPM=true
```
mokagio:
```suggestion
```

bin/install_swiftpm_dependencies
```diff
@@ -45,13 +45,21 @@ SPM_CACHE_LOCATION="${HOME}/Library/Caches/org.swift.swiftpm"
...
+  if [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 1 && -d "${FOUND_ROOT_WORKSPACE_PATHS[0]}" && ! -f "Package.swift" ]]; then
     XCWORKSPACE_PATH="${FOUND_ROOT_WORKSPACE_PATHS[0]}"
     echo " --> Found a single \`.xcworkspace\` file, and no \`Package.swift\` in the root of the repo."
     echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
-  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.swift" ]]; then
-    echo " --> Found a \`Package.swift\`, and no \`.xcworkspace\` in the root of the repo."
+  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && ${#FOUND_ROOT_PROJECT_PATHS[@] } -eq 1 && ! -f "Package.swift" ]]; then
+    XCODEPROJ_PATH="${FOUND_ROOT_PROJECT_PATHS[0]}"
+    echo " --> Found an \`.xcodeproj\`, and no \`Package.swift\` or \`.xcworkspace\` or \`.xcodeproj\` in the root of the repo."
+    echo "     Defaulting to \`--project \"${XCODEPROJ_PATH}\`"
```
mokagio: The tests made me noticed this path was not handled 😄

AliSoftware: Yeah, this was "kinda intentional" (as in, I initially didn't bother covering this case to avoid adding complexity to the implementation logic and because I didn't think it was worth it given that in practice and to my knowledge we don't have any project that have a sole `.xcodeproj` without having a `xcworkspace` alongside it.

But looking at the added code, and also now that we have integration tests to ensure said implementation is correct, we might as well covering this path after all indeed 👍

bin/install_swiftpm_dependencies
```diff
@@ -45,13 +45,21 @@ SPM_CACHE_LOCATION="${HOME}/Library/Caches/org.swift.swiftpm"
 # Try to guess what to do if no option provided explicitly
 if [[ -z "${XCWORKSPACE_PATH}" && -z "${XCODEPROJ_PATH}" && "${USE_SPM}" != "true" ]]; then
   echo "No \`--workspace\`, \`--project\` or \`--use-spm\` flag provided. Trying to guess the correct one to use..."
+
+  shopt -s nullglob
```
AliSoftware, line 49: 👍

tests/install_swiftpm_dependencies/README.md
```diff
@@ -0,0 +1,8 @@
+# Integration tests for `install_swiftpm_dependencies`
+
+This folder contains a variety of Xcode and Swift Package projects to test the behavior of the `install_swiftpm_dependencies` in various configurations.
+
+Note that currently with "testing" we mean: making sure the CI build doesn't fail. There isn't yet an automated setup to verify the correct command behavior.
+
+Also note that the tests are intentionally at the integration / end-to-end level. We could structure `install_swiftpm_dependencies` so that it spits out the `xcodebuild` or `swift` it plans to run and test that, but we would lose important feedback on how `xcodebuild` and `swift` themselves behave.
+Feedback on Apple's tooling behavior is crucial to get, given their known quirks.
```
AliSoftware, line 8: 💯

tests/install_swiftpm_dependencies/test_scripts/run_tests_with_xcodebuild.sh
```diff
@@ -0,0 +1,10 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+echo "--- :xcode: Run tests to verify packages have been fetched and are available"
```
AliSoftware, line 5: _[EDIT] Realized later that purpose of the test was to validate the flags (or lack thereof) passed to `install_swiftpm_dependencies` used the correct `Package.resolved`, rather than the purpose of those tests being to ensure the cache was used. My comment might thus not be fitting there after all.[/EDIT]_

---

We should find a way to verify that the SwiftPM cache has indeed been pulled and put in place prior to running `xcodebuild`, or alternatively find a way to prevent `xcodebuild` to access to the network, so that the CI and test would fail if it were to try. Because here I'm not sure that the test would actually make CI fail if the cache from `install_swiftpm_dependencies` didn't work as expected, but would probably instead just download the packages from internet in such case instead?

---

Maybe using flags like `-skipPackageUpdates` and `-onlyUsePackageVersionsFromResolvedFile` could help make this not access the network, though I doubt it'd go as far as failing in case of cache miss.

Another idea I had was to use `-dry-run` option of `xcodebuild` and somehow inspect the result; but from a quick test, this dry-run mode doesn't seem to apply to swift package resolution step; besides, as you point out in the README it's probably better do do integration tests than rely on checking the expected command.

A 3rd option that I thought of but wouldn't work would be using `networksetup -setnetworkserviceenabled <service> off` before running the `xcodebuild` command… but I am pretty sure that would have side effects (like cutting the connection to the buildkite-agent etc), so that's probably a hard no.

Finally, maybe there's some magic environment variable that `xcodebuild` accounts for and would disable network internally for it.
 - Either some hidden (and potentially poorly documented) env var specific to Xcode exists for this (Xcode is known to have some of those secret env vars that could tweak the behavior… but not sure where to find such one about network if one even exists)
 - Or using some more common system env var or setup to break network calls; I'm thinking about things like maybe setting an invalid `HTTP_PROXY`/`HTTPS_PROXY` value, or tweaking the SSH config or git commands (potentially via env vars, like `GIT_SSH_COMMAND`) to make it fail if Xcode were trying to fetch those…?

So far I had no luck with any of those ideas though.

AliSoftware: Mhh ok I had some luck with using fake `HTTP_PROXY` as workaround…
```
$ HTTP_PROXY=test-expects-no-network-access HTTPS_PROXY=test-expects-no-network-access xcodebuild -workspace podcasts.xcworkspace -scheme pocketcasts -resolvePackageDependencies

xcodebuild: error: Could not resolve package dependencies:
  Couldn’t fetch updates from remote repositories:
    fatal: unable to access 'https://github.com/dagronf/SwiftSubtitles/': Could not resolve proxy: test-expects-no-network-access
  binary target 'absl' could not be mapped to an artifact with expected name 'absl'
  binary target 'FirebaseFirestoreInternal' could not be mapped to an artifact with expected name 'FirebaseFirestoreInternal'
  binary target 'GoogleAppMeasurementOnDeviceConversion' could not be mapped to an artifact with expected name 'GoogleAppMeasurementOnDeviceConversion'
  binary target 'openssl_grpc' could not be mapped to an artifact with expected name 'openssl_grpc'
  binary target 'Sentry' could not be mapped to an artifact with expected name 'Sentry'
  binary target 'Clibsodium' could not be mapped to an artifact with expected name 'Clibsodium'
```
But… that also fails if the `~/Library/Cache` exists while the DerivedData doesn't exist yet (which is the case we want to test to ensure it uses the cache instead of network…), I think because even if it uses the cache for repository cloning, it still expects to check with the remote to validate the locale cache is not stale.

So overall not sure how we can make the test ensure cache is used and not skipped 🤔

tests/install_swiftpm_dependencies/test_scripts/test_project_automatic.sh
```diff
@@ -0,0 +1,18 @@
...
+echo "--- :computer: Jump to test folder"
+pushd "$TESTS_LOCATION/../project"
+
+brew install xcodegen
+make
+# Also resolve packages to generate Package.resolved in expected location
+xcodebuild -resolvePackageDependencies -project Demo.xcodeproj
+
+echo "--- :wrench: Run install_swiftpm_dependencies"
+install_swiftpm_dependencies
```
AliSoftware: Doesn't it defeat the purpose of the test to run `xcodebuild -resolvePackageDependencies` **before** running `install_swiftpm_dependencies`? 🤔 💭

.buildkite/pipeline.yml
AliSoftware: Ooof. Nice catch.

tests/install_swiftpm_dependencies/test_scripts/run_tests_with_xcodebuild.sh
AliSoftware: Ok after taking a break and thinking back on this, I realized that the goal of those intregration tests was not really to validate that the subsequent `xcodebuild -resolvePackageDependencies` calls would use the cache instead of the network, but rather than the call to `install_swiftpm_dependencies` was using the right `Package.resolved` and especially the right one for the right workspace/project/package based on the input `--flags` used.

So maybe it's all ok after all and we can disregard my whole comments above…

tests/install_swiftpm_dependencies/test_scripts/test_project_automatic.sh
AliSoftware: Ok I now realize that the goal is to validate that `install_swiftpm+dependencies` uses the rght `Package.resolved` from the right path based on the `--flags` passed to it (or lack thereof), so maybe that makes sense after all…

But still, this feels like this initial `xcodebuild -resolvePackageDependencies -project Demo.xcodeproj` could easily generate a false positive on the integration test.

What do you think about instead just keep a `Fixture-Package.resolved` file in the `project/` folder, and replace this `xcodebuild -resolvePackageDependencies -project Demo.xcodeproj` with a `cp Fixture-Package.resolved Demo.xcodeproj/project.xcworkspace/xcshareddata/swiftpm/Package.resolved` so that we still have a Package.resolved but without invoking `xcodebuild` before calling the `install_swiftpm_dependencies` SUT and risking side effects?

Granted, that kinda violates what you mention in the `README` about testing Apple tools as integration tests, as we'd instead assume the path where those are generated by `xcodebuild`. But I think that's still a valid assumption to make and unlikely to change, and I'd be ok to do that if that allows us to avoid the risk of side effects and test false positives.

Besides, if we go the route of using `cp` to put the `Package.resolved` in place:
 - That will make the test more stable by using a fixture for the `Package.resolved`, while calling `xcodebuild -resolvePackageDependencies` to let Xcode resolve the dependencies and create it will potentially create a different `.resolved` file at different times (depending on the latest versions of the dependencies available online at the time), and thus make for harder to reproduce test failures (can't easily `git bisect` to try to repro failures on an old build for example)
 - That will allow use to also `cp` an `InvalidPackageFixture.resolved`(one that would fail compiling, could even just be a file with sole content being `let package = invalid`) in the `Demo.xcodeproj`, while copying the valid `PackageFixture.resolved` into the `Demo.xcworkspace/project.xcodeproj/…`, which would allow us to validate that `install_swiftpm_dependencies` would use the right one when testing it on the `workspace_and_project` integration test below…

AliSoftware:
> At the time of writing, the lint and tests tasks failed, but because of rate liminting

CI is indeed failing with:
```
docker: Error response from daemon: toomanyrequests: You have reached your pull rate limit. You may increase the limit by authenticating and upgrading: https://www.docker.com/increase-rate-limit
```

This is likely unrelated to your changes, but begs the question if there would be a way to pull this image from ECR instead of dockerhub to avoid this from happening or something. That being said, I took a look at the https://gallery.ecr.aws/ and didn't find the `buildkite/plugin-linter` image there.

I think it's ok for now; but if we start encountering this error more frequently, we might want to investigate more about alternate solutions (e.g. mirror that plugin in our own AWS account and ECR private registry and use that?). I'm just hoping that the mere fact that this PR adds a lot of CI jobs to the pipeline is not the cause for this, though I don't see how that would be the case, because we only invoke `docker` for `lint` and `test` jobs… 🤷

AliSoftware (commented):
After some back and forth in misunderstanding what we were trying to test, I think my only remaining suggestion/feedback is to `cp` a `Package.resolved` fixture within the `xcodeproj`/`xcworkspace` folders generated by `xcodegen`, as opposed to using `xcodebuild -resolvePackageDependencies` to generate those and potentially creating false positives on the tests themselves.

bin/install_swiftpm_dependencies
```diff
@@ -111,15 +119,15 @@ elif [[ "${USE_SPM}" == "true" ]]; then
...
-# 
-# `artifacts` should be removed because it causes issues when downloading 
-# certain packages to have the artifacts already present after extracting 
+#
+# `artifacts` should be removed because it causes issues when downloading
+# certain packages to have the artifacts already present after extracting
 # cache
 echo "~~~ Cleaning up cache files before saving cache"
 rm -rf "${SPM_CACHE_LOCATION}/checkouts" "${SPM_CACHE_LOCATION}/artifacts"
+echo "Done. Removed checkouts and artifacts subfolders from $SPM_CACHE_LOCATION"
```
mokagio, line 130: Noticed this nitpick while checking the logs in the integration tests... The group had no logs under it, which made expanding it look odd.

Before (without `echo`)

image

After

image

tests/install_swiftpm_dependencies/test_scripts/test_project_no_package.sh
```diff
@@ -0,0 +1,42 @@
...
+CMD_EXIT_STATUS=$?
+set -e
+
+EXPECTED="Unable to find \`Package.resolved\` file ($PROJECT/project.xcworkspace/xcshareddata/swiftpm/Package.resolved)"
+
+if [[ $CMD_EXIT_STATUS -eq 0 ]]; then
+  echo "+++ :x: install_swiftpm_dependencies unexpectedly succeeded without a Package.resolved in the project folder!"
+  echo "Expected: $EXPECTED"
+  echo "Got: $(cat $LOGS_PATH)"
+  exit 1
```
mokagio, line 31: See test-the-tests PR https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/110

image

tests/install_swiftpm_dependencies/test_scripts/test_project_no_package.sh
```diff
@@ -0,0 +1,42 @@
...
+else
+  if grep -qF "$EXPECTED" "$LOGS_PATH"; then
+    echo "^^^ +++"
+    echo "+++ :white_check_mark: install_swiftpm_dependencies failed as expected because there is no Package.resolved in the project folder."
+  else
+    echo "+++ :x: install_swiftpm_dependencies failed, but the message it printed is not what we expected."
+    echo "Expected: $EXPECTED"
+    echo "Got: $(cat $LOGS_PATH)"
+    exit 1
```
mokagio, line 40: Again, see https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/110

image

mokagio:
Thank you @AliSoftware for the `Package.resolved` fixture behavior and for the time you put in researching other possible ways of testing `install_swiftpm_dependencies`.

I took the fixture suggestion on board. It also made me realize it would be "easy" to verify the command fails if no `Package.resolved` is present in the `xcodeproj` SwiftPM folder.

I could have kept going adding more such tests, and maybe at some point I will, but I decided to add only a token one and see if we can get this merged. 😄

bin/install_swiftpm_dependencies
```diff
@@ -45,13 +45,21 @@ SPM_CACHE_LOCATION="${HOME}/Library/Caches/org.swift.swiftpm"
...
+
+  if [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 1 && -d "${FOUND_ROOT_WORKSPACE_PATHS[0]}" && ! -f "Package.swift" ]]; then
     XCWORKSPACE_PATH="${FOUND_ROOT_WORKSPACE_PATHS[0]}"
     echo " --> Found a single \`.xcworkspace\` file, and no \`Package.swift\` in the root of the repo."
     echo "     Defaulting to \`--workspace \"${XCWORKSPACE_PATH}\"\`."
-  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && -f "Package.swift" ]]; then
-    echo " --> Found a \`Package.swift\`, and no \`.xcworkspace\` in the root of the repo."
+  elif [[ ${#FOUND_ROOT_WORKSPACE_PATHS[@]} -eq 0 && ${#FOUND_ROOT_PROJECT_PATHS[@] } -eq 1 && ! -f "Package.swift" ]]; then
+    XCODEPROJ_PATH="${FOUND_ROOT_PROJECT_PATHS[0]}"
+    echo " --> Found an \`.xcodeproj\`, and no \`Package.swift\` or \`.xcworkspace\` or \`.xcodeproj\` in the root of the repo."
```
AliSoftware:
```suggestion
    echo " --> Found an \`.xcodeproj\`, and no \`Package.swift\` nor \`.xcworkspace\`in the root of the repo."
```

tests/install_swiftpm_dependencies/test_scripts/test_project_automatic.sh
```diff
@@ -0,0 +1,23 @@
...
+
+echo "--- :computer: Generate project"
+brew install xcodegen
+make
+
+echo "--- :computer: Copy Package.resolved fixture"
+PROJECT="Demo.xcodeproj"
+XCODE_SPM_PATH="$PROJECT/project.xcworkspace/xcshareddata/swiftpm"
+mkdir -p "$XCODE_SPM_PATH"
+cp "$TESTS_LOCATION/../package_resolved_fixtures/valid.resolved" "$XCODE_SPM_PATH/Package.resolved"
```
AliSoftware, line 18: 👍

tests/install_swiftpm_dependencies/test_scripts/test_project_no_package.sh
AliSoftware: Love it!

tests/install_swiftpm_dependencies/project/Makefile
```diff
@@ -0,0 +1,7 @@
+.DEFAULT_GOAL := generate
+
+generate:
+	xcodegen generate
+
+update_tracked_package_resolved:
+	xcodebuild -resolvePackageDependencies -project Demo.xcodeproj
```
AliSoftware: Now that we decided to `cp package_resolved_fixtures/valid.resolved` after `xcodegen generate`-ing the project, I'm guessing this is not needed anymore (and that one should use `make regenerate_fixtures` from the `tests/install_swiftpm_dependencies/Makefile` instead) and we should thus probably remove that one, right?

```suggestion
```

tests/install_swiftpm_dependencies/workspace_and_project/Makefile
```diff
@@ -0,0 +1,3 @@
+generate:
+	cp ../project/project.yml .
+	xcodegen generate
```
AliSoftware, line 5: Just to be consistent with how you do it in other `Makefile`s:
```suggestion
	xcodegen generate --spec ../project/project.yml
```

tests/install_swiftpm_dependencies/test_scripts/test_workspace_and_project_automatic.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/set_up_environment.sh"
+
+echo "--- :computer: Jump to test folder"
+pushd "$TESTS_LOCATION/../workspace_and_project"
+
+brew install xcodegen
```
AliSoftware, line 11:
```suggestion
echo "--- :computer: Generate project"
brew install xcodegen
```

tests/install_swiftpm_dependencies/test_scripts/test_workspace_and_project_explicit.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/set_up_environment.sh"
+
+echo "--- :computer: Jump to test folder"
+pushd "$TESTS_LOCATION/../workspace_and_project"
+
+brew install xcodegen
```
AliSoftware, line 11:
```suggestion
echo "--- :computer: Generate project"
brew install xcodegen
```

tests/install_swiftpm_dependencies/test_scripts/test_workspace_automatic.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/set_up_environment.sh"
+
+echo "--- :computer: Jump to test folder"
+pushd "$TESTS_LOCATION/../workspace"
+
+brew install xcodegen
```
AliSoftware, line 11:
```suggestion
echo "--- :computer: Generate project"
brew install xcodegen
```

tests/install_swiftpm_dependencies/test_scripts/test_workspace_explicit.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash -eu
+
+set -o pipefail
+
+source "$(dirname "${BASH_SOURCE[0]}")/set_up_environment.sh"
+
+echo "--- :computer: Jump to test folder"
+pushd "$TESTS_LOCATION/../workspace"
+
+brew install xcodegen
```
AliSoftware, line 11:
```suggestion
echo "--- :computer: Generate project"
brew install xcodegen
```

AliSoftware (commented):
LGTM; only found consistency nitpicks, will apply and validate it still works then merge.

tests/install_swiftpm_dependencies/workspace_and_project/Makefile
```diff
@@ -0,0 +1,2 @@
+generate:
+	xcodegen generate --spec ../project/project.yml
```
AliSoftware: Uh, interesting, I guess doing so broke it. I'm guessing this is the reason?
```suggestion
	# Note: we need to copy the project file in place (as opposed to using --spec)
	# so that xcodegen can add the scheme for the Demo project into the `Demo.xcworkspace`
	cp ../project/project.yml .
	xcodegen generate
```

mokagio: I should have documented it myself. Thanks! 👍

tests/install_swiftpm_dependencies/project/Makefile
mokagio: Correct. I clearly missed this cleanup. Thanks!

PR #111, opened 2024-09-05
Upload sarif report file to gh registry

wzieba:
### Description

This PR introduces a new script that sends a `sarif` file to GH registry using REST API. 

- [about-sarif-file-uploads-for-code-scanning](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/uploading-a-sarif-file-to-github#about-sarif-file-uploads-for-code-scanning)
- [REST API reference](https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#upload-an-analysis-as-sarif-data)
- Sample integration: https://github.com/wordpress-mobile/WordPress-Android/pull/21207
- ...and proof that report has been received: https://github.com/wordpress-mobile/WordPress-Android/security/code-scanning?query=is%3Aopen+pr%3A21207

---


bin/upload_sarif_to_gh
```diff
@@ -0,0 +1,44 @@
+#!/bin/bash -eu
+
+sarif_file=$1
+owner=$2
+repo=$3
+
+if [ -z "$sarif_file" ] || [ -z "$owner" ] || [ -z "$repo" ]; then
+  echo "No argument provided. Usage: ./upload_sarif_to_gh.sh <path to .sarif report> <gh owner> <gh repo>"
```
AliSoftware:
```suggestion
  echo "Not enough arguments provided. Usage: ./upload_sarif_to_gh.sh <path to .sarif report> <gh owner> <gh repo>"
```

bin/upload_sarif_to_gh
AliSoftware, line 1: May I suggest a more explicit `upload_sarif_to_github` name for this command?

bin/upload_sarif_to_gh
```diff
@@ -0,0 +1,44 @@
...
+  exit 1
+fi
+
+gzip -c "$sarif_file" | base64 >sarif_base64.tmp
```
AliSoftware: Could be worth adding more error handling. For example:
 - check that the `$sarif_file` exists / is a regular file / is readable before calling `gzip`
 - check early that `$GITHUB_TOKEN` is defined and not empty, to exit early instead of having the error in those cases only happen during `curl` at the end of the script.

You might also want to `set -o pipefail` (or update your shebang to `#!/bin/bash -euo pipefail`) so that errors in a `|` pipe (like here where `gzip` output is piped thru `base64`) are not swallowed by the pipe and ignored. To be honest, I'm actually surprised that `shellcheck` (that is run on CI as part of the call to `make lint` in this repo's pipeline) didn't recommend this when linting that script 🤔

bin/upload_sarif_to_gh
```diff
@@ -0,0 +1,44 @@
...
+    --arg branch "$BUILDKITE_BRANCH" \
+    --rawfile sarif sarif_base64.tmp \
+    '{
+     "commit_sha": $commit_sha,
+     "ref": ("refs/heads/$branch"),
+     "sarif": $sarif
+   }')
+fi
+
+echo "$json" > payload.json
```
AliSoftware: 💭  It could probably be worth writing this `payload.json` file in a temp directory rather than the current directory?

One could argue that it doesn't really matter because the CI will clean up the checkout at the end of the job anyway. But I'm thinking of edge cases like if the step calling that command ends up having `artifact_paths: **/*.json` set up in the pipeline for other reasons (e.g with the goal of collecting unrelated `json` files) and would end up also picking up that extra `payload.json`. Granted, this is unlikely, as I'd expect us to be more precise than a generic `**/*.json` for our artifacts paths (e.g. `builds/results/*.json`), but still, just out of principle

Another option would be to give this file a more explicit name (e.g. `sarif-file-upload-payload.json` or something), so that at least if it's picked up in the artifacts by the job calling this it will be clearer what it's supposed to be.

wzieba:
Thank you @AliSoftware for your suggestions! I think I applied all of them, TIL `mktemp`.

Sarif file upload seems to still work fine: https://buildkite.com/automattic/wordpress-android/builds/19758#0191c2f0-4126-4a92-8873-9cd82349888c/245-294

Ready for a second round

bin/upload_sarif_to_github
```diff
@@ -0,0 +1,60 @@
...
+
+if [ -z "$sarif_file" ] || [ -z "$owner" ] || [ -z "$repo" ]; then
+  echo "Not enough arguments provided. Usage: ./upload_sarif_to_gh.sh <path to .sarif report> <gh owner> <gh repo>"
+  exit 1
+fi
+
+if [ ! -r "$sarif_file" ]; then
+  echo "Error: The specified sarif file '$sarif_file' is not readable."
+  exit 1
+fi
```
AliSoftware, line 22:
```suggestion
if [ ! -f "$sarif_file" ]; then
  echo "Error: The specified sarif file '$sarif_file' does not exist or is not a regular file."
  exit 1
fi

if [ ! -r "$sarif_file" ]; then
  echo "Error: The specified sarif file '$sarif_file' is not readable."
  exit 1
fi
```

bin/upload_sarif_to_github
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash
+
+set -euo pipefail
+
+sarif_file=$1
+owner=$2
+repo=$3
+
+if [ -z "$sarif_file" ] || [ -z "$owner" ] || [ -z "$repo" ]; then
```
AliSoftware: If I'm not mistaken, the `set -u` from line 3 will make any of those fail if there is a missing parameter. Because `set -u` will raise an error if the var does not exist, and `$1` won't exist if you don't pass any parameter to the call site.

One way to fix this would be to remove `set -u` but I wouldn't recommend that, because that's a useful safety net to keep.
A better way to fix this would be to use `sarif_file="${1:-}"` instead of `$1` (and same for `$2` and `$3`) so that if `$1` is not defined (missing parameter when calling the script), then `sarif_file` will be set to the empty string—that you can then test using `-z` as you already do here—instead of the script failing during the assignment.

```suggestion
sarif_file="${1:-}"
owner="${2:-}"
repo="${3:-}"

if [ -z "$sarif_file" ] || [ -z "$owner" ] || [ -z "$repo" ]; then
```

bin/upload_sarif_to_github
```diff
@@ -0,0 +1,60 @@
...
+  exit 1
+fi
+
+if [ ! -r "$sarif_file" ]; then
+  echo "Error: The specified sarif file '$sarif_file' is not readable."
+  exit 1
+fi
+
+# Check that GITHUB_TOKEN is set
+if [ -z "$GITHUB_TOKEN" ]; then
```
AliSoftware: Same as above, `set -u` will make this fail and exit the script if `GITHUB_TOKEN` is not defined, so the trick is:
```suggestion
if [ -z "${GITHUB_TOKEN:-}" ]; then
```

bin/upload_sarif_to_github
wzieba: True 👍 I got from 

```cmd
./upload_sarif_to_github: line 5: $1: unbound variable
```

to 

```cmd
Not enough arguments provided. Usage: ./upload_sarif_to_gh.sh <path to .sarif report> <gh owner> <gh repo>
```

when I run script without expected values. Thanks!

AliSoftware (approved):
⚠️  Don't forget to add an entry for this new feature in our `CHANGELOG.md`.

Apart from this, LGTM, so approving to unblock.

PR #116, opened 2024-09-12
Check for `$BUILDKITE_PULL_REQUEST = false`  when uploading sarif file.

wzieba:
If build is not associated with a pull request, Buildkite puts value `false` for `$BUILDKITE_PULL_REQUEST`
Initially, I thought it just doesn't create such variable.

Proof can be found here: https://buildkite.com/automattic/pocket-casts-android/builds/8308#0191e617-a1e2-4ae6-a7bb-4ef3dda51370

Because of this bug, after merge to trunk in Pocket Casts we can see:
```
{
  | [2024-09-12T12:07:17Z]   "message": "ref 'refs/pull/false/head' not found in this repository",
  | [2024-09-12T12:07:17Z]   "documentation_url": "https://docs.github.com/rest/code-scanning/code-scanning#upload-an-analysis-as-sarif-data",
  | [2024-09-12T12:07:17Z]   "status": "404"
  | [2024-09-12T12:07:17Z] 
}
```

---

### DIFF
--- a/bin/upload_sarif_to_github
+++ b/bin/upload_sarif_to_github
@@ -31,7 +31,7 @@ sarif_base64_temp_file=$(mktemp)
 
 gzip -c "$sarif_file" | base64 > "$sarif_base64_temp_file"
 
-if [[ -n $BUILDKITE_PULL_REQUEST ]]; then
+if [[ $BUILDKITE_PULL_REQUEST != "false" ]]; then
   json=$(jq -n \
     --arg commit_sha "$BUILDKITE_COMMIT" \
     --arg pr_number "$BUILDKITE_PULL_REQUEST" \
--- a/bin/upload_sarif_to_github
+++ b/bin/upload_sarif_to_github
@@ -31,7 +31,7 @@ sarif_base64_temp_file=$(mktemp)
 
 gzip -c "$sarif_file" | base64 > "$sarif_base64_temp_file"
 
-if [[ $BUILDKITE_PULL_REQUEST != "false" ]]; then
+if [[ "${BUILDKITE_PULL_REQUEST:-false}" != "false" ]]; then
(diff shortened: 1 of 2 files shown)


bin/upload_sarif_to_github
```diff
@@ -31,7 +31,7 @@ sarif_base64_temp_file=$(mktemp)
 
 gzip -c "$sarif_file" | base64 > "$sarif_base64_temp_file"
 
-if [[ -n $BUILDKITE_PULL_REQUEST ]]; then
+if [[ $BUILDKITE_PULL_REQUEST != "false" ]]; then
```
wzieba: and maybe I should add

```sh
else
  exit 0
```

because we don't want to upload if it's not main branch or PR 🤔

bin/upload_sarif_to_github
```diff
@@ -51,6 +51,8 @@ elif [[ "$BUILDKITE_BRANCH" == "$BUILDKITE_PIPELINE_DEFAULT_BRANCH" ]]; then
      "ref": ("refs/heads/$branch"),
      "sarif": $sarif
    }')
+else
+  exit 0
```
AliSoftware, line 55: 👍

bin/upload_sarif_to_github
AliSoftware: Just for being extra paranoid, you could cover both cases (`false` and empty/not defined) by doing something like this:

```suggestion
if [[ "${BUILDKITE_PULL_REQUEST:-false}" != "false" ]]; then
```

Not that I'd expect Buildkite to suddenly change the way they hand the `BUILDKITE_PULL_REQUEST` env var in the future, but, you know, at least we'd cover our bases.

AliSoftware (approved):

PR #120, opened 2024-09-13
Comment with dependency diff

wzieba:
### Description

This PR introduces `comment_with_dependency_diff` script that is aiming to help Android product teams to better understand what changes to the dependency tree a PR might introduce.

We use similar scripts in WooCommerce and Day One projects. I want to extend support to other projects and share the implementation between them for easier maintenance.

Additionally, this PR introduces some improvements to the script:
- comment on PR without using Gradle (so our Android apps can remove another, not really needed dependency from their codebase)
- print build environment changes as well (previously, the script was only printing runtime dependencies changes)

To see the result, please see:
- Pocket Casts: https://github.com/Automattic/pocket-casts-android/pull/2857
- Matticspace Mobile (for the case of a large diff): https://github.com/Automattic/Matticspace-Mobile/pull/67

---

### DIFF
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -132,28 +132,32 @@ comment_body_file=$(mktemp)
 echo "$comment_body" > "$comment_body_file"
 
 # Use jq with the temporary file for payload
-json_payload_file=$(mktemp)
-jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}' > "$json_payload_file"
+#jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}' > "$json_payload_file"
 
-# Clean up temporary files
-rm "$comment_body_file"
+json=$(jq -c --null-input \
+  --rawfile body "$comment_body_file" \
+  '{
+    body: $body
+   }
+  '
+)
+
+json_payload_file=$(mktemp)
+echo "$json" > "$json_payload_file"
 
 if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
     echo "Update the existing comment: $existing_comment_id"
     github_api "$issues_endpoint/comments/$existing_comment_id" \
         --fail-with-body -X PATCH \
         -H "Content-Type: application/json" \
-        -d @"$json_payload_file"
+        --data-binary @"$json_payload_file"
 elif [[ -n "$arg_pr_comment" ]]; then
     echo "Post a new comment"
     github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
         --fail-with-body -X POST \
         -H "Content-Type: application/json" \
-        -d @"$json_payload_file"
+        --data-binary @"$json_payload_file"
 else
     # No comment body was given in CLI, so no new comment to post
     echo "No new comment to post"
 fi
-
-# Clean up json payload file
-rm "$json_payload_file"
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -126,7 +126,16 @@ $comment_body_id
 $arg_pr_comment
 EOF
 )
-json_payload=$(jq -c --null-input --arg body "$comment_body" '{body: $body}')
+
+# Write the comment body to a temporary file to fix possible "Argument list too long" error
+comment_body_file=$(mktemp)
+echo "$comment_body" > "$comment_body_file"
+
+# Use jq with the temporary file
+json_payload=$(jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}')
+
+# Clean up temporary files
+rm "$comment_body_file"
 
 if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
     echo "Update the existing comment: $existing_comment_id"
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -131,8 +131,9 @@ EOF
 comment_body_file=$(mktemp)
 echo "$comment_body" > "$comment_body_file"
 
-# Use jq with the temporary file
-json_payload=$(jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}')
+# Use jq with the temporary file for payload
+json_payload_file=$(mktemp)
+jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}' > "$json_payload_file"
 
 # Clean up temporary files
 rm "$comment_body_file"
@@ -142,14 +143,17 @@ if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
     github_api "$issues_endpoint/comments/$existing_comment_id" \
         --fail-with-body -X PATCH \
         -H "Content-Type: application/json" \
-        -d "${json_payload}"
+        -d @"$json_payload_file"
 elif [[ -n "$arg_pr_comment" ]]; then
     echo "Post a new comment"
     github_api "$issues_endpoint/${BUILDKITE_PULL_REQUEST}/comments" \
         --fail-with-body -X POST \
         -H "Content-Type: application/json" \
-        -d "${json_payload}"
+        -d @"$json_payload_file"
 else
     # No comment body was given in CLI, so no new comment to post
     echo "No new comment to post"
 fi
+
+# Clean up json payload file
+rm "$json_payload_file"
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -131,16 +131,11 @@ EOF
 comment_body_file=$(mktemp)
 echo "$comment_body" > "$comment_body_file"
 
-json=$(jq -c --null-input \
-  --rawfile body "$comment_body_file" \
-  '{
-    body: $body
-   }
-  '
-)
-
 json_payload_file=$(mktemp)
-echo "$json" > "$json_payload_file"
+jq -c --null-input \
+  --rawfile body "$comment_body_file" \
+  '{ body: $body }' \
+  > "$json_payload_file"
 
 if [[ -n "$existing_comment_id" && "$opt_if_exist" == "update" ]]; then
     echo "Update the existing comment: $existing_comment_id"
--- a/bin/comment_on_pr
+++ b/bin/comment_on_pr
@@ -131,9 +131,6 @@ EOF
 comment_body_file=$(mktemp)
 echo "$comment_body" > "$comment_body_file"
 
-# Use jq with the temporary file for payload
-#jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}' > "$json_payload_file"
-
 json=$(jq -c --null-input \
   --rawfile body "$comment_body_file" \
   '{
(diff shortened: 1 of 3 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/comment_on_pr
```diff
@@ -126,20 +126,37 @@ $comment_body_id
...
 )
-json_payload=$(jq -c --null-input --arg body "$comment_body" '{body: $body}')
+
+# Write the comment body to a temporary file to fix possible "Argument list too long" error
+comment_body_file=$(mktemp)
+echo "$comment_body" > "$comment_body_file"
+
+# Use jq with the temporary file for payload
+#jq -c --null-input --arg body "$(cat "$comment_body_file")" '{body: $body}' > "$json_payload_file"
+
```
AliSoftware: 🧹  I guess this is leftover from debugging and can be removed?

bin/comment_on_pr
```diff
@@ -126,20 +126,37 @@ $comment_body_id
...
+json=$(jq -c --null-input \
+  --rawfile body "$comment_body_file" \
+  '{
+    body: $body
+   }
+  '
+)
+
+json_payload_file=$(mktemp)
+echo "$json" > "$json_payload_file"
```
AliSoftware: 💄  Alternative: avoid intermediate `json` variable altogether:

```suggestion
json_payload_file=$(mktemp)
jq -c --null-input \
  --rawfile body "$comment_body_file" \
  '{ body: $body }' \
  > "$json_payload_file"
```

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
+#!/bin/bash
+
+set -euo pipefail
```
AliSoftware, line 31: 📚  The more we add utility scripts like those to this plugin, the more I realize we should have made a better job as providing help / documentation at what each of those do. Especially so that people not as familiar with our CI (e.g. devs that are interested in using some of those helper scripts in their pipelines themselves but have no idea what each utility does) can better understand the purpose of each.

For example, [the comment at the start of `comment_on_pr` is a good example of having some code-comment documentation on what the utility does and how it's expected to be used](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/120/files#diff-d536c5bec9be6d9bc743cc0175b25457a7417a97a3afded630db75505b91882cL3-R22)

I wish we took that habit sooner and applied it to all of our other `bin/*` utilities in this repo, making them all start with a code-comment explaining their role / context in which those helpers would be used and could help, documenting what the script does and its expected parameters and inputs (e.g. "this script should only run after `./gradlew …` has been called and generated the build artifacts", etc).
We should probably make a pass at all of them at some point.

In the meantime, wdyt about us already starting moving towards this good habit and add some intro doc comments in any new helpers we introduce, starting from this one?

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
...
+DIFF_DEPENDENCIES_FOLDER="./build/reports/diff"
+BASE_BRANCH_DEPENDENCIES_FILE="$DIFF_DEPENDENCIES_FOLDER/base_branch_dependencies.txt"
+HEAD_BRANCH_DEPENDENCIES_FILE="$DIFF_DEPENDENCIES_FOLDER/head_branch_dependencies.txt"
+BASE_BRANCH_BUILD_ENV_FILE="$DIFF_DEPENDENCIES_FOLDER/base_branch_build_env.txt"
+HEAD_BRANCH_BUILD_ENV_FILE="$DIFF_DEPENDENCIES_FOLDER/head_branch_build_env.txt"
+DIFF_DEPENDENCIES_FILE="$DIFF_DEPENDENCIES_FOLDER/diff_dependencies.txt"
+DIFF_BUILD_ENV_FILE="$DIFF_DEPENDENCIES_FOLDER/diff_build_env.txt"
+DEPENDENCY_TREE_VERSION="1.2.1"
+
+PR_NUMBER=$BUILDKITE_PULL_REQUEST
```
AliSoftware: 💡 Even if one is only expected to call this helper on a Buildkite step for which we set `if: build.pull_request.id != null`, could be worth adding some protection about it anyway—typically at the start of the script, after your tests for `-z "$MODULE"` / `-z "$CONFIGURATION"`. Like:

```bash
if [ "${BUILDKITE_PULL_REQUEST:-false}" == "false" ]; then
  echo "This script should only be called on pull requests. Be sure to set `if: build.pull_request.id != null` on the step invoking it in your buildkite pipeline"
  exit 2
fi
```

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
...
+PR_URL="https://api.github.com/repos/$OWNER/$REPO/pulls/$PR_NUMBER"
+
+git config --global user.email "mobile@automattic.com"
+git config --global user.name "Dependency Tree Diff Tool"
+
+echo "--> Starting the check"
+echo "--> Fetching the base branch from $PR_URL"
+githubResponse="$(curl "$PR_URL" -H "Authorization: token $GITHUB_TOKEN")"
```
AliSoftware: 💭  I have not checked if this would be strictly equivalent or not, but what do you think about using [our `github_api` helper from this repo](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/github_api) instead of calling `curl` directly, now that this script lives in the same repo/plugin? 🤔

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
...
+baseBranch=$(echo "$githubResponse" | tr '\r\n' ' ' | jq '.base.ref' | tr -d '"')
+echo "--> Base branch is $baseBranch"
```
AliSoftware: 🤔  Actually, is the purpose of this whole API call only to extract the name of the base (aka target) branch of the PR?

If so, why not just use [`BUILDKITE_PULL_REQUEST_BASE_BRANCH`](https://buildkite.com/docs/pipelines/environment-variables#BUILDKITE_PULL_REQUEST_BASE_BRANCH) directly?

_(I think I slight remember mentioning this a while ago when I reviewed the first iterations of this script back when it was added to the repositories themselves, and iirc the argument was mostly due to legacy with this script previously being also used in other CI, maybe while we were still in the middle of transitioning to Buildkite…? So maybe that's where it comes from?)_

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
...
+
+echo "--> Generating build environment dependencies for the base branch"
+./gradlew buildEnvironment >$BASE_BRANCH_BUILD_ENV_FILE
+
+git checkout "-"
+
+echo "--> Downloading dependency-tree-diff.jar"
+curl -v -L "https://github.com/JakeWharton/dependency-tree-diff/releases/download/$DEPENDENCY_TREE_VERSION/dependency-tree-diff.jar" -o dependency-tree-diff.jar
+sha=$(sha1sum dependency-tree-diff.jar | cut -d' ' -f1)
+if [[ $sha != "39c50f22cd4211a3e52afec43120b7fdf90f9890" ]]; then
```
AliSoftware: 💡  Nitpick: could be worth extracting this SHA in a `DEPENDENCY_TREE_CHECKSUM` constant to put next to line 23, so that it'd be easier to update to a new version in the future by changing the version and checksum together?

bin/comment_with_dependency_diff
```diff
@@ -0,0 +1,153 @@
...
+  true > $COMMENT_FILE  # 'true' used to clear the file as a no-op
+
+  append_content() {
+    local content="$1"
+    local github_pr_comment_max_size=65536
+    local current_file_size
+    current_file_size=$(wc -c < "$COMMENT_FILE")
+
+    if (( current_file_size + ${#content} > github_pr_comment_max_size )); then
+      echo "Content exceeds $github_pr_comment_max_size characters. Navigate to Buildkite build artifacts to see details." >> $COMMENT_FILE
```
AliSoftware: 💡  I think we could even add a link to the job for quicker access (users would still need to click on the "Artifacts" tab but at least they'd already be on the right build and job 🙂 )
```suggestion
      echo "Content exceeds $github_pr_comment_max_size characters. [Navigate to Buildkite build artifacts](${BUILDKITE_BUILD_URL}#${BUILDKITE_JOB_ID}) to see details." >> $COMMENT_FILE
```

AliSoftware: Btw:
 - There's a risk that `current_file_size` would be so close to `github_pr_comment_max_size` already that even just adding this "Content exceed …" line to it will make it go over the limit…
    - So the `github_pr_comment_max_size` limit value should really be 65536… minus the length of this line telling that the content exceeds the max size 😅 
 - Wouldn't the logic of `append_content` here add the `echo "Content exceeds …"` string multiple times if you call `append_content` multiple times and it's not just the last one but one before that that makes the length be exceeded?
    - Though I guess in this case it's not a big deal because you only call `append_content` a maximum number of 2 times, so this scenario can only happen if the content from `DIFF_DEPENDENCIES_FILE` is already too long, and in which case you'll print the `Content exceeds …` 2 times (one for when you call it for `DIFF_DEPENDENCIES_FILE`, then other when you call it for `DIFF_BUILD_ENV_FILE`)… so probably not that dramatic and not worth fixing I guess.
    - But just wanted to mention in case you wanted to look into it, or if you planned in future PRs to potentially add more info and call `append_content` more times in future iterations of this helper, just to keep that in mind.

AliSoftware (commented):
Looking good, just added some tweaking ideas.

bin/comment_with_dependency_diff
wzieba: ah yes, thank you. It simplified things a lot!

wzieba:
Thank you @AliSoftware for the review! You raised good points. I think I've adjusted the implementation. Here are also updated tests:
- Matticspace Mobile (case with exceeding size: https://github.com/Automattic/Matticspace-Mobile/pull/67)
- Pocket Casts (casual case: https://github.com/Automattic/pocket-casts-android/pull/2857)

Both seem to still work fine.

wzieba:
Correction, title is missing

image

AliSoftware (approved):
:shipit:

wzieba:
We're back on track!
image
https://github.com/Automattic/Matticspace-Mobile/pull/67

AliSoftware:
I just realized that there might be more to it to compute the exact comment body length, because:
 - I'm guessing GitHub counts `\n` as a character (i.e. comparing with the current length + length of header + length of body is off by one or two due to the extra newlines)
 - `comment_on_pr` also adds some hidden content (`<!-- … -->` that includes the comment ID (`--id`), so that subsequent calls (from subsequent CI builds on the same PR) can know which comment to edit / delete. So that adds a bit to the length as well

Maybe the easiest solution is to simply reduce the value of `github_comment_max_size` in your code to something like 65400, to give some margin for those extras characters and id comment?

PR #122, opened 2024-09-16
Append "tldr" to dependency diff

wzieba:
### Description

This PR adds a support for printing a "tldr" version of dependency diff: in addition to a tree, we'll print also a list in hope, that it'll make things more readable and useful in some contexts.

#### Why I include jar, instead of fetching it like `dependency-tree-diff`?

For safety reasons. The jar is placed in a place, where it has access to sensitive data, I wanted to build it from source.

I wonder: should we do the same with dependency diff tree jar? 🤔 

#### How I built it?

I built it from tag [`v0.0.6`](https://github.com/careem/dependency-diff-tldr/tree/v0.0.6) with `r8Jar` task.

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/comment_with_dependency_diff
```diff
@@ -151,7 +174,7 @@ if [ -n "$DIFF_DEPENDENCIES_FILE" ] || [ -n "$DIFF_BUILD_ENV_FILE" ]; then
   append_content() {
     local header="$1"
     local content="$2"
-    local github_comment_max_size=65536
+    local github_comment_max_size=65400 # GitHub constraint is 65536, but we decrease this value slightly to reserve for hidden comment_on_pr content and line breaks
```
wzieba, line 183: Reference: https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/120#issuecomment-2355921592

jar/dependency-diff-tldr.jar
AliSoftware, line 1: > Why I include jar, instead of fetching it like dependency-tree-diff?
> For safety reasons. The jar is placed in a place, where it has access to sensitive data, I wanted to build it from source.

I'm not sure I follow/agree with the rationale here.
 - In fact I feel like having the JAR committed could be an additional risk, because we don't check its checksum so if anyone were to make a sneaky PR to replace that JAR with a malicious one, we would not notice
 - Also, committing the JAR means that we have no idea where the source code of that JAR lives and what this binary does in details, in case we'd need to inspect its source code to ensure it doesn't do anything malicious or something.
 - Plus, if one day we want to update this `dependency-diff-tldr` tool to a newer version (1) there's no indication where the source is to build a newer version, or the place where to download pre-built binaries from a trusted source (2) we also can't know what the current version of the JAR is to know if it's already up-to-date or worth updating…

I'd be ok if you plan to commit the source code of this tool in the repo instead though, and then **not** commit the JAR itself (but compile it on-the-fly from the source), as it would resolve all by concerns above. And it seems from your PR description that it's what you initially intended to do ("I wanted to build it from source")… but not what I see from this diff, where I only see the JAR but not its source?

---

<details><summary>📖 Story time: concret example where committing a JAR in one of our other repo caused trouble</summary>


A concrete example where not having access to the source or to the repo that provided a pre-built JAR binary caused me trouble in the past is [with Tumblr Android having this pre-compiled JAR commited]( https://github.tumblr.net/Tumblr/android/blob/master/app/libs/buildSrc.jar). Not only does this `buildSrc.jar` not have a super clear file name to indicate what it is for (tip: it is **not** used to build source code), but we have no idea of knowing what invoking it actually does.

After a lot of digging and time spent spelunking, I discovered a while ago [that this JAR is responsible for handling the version numbers in Tumblr Android](https://github.com/orgs/Automattic/projects/841/views/1?filterQuery=version&pane=issue&itemId=51040340).
But we don't have access to the source of this JAR if we wanted to update how versioning is implemented by it. It's basically a black box, with a very unclear/misleading file name, and more importantly no telling how it works. 

I'd definitively want to avoid a similar situation here (even if `dependency-diff-tldr.jar` is a bit more descriptive of a filename than `buildSrc.jar` 😅  but still not having the source and implementation to be able to inspect what it does is problematic, not only security-wise, but also to know the version and what it really does if we need updating its implementation/behavior in the future.
</details>

wzieba: I see, let me explain my rationale in more details:

> In fact I feel like having the JAR committed could be an additional risk, because we don't check its checksum so if anyone were to make a sneaky PR to replace that JAR with a malicious one, we would not notice

I don't think this is an increased risk - someone can also make a sneaky PR to update checksum.

> Also, committing the JAR means that we have no idea where the source code of that JAR lives and what this binary does in details, in case we'd need to inspect its source code to ensure it doesn't do anything malicious or something.

Please see "how I built" part. I built the jar from the tag v0.0.6 from the original repository. 

After reading your comment, I've noticed a risk that someone could update v0.0.6 tag in the future, but I don't see it as any danger (more problematic debugging or things like this).

> Plus, if one day we want to update this dependency-diff-tldr tool to a newer version (1) there's no indication where the source is to build a newer version

Do you think we should fork the tldr then maybe? Would this address the concern?

> or the place where to download pre-built binaries from a trusted source (2) 

In my proposal we don't need such place, because the jar is in the repository

> we also can't know what the current version of the JAR is to know if it's already up-to-date or worth updating…

We use version `v0.0.6` - I thought documenting this in the PR is sufficient. Should I add a code comment maybe?

> and it seems from your PR description that it's what you initially intended to do ("I wanted to build it from source")… but not what I see from this diff, where I only see the JAR but not its source?

I think I haven't expressed myself clearly. Let me rephrase my sentence:

> Why I include jar, instead of fetching it like dependency-tree-diff?
For safety reasons. The jar is placed in a place, where it has access to sensitive data, I wanted to build it from source.

to 

> I include jar for safety reason. This jar has access to sensitive data, so I made a decission to build it from source (on my local machine) and include the resulting artifact to this repository. This way, I'm the "trust contat" instead of "someone on GitHub".


In other words: I don't eagerly trust the GitHub artifact from [here](https://github.com/careem/dependency-diff-tldr/releases/tag/v0.0.6) as it doesn't have to be built from the public source.

wzieba: >Not only does this buildSrc.jar not have a super clear file name to indicate what it is for (tip: it is not used to build source code), but we have no idea of knowing what invoking it actually does.

For me personaly it's quite clear, but this is only because of my experience with Gradle 😅 . I'd be guessing this artifact was built from [`buildSrc`](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) included project.

But that's just a side comment, I get your point 👍 

----

Our possible solutions:
1. Fork the "tldr" repository and build it on CI, then attach created `.jar` to release for transparency and then use it as dependency diff tree here
2. Fetch the source code of the "tldr" each time and execute Gradle build every time (more time-consuming, more build dependencies, I'm not sure)
3. Leave this as it is - add a code comment that this is locally built .jar (assuming we trust all A12s)

AliSoftware: Ah ok. I have to admit I read the "Why I included the JAR" but somehow missed the "How I built it" section 😓 

> We use version v0.0.6 - I thought documenting this in the PR is sufficient. Should I add a code comment maybe?

Yes, definitively worth adding a code comment to mention clearly (and in a place that stays with the code, instead of us having to dig into the git history in 3 months to find the last PR that updated the JAR and look at the description of that PR to find the URL of the repo where the source lives, and the tag which was last used).

The code comment should IMHO at least include the URL of the repository where the source of this JAR lives (https://github.com/careem/dependency-diff-tldr)

As for indicating the which tag was used, I'm not sure what would be the best.
 - You could also put that info in the code comment next to the URL of the source repo… but it'd make it easy to accidentally forget to update the code comment next time one rebuilds the JAR with a newer tag… rendering the information not only obsolete, but also misleading (comment would say it's still 0.0.6 while it's been updated to 0.0.7 or 0.0.8 since…)
 - Or you could include the tag name (or version, or commit sha of the original repo) in the JAR name (`dependency-diff-tldr-v0.0.6.jar`), that way if/when we decide to update it to a newer version and recompile it from source to then commit a new JAR file, the version would be updated alongside the new name we'd give to that new JAR…
 - Or maybe you have another idea?

I think putting the tag name (or commit sha) into the JAR filename is probably the easiest approach.
The most important thing is to be able to keep track what source code (repo URL and tag/commit) was used to build that JAR, so that we can track versions, debug behavior of the script, etc… if needed, inspect the source code for security vulnerabilities, know how to update it, etc.

AliSoftware: > I'd be guessing this artifact was built from [buildSrc](https://docs.gradle.org/current/userguide/organizing_gradle_projects.html#sec:build_sources) included project.

<details><summary>Funny, because that was not my guess, and in fact according to my digging, this is something completely different and the name `buildSrc` is a complete red herring</summary>

While, after reading the doc you linked, I think the name of the `buildSrc.jar` file that was committed probably comes from the fact that this mechanism exists in Gradle and thus that's maybe how the JAR was originally build then never renamed when it was extracted… but:

 - Back when I checked a while ago, I don't think there was a `buildSrc/` folder in Tumblr (where I'd expect the source code for this JAR would reside, according to the documentation you linked)
 - Looking at the repo today, I can see [a `buildSrc/` at the repo root](https://github.tumblr.net/Tumblr/android/tree/master/buildSrc) indeed now, but the source code in this dir don't match at all what the `app/libs/buildSrc.jar` binary is doing. So that's not the source that was used to build this `buildSrc.jar`, and instead that JAR was built separately, from a completely different source code, and committed to the repo.
 - Instead (and from my crude understanding of how Gradle works), I noticed that this committed `buildSrc.jar` seemed to be referenced by [this `versionBumpPlugin.gradle`](https://github.tumblr.net/Tumblr/android/blob/master/scripts/gradle/versionBumpPlugin.gradle)—which in turn was [applied here in `app/build.gradle`](https://github.tumblr.net/Tumblr/android/blob/aa9c06473c72b207cac66e34561d3126680e52e5/app/build.gradle#L22) so that [the `appVersion` object it declares](https://github.tumblr.net/Tumblr/android/blob/master/scripts/gradle/versionBumpPlugin.gradle#L17)… [could be used here to set the `versionName`](https://github.tumblr.net/Tumblr/android/blob/aa9c06473c72b207cac66e34561d3126680e52e5/app/build.gradle#L22-L29).
 - But where was the source of that `VersionExtension` that this plugin was using then? I saw that `VersionExtension` object type was imported from `com.phamousapps.gradle.VersionExtension`, which seems to come from `com.phamousapps.gradle:buildSrc` dependency… but how do you find the source code that created that dependency to look at its implementation and understand what the methods like `Version.values()` or `project.appVersion.bump(version)` actually do (is the bump implemented by using `-rc-` convention, or `x.y.z.n` convention? Is it only bumping the 2nd number or can it bump to hotfix number? etc
 - The only source file I found that looks like it could match how this code is used, I found it by doing a Google search for `com.phamousapps.gradle`. The repo that I found was [this one](https://github.com/kevinthecity/GradlePluginExample)… which clearly by its name is more meant to be used as an example on how to build a Gradle Plugin rather than a real plugin supposed to be used and maintained. But how can I be sure this `app/libs/buildSrc.jar` that is committed in the Tumblr repo was built from this repo, and not from someone having done a fork of that repo, modifying the implementation a bit, and rebuilding it… or even without officially forking, just checked out that example repo and tweaked its code to their Tumblr needs before building the JAR? And even if it's build from the source from that repo (and not some other repo that existed in the past but cannot be found with Google anymore), how do I know which commit/git tag it was built from?

In the end you can see how this `buildSrc.jar` not only had a non-descriptive name (not telling what it does), but more than that, that name was also misleading (proof: it misled you into thinking it was build from a source that would live in the magic `buildSrc/` folder that Gradle can auto-detect during builds, while it was not), it took me a lot of digging and guessing to find the potential source, and finally there's no way to really know what source code was used to build it (was this guessed repo the right one? if so, which tag was used? if not then what was used and does the source code even still exists on the net?)…
</details>

> For me personaly it's quite clear, but this is only because of my experience with Gradle 😅 . 
So you can see from my explanations above that not only was this not the case, but this also means that this `buildSrc.jar` naming was even worse / more misleading than I initially thought, and you fell into the wrong assumption that this JAR filename gave, leading you to misinformation/misunderstanding what it'd do and where it came from.

So TL;DR: really not an example to follow… on the contrary, even 😅

wzieba: Thanks for the suggestion and more context on the story. Yes, I totally agree it's not an example to follow 😅 especially for such a crucial part of a build system as in your example!

I applied a mix of suggestions here: 8ab9259

hooks/environment
```diff
@@ -12,5 +12,6 @@ PLUGIN_ROOT=$( dirname "$HOOKS_ROOT" )
 PLUGIN_BIN="$PLUGIN_ROOT/bin"
 
 export PATH="$PATH:$PLUGIN_BIN"
+export PLUGIN_ROOT=$PLUGIN_ROOT
 
 echo "~~~ :file_cabinet: Loaded Caching Plugin"
```
AliSoftware: I think this is a leftover from when this environment hook was used to load our git cache plugin (back when it was named bash-cache instead of a8c-ci-toolkit?) or something like that … or maybe it was just copied from the Buildkite plugin's own hook, hence why the env vars mention "PLUGIN" in their name?

Anyway, from the looks of the content of this hook nowadays, this seem to be used for a different usage (or is it even really needed at all, vs deriving those paths from the repo root directly or some existing env var already exposed by Buildkite, given that if I understand the logic correctly `PLUGIN_ROOT` will end up resolving to the root of the repo?), so at best this `echo` is misleading and should be updated to reflect what is done here, or the hook file is obsolete and should be removed completely…

bin/comment_with_dependency_diff
```diff
@@ -103,17 +106,34 @@ if [[ $sha != "$DEPENDENCY_TREE_CHECKSUM" ]]; then
 fi
 chmod +x dependency-tree-diff.jar
 
+JAR_DIR="$PLUGIN_ROOT/jar"
```
AliSoftware: Semantically speaking I'm not convinced that saying that the `/jar` dir is under `$PLUGIN_ROOT` makes sense—actually I don't think that having an env var named `PLUGIN_ROOT` but pointing to the root of a repo that is not a plugin makes sense as a var name in the first place…

```suggestion
REPO_ROOT=$( dirname "$( dirname "${BASH_SOURCE[0]}" )" )
JAR_DIR="$REPO_ROOT/jar"
```
Or
```suggestion
REPO_ROOT=$( git rev-parse --show-top-level )
JAR_DIR="$REPO_ROOT/jar"
```

wzieba: > (...) having an env var named PLUGIN_ROOT but pointing to the root of a repo that is not a plugin (...)

I might need a little more clarification: isn't `a8c-ci-toolkit-buildkite-plugin` a plugin?

hooks/environment
wzieba: I see, hm WDYT about creating an issue to review if `hooks/environment` is needed at all? I'd prefer to not handle this it in this PR 😅

bin/comment_with_dependency_diff
wzieba: I applied the suggestion anyway, as it doesn't require a diff in `environment`, which is neat! [4391e91](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/122/commits/4391e91c7648bb784089056eeff32c5590e96c52)

AliSoftware: > I might need a little more clarification: isn't `a8c-ci-toolkit-buildkite-plugin` a plugin?

🤦  this is what I get from being tired and thinking I was reviewing a different repo…

_(I think all those discussions around the JAR file being committed and my comparison with Tumblr-Android having a JAR etc … made me lose focus and forget that in this context we were still in the a8c-ci-toolkit plugin repo and not in an app repo like Tumblr-Android…)_

Never mind, ignore my previous feedback then

AliSoftware: > I applied the suggestion anyway, as it doesn't require a diff in `environment`, which is neat! [4391e91](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/122/commits/4391e91c7648bb784089056eeff32c5590e96c52)

Yeah actually my suggestion that would be using `git rev-parse --show-top-level` would have probably been wrong, because once the Buildkite plugin will be installed by Buildkite in the agent there's a risk that `git rev-parse --show-code top-level` will resolve in the context of the app repo being built instead of the repo when the Buildkite plugin will be checked out … so there's a risk that such a fix would actually break I think.

Sorry for the wrong interpretation on my part; now that I realize that this code indeed lives in a Buildkite plugin, I think it's totally fine if you want to revert back to export and use `PLUGIN_ROOT` after all 😓

AliSoftware (approved):

bin/comment_with_dependency_diff
wzieba: Gotcha, thank you for the help here 🙌 ! I double-checked if all works fine 
image [673c7ab](https://github.com/Automattic/pocket-casts-android/pull/2857/commits/673c7abf58ba814211e3897c558ee902b01ae231)

Merging

PR #127, opened 2024-09-19
Fix logic of github_api to autodetect host

AliSoftware:
Fixes https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/issues/125

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #124, opened 2024-09-19
Use BUILDKITE_REPO to retrieve repo owner and name

wzieba:
This way, the upload will also work on commits not associated with a PR

Test: 

- https://github.com/Automattic/Gravatar-SDK-Android/pull/320
- https://buildkite.com/automattic/gravatar-sdk-android/builds/932#01920a2a-6ffa-43b9-b753-972c1f6ac1de/181-207

---

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #128, opened 2024-09-19
Use `github_api` in sarif upload script

wzieba:
---

### DIFF
--- a/bin/upload_sarif_to_github
+++ b/bin/upload_sarif_to_github
@@ -62,4 +62,4 @@ fi
 sarif_json_temp_file=$(mktemp)
 echo "$json" > "$sarif_json_temp_file"
 
-github_api "https://api.github.com/repos/$slug/code-scanning/sarifs" --data-binary "@$sarif_json_temp_file"
+github_api "repos/$slug/code-scanning/sarifs" --data-binary "@$sarif_json_temp_file"
--- a/bin/upload_sarif_to_github
+++ b/bin/upload_sarif_to_github
@@ -62,10 +62,4 @@ fi
 sarif_json_temp_file=$(mktemp)
 echo "$json" > "$sarif_json_temp_file"
 
-curl -L \
-  -X POST \
-  -H "Accept: application/vnd.github+json" \
-  -H "Authorization: Bearer $GITHUB_TOKEN" \
-  -H "X-GitHub-Api-Version: 2022-11-28" \
-  --data-binary "@$sarif_json_temp_file" \
-  https://api.github.com/repos/"$slug"/code-scanning/sarifs
+github_api "https://api.github.com/repos/$slug/code-scanning/sarifs" --data-binary "@$sarif_json_temp_file"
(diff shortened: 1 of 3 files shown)


bin/upload_sarif_to_github
```diff
@@ -62,10 +62,4 @@ fi
...
 echo "$json" > "$sarif_json_temp_file"
 
-curl -L \
-  -X POST \
-  -H "Accept: application/vnd.github+json" \
-  -H "Authorization: Bearer $GITHUB_TOKEN" \
-  -H "X-GitHub-Api-Version: 2022-11-28" \
-  --data-binary "@$sarif_json_temp_file" \
-  https://api.github.com/repos/"$slug"/code-scanning/sarifs
+github_api "https://api.github.com/repos/$slug/code-scanning/sarifs" --data-binary "@$sarif_json_temp_file"
```
wzieba: It still requires `$slug`, but not rest of the fields.

I intentionally removed `X-GitHub-Api-Version: 2022-11-28` - I don't think it's actually necessary

WDYT @AliSoftware

AliSoftware: According to its code-comment doc and implementation, I think the first parameter of the `github_api` call is expected to only be the path (endpoint), so I'd expect this to be instead:
```suggestion
github_api "repos/$slug/code-scanning/sarifs" --data-binary "@$sarif_json_temp_file"
```

AliSoftware: I just noticed that your code used `Authorization: Bearer $GITHUB_TOKEN` while the `github_api` helper uses `Authorization: token $GITHUB_TOKEN`. I wonder if there's a difference, or if both of those are equivalent and interchangeable… 🤔

AliSoftware: [Answer to the above comment of mine](https://docs.github.com/en/rest/authentication/authenticating-to-the-rest-api?apiVersion=2022-11-28):

> Note: In most cases, you can use `Authorization: Bearer` or `Authorization: token` to pass a token. However, if you are passing a JSON web token (JWT), you must use `Authorization: Bearer`.

Maybe we should update `github_api` to use `Authorization: Bearer` while at it then, so that it is more generic and can cover all cases? 🤷

wzieba: > I think the first parameter of the github_api call is expected to only be the path (endpoint), so I'd expect this to be instead:

Yeah, of course, sorry. I moved to quickly between contexts 😅 Fixed.

> Maybe we should update github_api to use Authorization: Bearer while at it then, so that it is more generic and can cover all cases? 🤷

Yeah, sounds like a good direction! 85fa9f2

AliSoftware (approved):
You should also update the `CHANGELOG.md`; probably no need to add new entries/bullet points, just add the `[#128]` PR reference next to the 2 existing bugfix entries (in addition to them already referencing `[#124]`/`[#127]` respectively) should be enough?

Apart from that, LGTM 👍

PR #133, opened 2024-11-07
R&D - Error handling from Bash and PS1 in Buildkite/Windows

mokagio:
---

### DIFF
--- a/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
+++ b/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
@@ -1,6 +1,6 @@
 $ErrorActionPreference = "Stop"
 
-& "$PSScriptRoot\command-that-fails-without-set-e.sh"
+& "$PSScriptRoot\command-that-fails-without-set-e.sh" /c exit 1
 
 if ($LastExitCode -ne 0) {
     Write-Host "Script failed with exit code $LastExitCode"
--- a/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
+++ b/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
@@ -1,0 +1,6 @@
+& "$PSScriptRoot\command-that-fails-without-set-e.sh"
+
+if ($LastExitCode -ne 0) {
+    Write-Host "Script failed with exit code $LastExitCode"
+    exit $LastExitCode
+}
--- a/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
+++ b/.buildkite/commands/command-that-calls-bash-without-set-e.ps1
@@ -1,3 +1,5 @@
+$ErrorActionPreference = "Stop"
+
 & "$PSScriptRoot\command-that-fails-without-set-e.sh"
 
 if ($LastExitCode -ne 0) {
--- a/.buildkite/commands/command-that-calls-cmd-not-installed.ps1
+++ b/.buildkite/commands/command-that-calls-cmd-not-installed.ps1
@@ -1,6 +1,8 @@
 $ErrorActionPreference = "Stop"
 
-false
+# This should fail because npm is not available out of the box on our Windows
+# CI in 06/2025
+npm
 
 if ($LASTEXITCODE -ne 0) {
     Write-Host "Script failed with exit code $LASTEXITCODE"
--- a/.buildkite/commands/command-that-calls-other-ps1-failing.ps1
+++ b/.buildkite/commands/command-that-calls-other-ps1-failing.ps1
@@ -5,6 +5,10 @@
 & "$PSScriptRoot\command-that-fails-internally.ps1"
 
 if ($LASTEXITCODE -ne 0) {
-    Write-Error "Script failed with exit code $LASTEXITCODE"
+    # Using Write-Error doesn't print the exit code...
+    # Write-Error "Script failed with exit code $LASTEXITCODE"
+    #
+    # ...will using Write-Host be better?
+    Write-Host "Script failed with exit code $LASTEXITCODE"
     exit $LASTEXITCODE
 }
--- a/.buildkite/commands/command-that-calls-other-ps1-failing.ps1
+++ b/.buildkite/commands/command-that-calls-other-ps1-failing.ps1
@@ -1,1 +1,10 @@
+# Alone, this does not fail
+# & "$PSScriptRoot\command-that-fails-internally.ps1"
+
+# To make the script fail, we need to check the error
 & "$PSScriptRoot\command-that-fails-internally.ps1"
+
+if ($LASTEXITCODE -ne 0) {
+    Write-Error "Script failed with exit code $LASTEXITCODE"
+    exit $LASTEXITCODE
+}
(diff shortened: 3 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,33 @@ env:
   IMAGE_ID: xcode-15.4
 
 steps:
+  - label: Windows test inline Bash
+    command: |
+      echo "I'm about to exit 2. I expect this step to fail..."
+      exit 2
+    agents:
+      queue: windows
```
mokagio: Fails as expected in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1217

Meaning: **Calling `exit [>0]` from the Bulidkite `command` step results in the step failing**  no surprise there...

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,33 @@ env:
...
+
+  - label: Windows test Bash script no set -e
+    command: .buildkite/commands/command-that-fails-without-set-e.sh
+    agents:
+      queue: windows
+
+  - label: Windows test PS1 script that fails
+    command: .buildkite/commands/command-that-fails-internally.ps1
+    agents:
+      queue: windows
```
mokagio: Fails as expected in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1217

Meaning: **Calling `Exit [>0]` from a PS1 script result in the CI step failing.**

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,33 @@ env:
...
+
+  - label: Windows test PS1 script that call script that fails
+    command: .buildkite/commands/command-that-calls-other-ps1-failing.ps1
+    agents:
+      queue: windows
```
mokagio: Succeeds in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1217

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,33 @@ env:
...
+
+  - label: Windows test Bash script
+    command: .buildkite/commands/command-that-fails.sh
+    agents:
+      queue: windows
+
+  - label: Windows test Bash script no set -e
+    command: .buildkite/commands/command-that-fails-without-set-e.sh
+    agents:
+      queue: windows
```
mokagio: Succeeds in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1217

That's not too surprising, though, because there is no `set -e`, right?

.buildkite/pipeline.yml
```diff
@@ -8,6 +8,33 @@ env:
...
+    command: |
+      echo "I'm about to exit 2. I expect this step to fail..."
+      exit 2
+    agents:
+      queue: windows
+
+  - label: Windows test Bash script
+    command: .buildkite/commands/command-that-fails.sh
+    agents:
+      queue: windows
```
mokagio: Succeeds in https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1217

Might need additional config at call site to pass the failure along?

.buildkite/commands/command-that-fails-without-set-e.sh
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
+
+echo "I'm about to exit 2. I expect this step to fail..."
```
mokagio:
```suggestion
echo "I'm about to exit 2. I expect this step to pass if called directly, because the script does not set the -e flag."
```

.buildkite/commands/command-that-calls-cmd-not-installed.ps1
```diff
@@ -0,0 +1,10 @@
+$ErrorActionPreference = "Stop"
+
+# This should fail because npm is not available out of the box on our Windows
+# CI in 06/2025
+npm
+
+if ($LASTEXITCODE -ne 0) {
+    Write-Host "Script failed with exit code $LASTEXITCODE"
+    exit $LASTEXITCODE
+}
```
mokagio, line 10: Unless we set `$ErrorActionPreference = "Stop"`, this won't result in a failure. See [this build](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1229/steps/canvas?sid=0197a3c3-6d66-42c8-931c-b37bbad926c9)

image

My guess is that's because a missing command is not considered a terminating error by default 🤷‍♂️ 

Also notice that, once the non-terminating error behavior is set, the script fails _before_ the `$LastExitCode` as shown in the screenshot below where there is no "Script failed with..." log.

image

So, in this particular case, the `$LastExitCode` check is redundant. Of course in the real world one would assume that `npm` is available, and therefore should follow up calling it with the check.

.buildkite/commands/command-that-calls-other-ps1-failing.ps1
```diff
@@ -0,0 +1,14 @@
...
+& "$PSScriptRoot\command-that-fails-internally.ps1"
+
+if ($LASTEXITCODE -ne 0) {
+    # Using Write-Error doesn't print the exit code...
+    # Write-Error "Script failed with exit code $LASTEXITCODE"
+    #
+    # ...will using Write-Host be better?
+    Write-Host "Script failed with exit code $LASTEXITCODE"
+    exit $LASTEXITCODE
+}
```
mokagio, line 14: This is the experiment from which I derived the recommendation to **always follow an external command or script invocation with an exit code check.**

The check is verbose here because of the investigation, but we can do it in compact form like:

```ps1
if ($LastExitCode -ne 0) { exit $LastExitCode }
```

Also notice that by default PS1 is case-insenstive: So `$LASTEXITCODE`, `$LastExitCode`, and `$lastexitcode` are equivalent. My understanding is that `$LastExitCode` is the preferred form.

.buildkite/commands/command-that-calls-bash-without-set-e.ps1
```diff
@@ -0,0 +1,8 @@
+$ErrorActionPreference = "Stop"
+
+& "$PSScriptRoot\command-that-fails-without-set-e.sh" /c exit 1
+
+if ($LastExitCode -ne 0) {
+    Write-Host "Script failed with exit code $LastExitCode"
+    exit $LastExitCode
+}
```
mokagio, line 8: The script fails internally but a caller wouldn't notice it because there is no `set -e`.

```sh
#!/bin/bash

echo "I'm about to exit 2. I expect this step to pass if called directly, because the script does not set the -e flag."
exit 2
```

Because of that, Windows won't consider the result a failure:

image

The `/c exit 1` is a ChatGPT suggestion, but evidently it does not affect the false positive outcome.

mokagio:
Closing because the research is done.

See paaHJt-8BQ-p2

PR #141, opened 2025-01-21
Optionaly fail install_swiftpm_dependencies when `Package.resolved` is not current

andrewdmontgomery:
Add an optional flag (`--strict-package-check`) that will cause the plugin to exit non-zero if the `Package.resolved` is not current.

Update the `print_usage()` function output to handle the new optional flag better.

Since the argument parsing requires looping, add logic to prevent multiple target types from being passed.
---

### DIFF
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -22,6 +22,7 @@ Options:
   --project PATH            Specify the path to the Xcode project.
   --use-spm                 Use the Swift package manager at the root.
   --strict-package-check    Fail the installation if the \`Package.resolved\` file has uncommitted changes or is out of date.
+                            If specified more than once, the last value will be used.
 
 Examples:
   Install dependencies for an Xcode workspace:
@@ -41,18 +42,36 @@ XCODEPROJ_PATH=
 USE_SPM=false
 STRICT_PACKAGE_CHECK=false
 
+# Track the number of targets found in the arguments
+TARGET_COUNT=0
+
 while [[ $# -gt 0 ]]; do
   case ${1:-} in
     --workspace)
+      if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]; then
+        echo "Error: only one target can be used: --workspace, --project, or --use-spm"
+        exit 1
+      fi
       XCWORKSPACE_PATH="${2:?you need to provide a value after the --workspace option}"
+      TARGET_COUNT=$((TARGET_COUNT + 1))
       shift; shift
       ;;
     --project)
+      if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]; then
+        echo "Error: only one target can be used: --workspace, --project, or --use-spm"
+        exit 1
+      fi
       XCODEPROJ_PATH=${2:?you need to provide a value after the --project option}
+      TARGET_COUNT=$((TARGET_COUNT + 1))
       shift; shift
       ;;
     --use-spm)
+      if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]; then
+        echo "Error: only one target can be used: --workspace, --project, or --use-spm"
+        exit 1
+      fi
       USE_SPM=true
+      TARGET_COUNT=$((TARGET_COUNT + 1))
       shift
       ;;
     --strict-package-check)
@@ -65,6 +84,12 @@ while [[ $# -gt 0 ]]; do
       ;;
   esac
 done
+
+# Ensure one and only one mutually exclusive option is provided
+if [[ $TARGET_COUNT -eq 0 ]]; then
+  echo "Error: You must provide one of --workspace, --project, or --use-spm."
+  exit 1
+fi
 
 if [[ "$#" -gt 0 ]]; then
   echo "Unexpected extra arguments: $*" >&2
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -85,12 +85,6 @@ while [[ $# -gt 0 ]]; do
   esac
 done
 
-# Ensure one and only one mutually exclusive option is provided
-if [[ $TARGET_COUNT -eq 0 ]]; then
-  echo "Error: You must provide one of --workspace, --project, or --use-spm."
-  exit 1
-fi
-
 if [[ "$#" -gt 0 ]]; then
   echo "Unexpected extra arguments: $*" >&2
   print_usage
--- a/bin/install_swiftpm_dependencies
+++ b/bin/install_swiftpm_dependencies
@@ -14,25 +14,32 @@ print_usage() {
 XCWORKSPACE_PATH=
 XCODEPROJ_PATH=
 USE_SPM=false
+STRICT_PACKAGE_CHECK=false
 
-case ${1:-} in
-  --workspace)
-    XCWORKSPACE_PATH="${2:?you need to provide a value after the --workspace option}"
-    shift; shift
-    ;;
-  --project)
-    XCODEPROJ_PATH=${2:?you need to provide a value after the --project option}
-    shift; shift
-    ;;
-  --use-spm)
-    USE_SPM=true
-    shift
-    ;;
-  -h|--help)
-    print_usage
-    exit 0
-    ;;
-esac
+while [[ $# -gt 0 ]]; do
+  case ${1:-} in
+    --workspace)
+      XCWORKSPACE_PATH="${2:?you need to provide a value after the --workspace option}"
+      shift; shift
+      ;;
+    --project)
+      XCODEPROJ_PATH=${2:?you need to provide a value after the --project option}
+      shift; shift
+      ;;
+    --use-spm)
+      USE_SPM=true
+      shift
+      ;;
+    --strict-package-check)
+      STRICT_PACKAGE_CHECK="true"
+      shift
+      ;;
+    -h|--help)
+      print_usage
+      exit 0
+      ;;
+  esac
+done
 
 if [[ "$#" -gt 0 ]]; then
   echo "Unexpected extra arguments: $*" >&2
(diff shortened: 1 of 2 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

andrewdmontgomery:
@AliSoftware, do you think this is an appropriate place for this functionality?  Would we rather have a dedicated plugin that fails if certain paths have uncommitted changes?  I think this "uncommitted changes" check should target specific paths, since there could be other uncommitted changes that shouldn't cause this to fail. So this seems like a good place to enforce this.

I love the tests.  Adding this optional flag will complicate the testing.  If we think this is a good way to solve this need, I can:
1. create duplicate of `tests/install_swiftpm_dependencies/package` with an out-of-date `Package.Resolved`
2. same with `project` and `workspace`
3. create new `test_*` scripts for each scenario
4. update the pipeline accordingly

bin/install_swiftpm_dependencies
```diff
@@ -109,16 +160,27 @@ if [[ -d "${XCWORKSPACE_PATH}" ]]; then
   # (despite the help page of `xcodebuild` suggesting that it should work without `-scheme`). Since the dependency resolution doesn't really depend on the scheme
   # and we don't want to have to provide or guess it, using `-list` instead stops making `xcodebuild` complain about `-workspace` not being used in conjunction
   # with `-scheme` (even if in practice we don't care about the scheme list it returns)
-  xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list
+  xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -skipPackageUpdates -list
```
andrewdmontgomery, line 163: I'm not entirely sure that this functions equivalently for existing use cases.  Here's the problem:

1. Remove dependency from `Package.swift`
2. Do not update `Package.resolved`

If `-onlyUsePackageVersionsFromResolvedFile` is used, the `Package.resolved` does not change.  That's ok for existing use cases for this toolkit plugin.  But for our use case, we need to know that the `Package.resolved` is out of date.

The descriptions of these flags in `xcodebuild -h` are not specific enough to help.

andrewdmontgomery: Changing this flag for this case seems dicey to me.  And now that I know about the [Dangerfile plugin to check this](https://github.com/Automattic/dangermattic?tab=readme-ov-file#example-of-available-plugins-and-their-usage) (thank you), this PR seems like a risky place to handle this.

bin/install_swiftpm_dependencies
```diff
@@ -1,38 +1,89 @@
...
+  1. For an Xcode \`.xcworkspace\`:
+       ${0##*/} --workspace PATH [--strict-package-check]
+
+  2. For an Xcode \`.xcodeproj\`:
+       ${0##*/} --project PATH [--strict-package-check]
+
+  3. For a Swift package managed by \`swift package\` instead of Xcode:
+       ${0##*/} --use-spm [--strict-package-check]
+
+  4. Default behavior (Attept to find an appropriate workspace, project, or swift package):
```
AliSoftware, line 17:
```suggestion
  4. Default behavior (Attempt to find an appropriate workspace, project, or swift package automatically):
```

bin/install_swiftpm_dependencies
```diff
@@ -1,38 +1,89 @@
...
-    ;;
-  -h|--help)
-    print_usage
-    exit 0
-    ;;
-esac
+STRICT_PACKAGE_CHECK=false
+
+# Track the number of targets found in the arguments
+TARGET_COUNT=0
```
AliSoftware, line 46: Seems this variable is now completely unused (I'm guessing it was initially introduced as an attempt to detect argument conflicts but then you switched to test for `if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]` in each branch since?

bin/install_swiftpm_dependencies
```diff
@@ -1,38 +1,89 @@
...
+
+while [[ $# -gt 0 ]]; do
+  case ${1:-} in
+    --workspace)
+      if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]; then
+        echo "Error: only one target can be used: --workspace, --project, or --use-spm"
+        exit 1
+      fi
```
AliSoftware, line 54: I wonder if we couldn't DRY this check in a dedicated local function, e.g.
```bash
function check_options_conflicts() {
  if [[ -n $XCWORKSPACE_PATH || -n $XCODEPROJ_PATH || $USE_SPM == "true" ]]; then
    echo "Error: only one mode can be used: --workspace, --project, or --use-spm"
    exit 1
  fi
}
```
Then you'd simply call `check_options_conflicts()` in each of the `case` instead of repeating those lines.

I also changed the wording from "target" to "mode" as I felt that the word "target" was confusing (I initially thought you meant it as in "Xcode targets" and `xcodebuild`'s `--target` parameter)

---

That being said. since we'll end up removing the `--strict-package-check` option your PR initially added here (see other comments), this means we'll go back to not needing the `while` loop to wrap the `case ${1:-}` block… and thus this additional check for conflicting/duplicate options will not be necessary in the first place. And we already check for unexpected extra arguments on line L37 / R88 after the `case` block, so once you revert those changes and get rid of the added `while`, the script should still already detect unexpected (or duplicate) arguments.

bin/install_swiftpm_dependencies
AliSoftware: The use of `-skipPackageUpdates` is not the same as `-onlyUsePackageVersionsFromResolvedFile` (which is an equivalent to the legacy `-disableAutomaticPackageResolution` FYI).

According to the `xcodebuild -help` page:
```
  -resolvePackageDependencies                              resolves any Swift package dependencies referenced by the project or workspace
  -disableAutomaticPackageResolution                       prevents packages from automatically being resolved to versions other than those recorded in the `Package.resolved` file
  -onlyUsePackageVersionsFromResolvedFile                  prevents packages from automatically being resolved to versions other than those recorded in the `Package.resolved` file
  -skipPackageUpdates                                      Skip updating package dependencies from their remote
```

One of the important aspect of this helper running on our CI builds is that we don't want CI to risk updating any of the local packages used by the build, and instead want to ensure that it will only install the exact same set of packages, with the exact version being locked in the `.resolved` file, on CI every time, for reproducibility and idempotence of CI builds, which is a pretty important attribute to have on our CI.

In a way, `xcodebuild -onlyUsePackageVersionsFromResolvedFile` is to `pod install` what `xcodebuild -resolvePackageDependencies` is to `pod update` (and `Package.resolved` is SPM's pendant to `Podfile.lock`).
The former ensures that we install just exactly the versions that are listed in the lockfile and guarantee that we won't update any dependency while doing so but instead install the exact same set of dependencies on all other devs and machines building that commit. After all that's the whole point of having a lockfile, it's to lock in place the versions of the dependencies that should be installed, regardless on which machines they are installed on and when they are being installed (i.e. even if new versions of some deps have been released since you first ran the command). The latter is quite the opposite, as it allows you to disregard the versions that have been locked in the lockfile (`Package.resolved`) and allow to resolve to newer versions of those dependencies if any are available. Which is great for keeping things up-to-date with all the latest bugfixes and all, but is contrary to the idempotence and reproducibility key attributes important to us in a CI environment.

---

So TL;DR: I'd advice keeping `-onlyUsePackageVersionsFromResolvedFile` instead of switching to use `-skipPackageUpdates` which might still risk updating the dependencies under you.

Using `-onlyUsePackageVersionsFromResolvedFile` also implies that `Package.resolved` should not end up being modified by that command, and so your `git status --porcelain` check below would not make sense in such CI environment.

bin/install_swiftpm_dependencies
```diff
@@ -109,16 +160,27 @@ if [[ -d "${XCWORKSPACE_PATH}" ]]; then
   # (despite the help page of `xcodebuild` suggesting that it should work without `-scheme`). Since the dependency resolution doesn't really depend on the scheme
   # and we don't want to have to provide or guess it, using `-list` instead stops making `xcodebuild` complain about `-workspace` not being used in conjunction
   # with `-scheme` (even if in practice we don't care about the scheme list it returns)
-  xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile -list
+  xcodebuild -workspace "${XCWORKSPACE_PATH}" -resolvePackageDependencies -skipPackageUpdates -list
 elif [[ -d "${XCODEPROJ_PATH}" ]]; then
   echo "~~~ Resolving Swift Packages with \`xcodebuild\`"
   echo "Using -project \"${XCODEPROJ_PATH}\""
-  xcodebuild -project "${XCODEPROJ_PATH}" -resolvePackageDependencies -onlyUsePackageVersionsFromResolvedFile
+  xcodebuild -project "${XCODEPROJ_PATH}" -resolvePackageDependencies -skipPackageUpdates
```
AliSoftware, line 167: As discussed above, this change should be reverted

bin/install_swiftpm_dependencies
```diff
@@ -109,16 +160,27 @@ if [[ -d "${XCWORKSPACE_PATH}" ]]; then
...
+if [[ "$STRICT_PACKAGE_CHECK" == "true" ]]; then
+  CHANGES=$(git status --porcelain -- "$PACKAGE_RESOLVED_LOCATION")
+
+  if [[ -n "$CHANGES" ]]; then
+    echo "Uncommitted changes detected in $PACKAGE_RESOLVED_LOCATION:"
+    echo "$CHANGES"
+    exit 1
+  fi
+fi
+
```
AliSoftware, line 183: Given the discussion above that we want idempotence on CI, using `-onlyUsePackageVersionsFromResolvedFile` already guarantees that the `Package.resolved` will not be modified by CI, so this new code won't work in this context as we'll keep using `-onlyUsePackageVersionsFromResolvedFile`, and should thus be deleted.

AliSoftware (commented):
As discussed in [this other PR](https://github.com/Automattic/Gravatar-SDK-iOS/pull/643#issuecomment-2608045528) the addition of `--strict-package-check` to this helper might not be that necessary after all given that the Danger plugin should already prevent us from forgetting about updating the `Package.resolved`

Additionally, as discussed [in the comment above](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/141/files#r1925492356), we want to ensure CI idempotence of builds done on the same commit, so we want to use `-onlyUsePackageVersionsFromResolvedFile` to guarantee `Package.resolved` would never risk being modified by CI itself during a build.

---

This means that this PR is mostly obsolete after all, at least wrt to its original purpose of adding the optional check about `Package.resolved` not being current.

That being said, I like the improvements you provided to the `print_usage()`—reads much more clearly 👍 —so maybe we should still keep that part from this PR, even if you revert all the rest? Repurposing this PR into one that just improves the documentation of the helper instead.

Issue #130, opened 2024-10-25
Add documentation header to all our `bin/` utilities + auto-generate overall doc


AliSoftware:
Internal ref: p1729815426023749/1729802878.523709-slack-CC7L49W13

To better document all of the utilities we make available via this plugin, we should adopt the convention of adding some documentation lines at the start of each script.
 - [`comment_on_pr` is an example of a good citizen](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/comment_on_pr#L3-L22) that could be use as inspiration / starting point
 - I have recently added `.cursorrules` via #154 that can help writing and enforcing those documentation.

As a bonus, we could also have a script that take all the `bin/*` scripts in the plugin, extracts the doc header, and generates a `README` from it
 - We could then maybe have a Danger check that runs the script on CI and warns if it doesn't match the current `README` checked out in the repo, suggesting the PR author to run the script locally themselves and commit the changes
 - (we could also consider to have the CI push the generated README itself instead of asking the PR author to push it, but in general I'm not super keen to have CI push commits if we can avoid it)

PR #144, opened 2025-02-04
Windows: Install Win 10 SDK, clearer name, no Node.js setup

mokagio:
Tested via https://github.com/Automattic/studio/pull/882 and https://github.com/beeper/beeper-desktop-new/pull/1970

- After realizing that `nvm-buildkite-plugin` does a better job at setting up Node.js and the custom logic we had in the `prepare_windows_host_for_node.ps1` script, I remove said logic from the script
- Also renamed the script to `prepare_windows_host_for_app_distribution.ps1` which I find better explains the script purpose. Besides, now that it no longer sets up Node.js, the name would have been inappropriate.
- New command to install Windows 10 SDK, reading from the `.windows-10-sdk-version` file in the repo — Extracted [from Studio](https://github.com/Automattic/studio/blob/5d50385707d545a22c3b5427c28b3d46879528de/.buildkite/commands/install-windows-10-sdk.ps1) and currently duplicated in other apps
- `prepare_windows_host_for_app_distribution.ps1` automatically runs the Windows 10 install script if needed, so clients don't have to call two commands

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -34,11 +34,14 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- `prepare_windows_host_for_node.ps1` no longer install Node.js.
+  Clients should use [`nvm-buildkite-plugin`](https://github.com/Automattic/nvm-buildkite-plugin) instead [#144]
```
mokagio: Markdown renders know how to parse this syntax as a single line:

image

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -0,0 +1,99 @@
...
+# See https://github.com/hermit99/electron-windows-store/tree/v2.1.2?tab=readme-ov-file#usage
+
+$windowsSDKVersionFile = ".windows-10-sdk-version"
+if (Test-Path $windowsSDKVersionFile) {
+    Write-Host "Found $windowsSDKVersionFile file, installing Windows 10 SDK..."
+    & "$PSScriptRoot\install_windows_10_sdk.ps1"
+    If ($LastExitCode -ne 0) { Exit $LastExitCode }
+} else {
+    Write-Host "No $windowsSDKVersionFile file found, skipping Windows 10 SDK installation."
+}
```
mokagio: Example of this finding the SDK file 

image

Example of this not finding the SDK file

image

mokagio: One thing that I'm unsure of is whether we should provide a way to bypass or modify this behavior. 

Example:

- Something might be broken in the plugin and a client might need to install the SDK manually; or
- For whichever reason, it's not possible to use the default file name for the SDK version

I don't think we need to stress much about implementing this flexibility from the get go, so if you agree I'd leave it as a followup, but the code feels incomplete or suboptimal without it.

bin/prepare_windows_host_for_node.ps1
```diff
@@ -1,134 +1,10 @@
...
-    }
-    Write-Host "Your roles are:"
-    $roles | ForEach-Object { Write-Host "  - $_" }
-}
+$newScript = "prepare_windows_host_for_app_distribution.ps1"
 
-Write-Host "--- :windows: Setting up Windows for Node and Electron builds"
+Write-Host "+++ :warning: This command is deprecated"
+Write-Host "Please use $newScript instead"
+Write-Host "Now calling $newScript..."
```
mokagio: One could argue that given we already have a breaking change – no longer setting up Node.js from the PS1 command – we might as well delete this script altogether instead of marking it as deprecated. What do you think?

But I liked how it looked when it run 😢 

image

bin/path_aware_refreshenv.ps1
```diff
@@ -0,0 +1,20 @@
...
+#
+# To avoid the issue, we save the PATH pre-refreshenv and then manually add all
+# the components that were removed.
+
+Write-Host "PATH before refreshenv is $env:PATH"
+$originalPath = "$env:PATH"
+Write-Host "Calling refreshenv..."
+refreshenv
+$mergedPath = "$env:PATH;$originalPath" -split ";" | Select-Object -Unique -Skip 1
+$env:PATH = ($mergedPath -join ";")
```
AliSoftware, line 24: Nice 👍

bin/prepare_windows_host_for_node.ps1
AliSoftware: Yeah I'd argue to remove it immediately, otherwise just marking it as deprecated means that:

1. It is very likely that we'll forget to update client repos when we update the `a8c-ci-toolkit` version they use, not see the warning in the logs and let it go unnoticed
2. Removing that deprecated script later will require yet another breaking change / major bump of the plugin
3. Versions are cheap (just a tag), and we better keep our tooling clean to avoid legacy to accumulate.

If the breaking change of the new `a8c-ci-toolkit` plugin was unrelated to that and instead required changes to client repos in completely unrelated areas, I could see the benefit of keeping the old script around as deprecated, as maybe clients would want to update to the new version of the plugin for that other thing but would not be ready to also adop the new windows tooling. But given the only reason why one would want to bump the toolkit version to this new major version you're gonna introduce would be to get the updated windows tooling, and thus would use the occasion to migrate the client setup to call the new script directly instead, I say it's not work keeping legacy around.

bin/prepare_windows_host_for_app_distribution.ps1
AliSoftware: Nice!

I wouldn't mind adding an escape hatch, that might be a good idea indeed (though one could argue that'd be YAGNI…)

If it's easy enough to implement and doesn't complicate the script logic too much (which I wouldn't expect to), I'd say go for it.

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -0,0 +1,99 @@
+# Stop script execution when a non-terminating error occurs
```
AliSoftware, line 20: In anticipation for https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/issues/130, what do you think about adding some header doc to this script, as well as at the top of your other `.ps1` scripts in this PRs?

 - Describe what the script does (bullet points list?)
 - Provide usage example (especially if it accepts parameters)
 - Document customizable behavior (like the env var to bypass auto-Win10-installation you discussed below, if you end up going with it)

```suggestion
# Prepares a `windows` CI agent with all the necessary setup so it can build and distribute a windows app
#
#  - Enables long path behavior
#  - Disable Windows Defender on the CI agent
#  - Install the "Chocolatey" package manager
#  - Enable dev mode so the agent can support Linux-style symlinks
#  - Download Code Signing Certificates(1)
#  - Install the Windows 10 SDK if it detected a `.windows-10-sdk-version` file(2)
#
# (1) The certificate it installs is stored in our AWS SecretsManager storage (`windows-code-signing-certificate` secret ID)
# (2) You can skip the Win10 install even if `.windows-10-sdk-version` file is present by using the `SKIP_WINDOWS_10_SDK_INSTALL=1` env var before calling this script
#
# Note: In addition to calling this script, and depending on your client app, you might want to also install `npm` and the `Node.js` packages used by your client app on the agent too. For that part, you should use the `automattic/nvm` Buildkite plugin on the pipeline step's `plugins:` attribute.
#

# Stop script execution when a non-terminating error occurs
```

bin/install_windows_10_sdk.ps1
```diff
@@ -0,0 +1,59 @@
...
+If (-not (Test-Path $sdkPath)) {
+  Write-Output "[!] Failed to install Windows 10 SDK: Could not find SDK at $sdkPath."
+  If (-not (Test-Path $windowsSDKsRoot)) {
+    Write-Output "[!] Expected $windowsSDKsRoot to exist, but it does not."
+  } else {
+    Write-Output "    Found:"
+    Get-ChildItem -Path $windowsSDKsRoot | ForEach-Object { Write-Output "    - $windowsSDKsRoot\$_" }
+  }
+  Exit 1
+}
```
AliSoftware, line 99: Is there an easy way to check if `$windows10SDKVersion` is valid (i.e. an existing version, as opposed to if someone tried to use `latest` or `none` or whatever invalid text in the version file) and detect it early?

In particular, I'm wondering if the `vs_buildtools.exe` would fail with an explicit and clear error if that `-add …` argument we pass to it was invalid (i.e. if it didn't find the component), allowing us to print a specific error message in that case (like "Check the value you set in `$windowsSDKVersionFile`"), as opposed to the installer happily continuing with the installation of all the other default components—without complaining about this one not existing in the options—and us only finding out the component was not installed **_after the fact_**, thanks to your test on lines 42–51… (and even in that case, would it help to provide in the error messages a more explicit suggestion to check the syntax/version used in the version file?)

bin/install_windows_10_sdk.ps1
```diff
@@ -0,0 +1,59 @@
...
+
+Write-Host "--- :windows: Installing Windows 10 SDK and Visual Studio Build Tools"
+
+$windowsSDKVersionFile = ".windows-10-sdk-version"
+if (-not (Test-Path $windowsSDKVersionFile)) {
+  Write-Output "[!] No Windows 10 SDK version file found at $windowsSDKVersionFile."
+  exit 1
+}
+
+$windows10SDKVersion = Get-Content $windowsSDKVersionFile
```
AliSoftware: Question: how does this `Get-Content` handles possible newlines?

Especially, if the version file ends with a newline at end of file, is that trimmed?
Or maybe it's not, but that doesn't cause any issues because when you then reference `$windows10SDKVersion` in other commands like `--add Microsoft.VisualStudio.Component.Windows10SDK.$windows10SDKVersion` that's when the value is trimmed as usage/substitution site?
What if the file contains 2 newlines at end of file? 🤔

bin/prepare_windows_host_for_node.ps1
mokagio: Thanks for the input. Got rid of it in 2656bfa

bin/prepare_windows_host_for_app_distribution.ps1
mokagio: Sure! Thanks for providing the header doc already. 😄

bin/install_windows_10_sdk.ps1
mokagio: Great question! We clearly need to add something to strip them:

image

https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1007#01951c5c-320d-4346-8d45-3e9bda56ad1e

AliSoftware: I think the `-TotalCount 1` parameter of `Get-Content` can help?
Though even if that allows to limit only reading the first line of the file, not 100% sure if it'd properly trim the `\n` at the end of that read line, in that case you might need to also use `Trim()` or `TrimEnd()`.

Tentative suggestion (not tested):
```suggestion
$windows10SDKVersion = (Get-Content -TotalCount 1 $windowsSDKVersionFile).Trim()
```

mokagio: Spent more time than I care to admit on this but got a bunch of tests for the version parsing

image

I'll clean up tomorrow https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153

CHANGELOG.md
```diff
@@ -34,11 +34,15 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- `prepare_windows_host_for_node.ps1` has been to `prepare_windows_host_for_app_distribution` [#144]
```
iangmaia:
```suggestion
- `prepare_windows_host_for_node.ps1` has been renamed to `prepare_windows_host_for_app_distribution.ps1` [#144]
```

CHANGELOG.md
```diff
@@ -34,11 +34,15 @@ _None._
...
+- `prepare_windows_host_for_app_distribution.ps1` no longer install Node.js.
```
iangmaia:
```suggestion
- `prepare_windows_host_for_app_distribution.ps1` no longer installs Node.js.
```

CHANGELOG.md
```diff
@@ -34,11 +34,15 @@ _None._
...
+  Clients should use [`nvm-buildkite-plugin`](https://github.com/Automattic/nvm-buildkite-plugin) instead [#144]
 
 ### New Features
 
-_None._
+- Added new command to install Windows 10 SDK on Windows CI machines, `install_windows_10_sdk.ps1` [#144]
+- `prepare_windows_host_for_app_distribution` automatically installs the Windows 10 SDK if version file is found [#144]
```
iangmaia:
```suggestion
- `prepare_windows_host_for_app_distribution.ps1` automatically installs the Windows 10 SDK if version file is found [#144]
```

CHANGELOG.md
```diff
@@ -34,11 +34,15 @@ _None._
...
+- Added new command to run `refreshenv` on Windows preserving the `PATH`, `path_aware_refreshenv` [#144]
```
iangmaia:
```suggestion
- Added new command to run `refreshenv` on Windows preserving the `PATH`, `path_aware_refreshenv.ps1` [#144]
```

bin/install_windows_10_sdk.ps1
```diff
@@ -0,0 +1,59 @@
+# Stop script execution when a non-terminating error occurs
+$ErrorActionPreference = "Stop"
+
+Write-Host "--- :windows: Installing Windows 10 SDK and Visual Studio Build Tools"
+
+$windowsSDKVersionFile = ".windows-10-sdk-version"
```
iangmaia, line 32: Do you think it would be worth it to add an example / expected format for `.windows-10-sdk-version`?

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -1,19 +1,24 @@
...
+#
+#  - Enables long path behavior
+#  - Disable Windows Defender on the CI agent
+#  - Install the "Chocolatey" package manager
+#  - Enable dev mode so the agent can support Linux-style symlinks
+#  - Download Code Signing Certificates(1)
+#  - Install the Windows 10 SDK if it detected a `.windows-10-sdk-version` file(2)
+#
+# (1) The certificate it installs is stored in our AWS SecretsManager storage (`windows-code-signing-certificate` secret ID)
+# (2) You can skip the Win10 install even if `.windows-10-sdk-version` file is present by using the `SKIP_WINDOWS_10_SDK_INSTALL=1` env var before calling this script
```
iangmaia: > SKIP_WINDOWS_10_SDK_INSTALL

I'm probably missing something, but where is this used? Is it still TODO or maybe implemented in the agent? 🤔

bin/install_windows_10_sdk.ps1
AliSoftware: +1

bin/prepare_windows_host_for_app_distribution.ps1
AliSoftware: This is from me providing this documentation as a GitHub `suggestion` block in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/144#discussion_r1959643030, in which I already included the wording for that `SKIP_WINDOWS_10_SDK_INSTALL` in my documentation suggestion in anticipation that @mokagio would at some point implement it in this PR as a follow-up to https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/144#discussion_r1957636032

As of now he has yet to implement this feature discussed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/144#discussion_r1957636032 indeed, which should not be forgotten to do before merging that PR (or alternatively if we change our minds and decide it's not worth it to implement or at least not yet, we should remove the mention of it in the documentation indeed)

iangmaia: Ahh got it, I hadn't seen it! 🤦‍♂️ All clear.

AliSoftware: Btw @mokagio now that I've [TIL'd from your testing PR here](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153/files#diff-0c2ff45d12af18650ff59434e9e92c9e77243558e77f4f209ceeaf3afd8ccfa4R1-R3) that one can use `param ( … )` at the top of `.ps1` files to provide custom params, I am now thinking that such an approach would probably be better than expecting us to set an external env var for toggling this behavior like I initially suggested in my documentation snippet 🤔 💭 

e.g.
```powershell
param ( # parameters need to be defined at the top of the script
  [switch]$SkipWin10SDKInstall = $false
)
```
(then adjust the documentation + implementation to handle this param accordingly of course)

bin/install_windows_10_sdk.ps1
mokagio: >  I'm wondering if the `vs_buildtools.exe` would fail with an explicit and clear error

I've done a few experiments in my VM. It fails silently, with exit code `1`.

mokagio: I'll duplicated what I wrote in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153#discussion_r1962609092

> This all makes me wonder whether we might be better off using a default SDK version in the installer, and only falling back to the version file and/or a parameter if the user provides one. After all, I implemented the version file because other tools use it and because the app I was working on had a check for the SDK being installed and so I wanted a single source of truth. But I'd argue now that our tooling takes care of the installing, that check is redundant: the build script should trust the SDK is available.

I think the conversation should continue there, given that's the PR that is editing the script.

I also like the idea of having a list of valid numbers to pick from, given it's published at https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022 but I think that would be a good additional option, while we should still default to installing the latest valid version for the user ourselves.

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -1,19 +1,24 @@
...
-    Write-Host "--- :bug: Running as Administrator"
-} else {
-    Write-Host "--- :bug: Running as not Administrator"
-    $principal = New-Object Security.Principal.WindowsPrincipal([Security.Principal.WindowsIdentity]::GetCurrent())
-    $roles = $principal.Identity.Groups | ForEach-Object {
-        $_.Translate([Security.Principal.NTAccount]).Value
-    }
-    Write-Host "Your roles are:"
-    $roles | ForEach-Object { Write-Host "  - $_" }
-}
```
AliSoftware, line 14: Any reason why running on administrator seemed to have been a requirement in the past but isn't anymore?

TBH I came here to comment about the logic of that `if` condition being incorrect… because when I asked Cursor to analyze our PowerShell scripts using [the latest Cursor Rules I added to this repo](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/155), it mentioned that this logic was inverted (saying "Running as Administrator" in the `if (-not admin)` branch and vice-versa).

Now that that code is gone I guess it doesn't matter anymore that the logic was incorrect to begin with though…

mokagio:
Thanks @iangmaia for the missing `ps1` suggestions.

Incidentally, yesterday I looked into whether it was possible to remove the extension from the scripts to keep them consistent with the other bash ones. As far as I could see, it was not possible short of wrapping them in a cmdlet thus adding two files for each command. I think the "overhead" of adding the `.ps1` extension is acceptable in this case.

bin/prepare_windows_host_for_app_distribution.ps1
mokagio: Ahah, goes to show how useful and on point that check I wrote was 🤦‍♂️ 

Running as admin was never required. I don't remember exactly how I came to add that `-- :bug:` section but it would have been as part of trying to get something to run at a time where it didn't and me wondering whether the issue had to do with the role CI has.

It _might_ have been at the time of experimenting with [Podman](https://podman.io/) and [Fyne](https://fyne.io) for https://github.com/Automattic/download/

bin/prepare_windows_host_for_app_distribution.ps1
mokagio: I loaded this PR to leave a comment about adding the escape hatch in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153/commits/b78effb11bfe57b459c004dc120ecf4d009c87da and I had to smile because using `param` is exactly the implementation I got the LLM to do for me. 

image

It even used a similar name to the one you suggested @AliSoftware . However, the one I went for was a more verbose `SkipWindows10SDKInstallation`

AliSoftware: > It even used a similar name to the one you suggested @AliSoftware 

not-a-robot

bin/install_windows_10_sdk.ps1
mokagio: See de4505e

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is larger than 500 lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.</td>
    </tr>
  </tbody>
</table>

</p>

.buildkite/pipeline.yml
```diff
@@ -118,3 +118,98 @@ steps:
         notify:
           - github_commit_status:
               context: "pr_changed_files Tests: Edge Cases"
+
+  - group: ":windows: install_windows_10_sdk Tests"
```
AliSoftware, line 122: 💡  For future PR:

```suggestion
  - group: ":windows: install_windows_10_sdk Tests"
    if: build.env('INSTALL_WINDOWS_10_SDK_SCRIPT_CHANGED') == 'true'
```

+ in `shared-pipelines-vars` (that we'll need to introduce in this repo, apparently):
```bash
export INSTALL_WINDOWS_10_SDK_SCRIPT_CHANGED=$(bin/pr_changed_files --stdout --any-match 'bin/install_windows_10_sdk.ps1' 'tests/test-install-windows-10-sdk.ps1')
```
Which we'd be able to do in this repo despite the Uploader agent not loading the `a8c-ci-toolkit` plugin (which is what prevents us from doing that in other repos)… because in the particular case of this repo being the plugin itself, `bin/pr_changed_files` should be available directly after checkout 😉 

(PS: I thought I mentioned that idea somewhere already somewhere but couldn't find it, maybe it was in another PR… so I figured I'd make it appear here too anyway)

bin/install_windows_10_sdk.ps1
```diff
@@ -0,0 +1,106 @@
+# Install the Windows 10 SDK and Visual Studio Build Tools using the value in .windows-10-sdk-version.
+#
+# The expected .windows-10-sdk-version format is a integer representing a valid SDK component id.
```
```suggestion
# The expected .windows-10-sdk-version format is an integer representing a valid SDK component id.
```

bin/install_windows_10_sdk.ps1
AliSoftware: Maybe if we end up implementing the behavior of defaulting to install the latest valid version, we could allow `latest` to be a special valid content for `.windows-10-sdk-version`, that way if the file is not present it won't install the SDK at all, but if it's present with `latest` it will install the latest (and if it's present with a specific version it will deterministically install that specific version, acting as a lockfile) 🤔 

That way the implicit auto-install from `prepare_windows_host_for_app_distibution` if file is present would still work for that default-version fallback… 🤔

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -0,0 +1,123 @@
...
+#
+
+param (
+  [switch]$SkipWindows10SDKInstallation = $false
+)
+
+# Stop script execution when a non-terminating error occurs
+$ErrorActionPreference = "Stop"
+
+Write-Output "--- :windows: Setting up Windows for app distribution"
```
AliSoftware, line 23: Nitpick: I wonder if all those `Write-Output` (used in that script and the others) shouldn't be `Write-Host`, given they're intended to print informational logs (rather than passing results to a follow-up process down the pipeline) 🤔

AliSoftware (approved):
Looking good, left some nitpicks but none being blockers

PR #148, opened 2025-02-04
Add `pr_changed_files` script to detect changed files in a PR

### LINKED ISSUE #131: Add `pr_only_includes_changed_files_from` utility — to make it possible to skip CI jobs if e.g. no code change
Internal References: pe5sF9-38W-p2#comment-3981 , paaHJt-6pg-p2#comment-8788

## The Problem

Some CI jobs only makes sense to run on code changes (e.g. UI Tests), yet they run even if the changes in the PR are only about a `CHANGELOG.md` or `release-notes.txt` or similar things.

Because some of those jobs might take quite a long time to run by nature (e.g UI Tests), it would be useful to be able to skip them if we detect that no code file were changed.

This is an annoyance that has been reported a couple of times, especially by developers providing feedback after doing their Release Rotation (e.g. @itsmeichigo in pe5sF9-38W-p2#comment-3981)

Another use case would be for monorepos (e.g. `buildkite-ci`) for which some steps only make sense to run if the corresponding subproject / subdirectory has changed.

## Proposed Solution

 - Have a script in `a8c-ci-toolkit` that would allow to easily detect if the list of files changed by the PR only contains files from a specific set
 - Run that script in `.buildkite/shared-pipeline-vars` to set the value of an env var based on the result of the script
 - Use that env var as a condition (`if: !build.env('THEVAR')` on the step in the `.buildkite/*.yml` pipeline

## Avoid false-negative for use case of _skipping_ jobs

❌  At first one could think that we could implement a logic of "**only run** UI Tests **if app or test files changed**". For example only if the PR contains changes in `*.swift` files.

But this would risk leading to **false-negative** and _skip_ UI Tests for cases where they **_should_** actually have run—e.g. if one have updated image assets in the codebase that ends up make a screen higher than before and thus making a UI test fail to scroll enough; or if a project uses wiremock or similar and updated a `.json` file describing network calls…
And it would be cumbersome to list all the possible files that should led to a run of UI Tests, and easy to forget some cases.

👍  It would be less risky to approach this kind of logic the other way around, i.e. "**only skip** UI Tests **if only _non_-app-nor-test files changed**. (e.g. only skip UI Tests if the only files that changed are amongst `README.md`, `RELEASE-NOTES.txt`, …)

**This is because it's less problematic to have the UI Tests run when they could have been skipped than the other way around.**

So with that approach, even if we forget to list a file (e.g. `Config/Version.xcconfig`) for which the UI Tests could be skipped and UI Tests end up running, it's only a waste of CI time that could have been shaved, as opposed to UI Tests not being run when they should have and us failing to detect that the PR would break tests.

## Implementation Suggestions

Add a `a8c-ci-toolkit` utility named something like `pr_changed_files`, which could take a flag of either `--only-amongst`[^1]
or `--contains` [^2] + a list of file patterns, and return `true` or `false` accordingly.

[^1]: `--only-amongst` operator would be ideal for the use case of skipping steps like UI Tests
[^2]: `--contains` operator would be ideal for use cases of a monorepo wanting to only run a step if files from a subproject/subdir changed

 - The call site in `shared-pipeline-vars` could then look something like `SKIP_UI_TESTS=$(pr_changed_files --only-amongst RELEASE-NOTES.txt CHANGELOG.md fastlane/metadata/** Scripts/** docs/** README.md CODE-OF-CONDUCT.md CODEOWNERS CONTRIBUTING.md LICENSE)`
 - And finally the `pipeline.yml` can use it by adding e.g. `if: build.env('SKIP_UI_TESTS') == "true" on the `step` that runs UI Tests

Here are a couple of `git` tricks we can use to get the list of changed files from the PR
```bash
MERGE_BASE=$(git merge-base origin/"${BUILDKITE_PULL_REQUEST_BASE_BRANCH}" HEAD)
CHANGED_FILES=$(git --no-pager diff --name-only "$MERGE_BASE" HEAD)
```
Or more succintly:
```bash
CHANGED_FILES=$(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD)
```

Then we can loop through the list of `CHANGED_FILES` to  check if we find any in that list that is not in the list provided by`"$@"`.

Or if we want to avoid a double-nested-loop, we might be able to use `diff` to compare the two lists. Something like this maybe? (not tested)
```bash
diff <(git diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort) <(printf '%s\n' "$@" | sort | uniq) | grep -q '^< '
```
And check if the command returned 0 (i.e. it found at least one line in the `diff` that was on the `CHANGED_FILES` list but was not in the `"$@"` list) then we have extra files not in the list and we should make the script exit with non-zero, while if the command returned 1 (i.e. it did not find any diff line starting with `< ` so all the files listed on the left list were also present in the right list) we should make the script `exit 0` and succeed.

iangmaia:
Closes #131.

## Description

This PR adds a new utility script that will help optimize CI pipelines by avoiding unnecessary builds, allowing steps to be skipped based on which files changed in a PR, for example.
It works by comparing PR changes against the base branch and supports glob patterns for file matching.

## Usage

The `pr_changed_files` script can be used in the following ways:
- Check if there are any changes in a given PR (no arguments)
- Check if changes are limited to specific files (`--only-in`)
- Check if changes include specific files (`--includes`)

### Skip UI Tests when only docs changed

In `.buildkite/shared-pipeline-vars`, you can create a var that can later be used by conditions in YAML files:
```bash
SKIP_UI_TESTS=$(pr_changed_files --only-in "README.md" "docs/")
```

In `.buildkite/pipeline.yml`:
```yml
steps:
  label: "UI Tests"
  command: ./commands/run-ui-tests.sh
  if: !build.env('SKIP_UI_TESTS')
```

### Or run specific checks according to what was changed
```
CODE_CHANGED=$(pr_changed_files --includes "*.swift")
```

And in `.buildkite/pipeline.yml`:
```yml
steps:
  label: SwiftLint
  agent: linter
  command: swiftlint
  if: build.env('CODE_CHANGED')
```

### Just check if a PR has changes in a script

```bash
if [[ $(pr_changed_files) == "true" ]]; then
  echo "PR contains changes"
fi
```

---

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -93,27 +93,27 @@ steps:
           - github_commit_status:
               context: "install_swiftpm_dependencies Tests: Xcode Workspace and Project - Explicit"
 
-  - group: ":git: pr_changed_files Tests"
+  - group: ":github: pr_changed_files Tests"
     steps:
-      - label: ":git: Basic Changes Detection"
+      - label: ":github: pr_changed_files Tests - Basic Changes"
         command: tests/pr_changed_files/test_basic_changes.sh
         notify:
           - github_commit_status:
-              context: "pr_changed_files Tests: Basic Changes Detection"
+              context: "pr_changed_files Tests: Basic Changes"
 
-      - label: ":git: Pattern Matching - Includes"
+      - label: ":github: pr_changed_files Tests - Includes"
         command: tests/pr_changed_files/test_includes_patterns.sh
         notify:
           - github_commit_status:
-              context: "pr_changed_files Tests: Pattern Matching - Includes"
+              context: "pr_changed_files Tests: Includes"
 
-      - label: ":git: Pattern Matching - Only In"
+      - label: ":github: pr_changed_files Tests - Only In"
         command: tests/pr_changed_files/test_only_in_patterns.sh
         notify:
           - github_commit_status:
-              context: "pr_changed_files Tests: Pattern Matching - Only In"
+              context: "pr_changed_files Tests: Only In"
 
-      - label: ":git: Edge Cases"
+      - label: ":github: pr_changed_files Tests - Edge Cases"
         command: tests/pr_changed_files/test_edge_cases.sh
         notify:
           - github_commit_status:
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -92,3 +92,29 @@ steps:
         notify:
           - github_commit_status:
               context: "install_swiftpm_dependencies Tests: Xcode Workspace and Project - Explicit"
+
+  - group: ":git: pr_changed_files Tests"
+    steps:
+      - label: ":git: Basic Changes Detection"
+        command: tests/pr_changed_files/test_basic_changes.sh
+        notify:
+          - github_commit_status:
+              context: "pr_changed_files Tests: Basic Changes Detection"
+
+      - label: ":git: Pattern Matching - Includes"
+        command: tests/pr_changed_files/test_includes_patterns.sh
+        notify:
+          - github_commit_status:
+              context: "pr_changed_files Tests: Pattern Matching - Includes"
+
+      - label: ":git: Pattern Matching - Only In"
+        command: tests/pr_changed_files/test_only_in_patterns.sh
+        notify:
+          - github_commit_status:
+              context: "pr_changed_files Tests: Pattern Matching - Only In"
+
+      - label: ":git: Edge Cases"
+        command: tests/pr_changed_files/test_edge_cases.sh
+        notify:
+          - github_commit_status:
+              context: "pr_changed_files Tests: Edge Cases"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -101,17 +101,17 @@ steps:
           - github_commit_status:
               context: "pr_changed_files Tests: Basic Changes"
 
-      - label: ":github: pr_changed_files Tests - Includes"
-        command: tests/pr_changed_files/test_includes_patterns.sh
+      - label: ":github: pr_changed_files Tests - Any Match"
+        command: tests/pr_changed_files/test_any_match_patterns.sh
         notify:
           - github_commit_status:
-              context: "pr_changed_files Tests: Includes"
+              context: "pr_changed_files Tests: Any Match"
 
-      - label: ":github: pr_changed_files Tests - Only In"
-        command: tests/pr_changed_files/test_only_in_patterns.sh
+      - label: ":github: pr_changed_files Tests - All Match"
+        command: tests/pr_changed_files/test_all_match_patterns.sh
         notify:
           - github_commit_status:
-              context: "pr_changed_files Tests: Only In"
+              context: "pr_changed_files Tests: All Match"
 
       - label: ":github: pr_changed_files Tests - Edge Cases"
         command: tests/pr_changed_files/test_edge_cases.sh
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -4,18 +4,12 @@
 #
 # Usage:
 #   pr_changed_files                                  # Check if any files changed
-#   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
 #   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
+#   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
 #
 # Behavior:
 #   With no arguments:
 #     Returns "true" if the PR contains any changes, "false" otherwise
-#
-#   With --any-match:
-#     Returns "true" if ANY changed file matches ANY of the patterns
-#     Returns "false" if NONE of the changed files match ANY pattern
-#     Note: Will return "true" even if the PR includes other files not matching the patterns.
-#           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
 #
 #   With --all-match:
 #     Returns "true" if ALL changed files match AT LEAST ONE of the patterns
@@ -23,21 +17,27 @@
 #     Note: Will return "true" even if not all patterns are matched by the changed files.
 #           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
 #
+#   With --any-match:
+#     Returns "true" if ANY changed file matches ANY of the patterns
+#     Returns "false" if NONE of the changed files match ANY pattern
+#     Note: Will return "true" even if the PR includes other files not matching the patterns.
+#           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
+#
 # Examples with expected outputs:
 #   # Check if any files changed, returning "true" if PR has changes, "false" otherwise
 #   $ pr_changed_files
-#
-#   # Check if any Swift files changed (to decide if we should run SwiftLint)
-#   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
-#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
-#   → "true"  if PR changes `.swiftlint.yml`
-#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
 #
 #   # Check if only documentation files changed (to skip UI tests for example)
 #   $ pr_changed_files --all-match "*.md" "docs/*"
 #   → "true"  if PR only changes `docs/guide.md` and `README.md`
 #   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
 #   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
+#
+#   # Check if any Swift files changed (to decide if we should run SwiftLint)
+#   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
+#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → "true"  if PR changes `.swiftlint.yml`
+#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
 #
 # Returns:
 #   Prints "true" if the condition is met, "false" otherwise
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -27,6 +27,12 @@ if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
     exit 1
 fi
 
+# Ensure we have the base branch locally
+git fetch origin "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" >/dev/null 2>&1 || {
+    echo "Error: failed to fetch base branch '$BUILDKITE_PULL_REQUEST_BASE_BRANCH'" >&2
+    exit 1
+}
+
 mode=""
 patterns=()
 
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -14,28 +14,30 @@
 #   With --any-match:
 #     Returns "true" if ANY changed file matches ANY of the patterns
 #     Returns "false" if NONE of the changed files match ANY pattern
-#     Note: Will return "true" even if the PR includes other files not matching the patterns
+#     Note: Will return "true" even if the PR includes other files not matching the patterns.
+#           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
 #
 #   With --all-match:
 #     Returns "true" if ALL changed files match AT LEAST ONE of the patterns
 #     Returns "false" if ANY changed file doesn't match ANY pattern
-#     Note: Will return "true" even if not all patterns are matched by the changed files
+#     Note: Will return "true" even if not all patterns are matched by the changed files.
+#           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
 #
 # Examples with expected outputs:
 #   # Check if any files changed, returning "true" if PR has changes, "false" otherwise
 #   $ pr_changed_files
 #
-#   # Check if only documentation files changed (to skip UI tests for example)
-#   $ pr_changed_files --all-match "*.md" "docs/**"
-#   → "true"  if PR only changes docs/guide.md and README.md
-#   → "true"  if PR only changes docs/guide.md (not all patterns need to match)
-#   → "false" if PR changes docs/guide.md AND src/main.swift
-#
 #   # Check if any Swift files changed (to decide if we should run SwiftLint)
 #   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
-#   → "true"  if PR changes src/main.swift AND README.md
-#   → "true"  if PR changes .swiftlint.yml
-#   → "false" if PR only changes README.md
+#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → "true"  if PR changes `.swiftlint.yml`
+#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
+#
+#   # Check if only documentation files changed (to skip UI tests for example)
+#   $ pr_changed_files --all-match "*.md" "docs/**"
+#   → "true"  if PR only changes `docs/guide.md` and `README.md`
+#   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
+#   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
 #
 # Returns:
 #   Prints "true" if the condition is met, "false" otherwise
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -61,6 +61,10 @@ done
 # Get list of changed files as an array
 changed_files=()
 while IFS= read -r file; do
+    # Remove surrounding quotes if present and unescape special characters
+    file="${file#\"}"
+    file="${file%\"}"
+    file="${file//\\/}"
     changed_files+=("$file")
 done < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
 
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -11,15 +11,16 @@
 #   # Check if any files changed
 #   pr_changed_files
 #
-#   # Check if only documentation files changed
+#   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
 #   pr_changed_files --only-in "*.md" "docs/**"
 #
-#   # Check if any Swift files changed
-#   pr_changed_files --includes "*.swift"
+#   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
+#   pr_changed_files --includes "*.swift" ".swiftlint.yml"
 #
 # Returns:
 #   Prints "true" if the condition is met, "false" otherwise
-#   Always exits with code 0 to allow easy variable assignment
+#   Exits with code 0 regardless of if the condition was met or not, to allow easy variable assignment.
+#   Only exits with non-zero if the command invocation itself was incorrect (called outside of a PR context, incorrect arguments…)
 
 if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
     echo "Error: this tool can only be called from a Buildkite PR job" >&2
@@ -99,7 +100,7 @@ fi
 
 if [[ "$mode" == "only-in" ]]; then
     # Check if all changed files match at least one pattern
-    for file in $changed_files; do
+    for file in "$changed_files"; do
         matches_any=false
         for pattern in "${patterns[@]}"; do
             # shellcheck disable=SC2053
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -12,7 +12,7 @@
 #   pr_changed_files
 #
 #   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
-#   pr_changed_files --only-in "*.md" "docs/**"
+#   pr_changed_files --only-in "*.md" "docs/*"
 #
 #   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
 #   pr_changed_files --includes "*.swift" ".swiftlint.yml"
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -1,0 +1,124 @@
+#!/bin/bash -eu
+
+# This script checks if files are changed in a PR.
+#
+# Usage:
+#   pr_changed_files                                  # Check if any files changed
+#   pr_changed_files --includes <file-patterns...>    # Check if changes include files matching patterns
+#   pr_changed_files --only-in <file-patterns...>     # Check if changes are limited to given patterns
+#
+# Examples:
+#   # Check if any files changed
+#   pr_changed_files
+#
+#   # Check if only documentation files changed
+#   pr_changed_files --only-in "*.md" "docs/**"
+#
+#   # Check if any Swift files changed
+#   pr_changed_files --includes "*.swift"
+#
+# Returns:
+#   Prints "true" if the condition is met, "false" otherwise
+#   Always exits with code 0 to allow easy variable assignment
+
+if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
+    echo "Error: this tool can only be called from a Buildkite PR job" >&2
+    exit 1
+fi
+
+mode=""
+patterns=()
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --only-in)
+            if [[ -n "$mode" ]]; then
+                echo "Error: cannot specify both --only-in and --includes" >&2
+                exit 1
+            fi
+            mode="only-in"
+            shift
+            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
+                patterns+=("$1")
+                shift
+            done
+            ;;
+        --includes)
+            if [[ -n "$mode" ]]; then
+                echo "Error: cannot specify both --only-in and --includes" >&2
+                exit 1
+            fi
+            mode="includes"
+            shift
+            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
+                patterns+=("$1")
+                shift
+            done
+            ;;
+        --*) 
+            echo "Error: unknown option $1" >&2
+            exit 1
+            ;;
+        *)
+            echo "Error: unexpected argument $1" >&2
+            exit 1
+            ;;
+    esac
+done
+
+# Get list of changed files
+changed_files=$(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
+
+if [[ -z "$mode" ]]; then
+    # No arguments = any change
+    if [[ -n "$changed_files" ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ ${#patterns[@]} -eq 0 ]]; then
+    echo "Error: must specify at least one file pattern" >&2
+    exit 1
+fi
+
+# No files changed
+if [[ -z "$changed_files" ]]; then
+    if [[ "$mode" == "only-in" ]]; then
+        echo "true"
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ "$mode" == "only-in" ]]; then
+    # Check if all changed files match at least one pattern
+    for file in $changed_files; do
+        matches_any=false
+        for pattern in "${patterns[@]}"; do
+            if [[ "$file" == $pattern ]]; then
+                matches_any=true
+                break
+            fi
+        done
+        if [[ "$matches_any" == "false" ]]; then
+            echo "false"
+            exit 0
+        fi
+    done
+    echo "true"
+elif [[ "$mode" == "includes" ]]; then
+    # Check if any changed file matches any pattern
+    for file in $changed_files; do
+        for pattern in "${patterns[@]}"; do
+            if [[ "$file" == $pattern ]]; then
+                echo "true"
+                exit 0
+            fi
+        done
+    done
+    echo "false"
+fi
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -4,18 +4,18 @@
 #
 # Usage:
 #   pr_changed_files                                  # Check if any files changed
-#   pr_changed_files --includes <file-patterns...>    # Check if changes include files matching patterns
-#   pr_changed_files --only-in <file-patterns...>     # Check if changes are limited to given patterns
+#   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
+#   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
 #
 # Examples:
 #   # Check if any files changed
 #   pr_changed_files
 #
 #   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
-#   pr_changed_files --only-in "*.md" "docs/*"
+#   pr_changed_files --all-match "*.md" "docs/*"
 #
 #   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
-#   pr_changed_files --includes "*.swift" ".swiftlint.yml"
+#   pr_changed_files --any-match "*.swift" ".swiftlint.yml"
 #
 # Returns:
 #   Prints "true" if the condition is met, "false" otherwise
@@ -31,11 +31,11 @@ mode=""
 patterns=()
 
 # Define error message for mutually exclusive options
-EXCLUSIVE_OPTIONS_ERROR="Error: either specify --only-in or --includes; cannot specify both"
+EXCLUSIVE_OPTIONS_ERROR="Error: either specify --all-match or --any-match; cannot specify both"
 
 while [[ "$#" -gt 0 ]]; do
     case $1 in
-        --only-in | --includes)
+        --all-match | --any-match)
             if [[ -n "$mode" ]]; then
                 echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
                 exit 1
@@ -85,7 +85,7 @@ fi
 
 # No files changed
 if [[ ${#changed_files[@]} -eq 0 ]]; then
-    if [[ "$mode" == "only-in" ]]; then
+    if [[ "$mode" == "all-match" ]]; then
         echo "true"
     else
         echo "false"
@@ -93,7 +93,7 @@ if [[ ${#changed_files[@]} -eq 0 ]]; then
     exit 0
 fi
 
-if [[ "$mode" == "only-in" ]]; then
+if [[ "$mode" == "all-match" ]]; then
     # Check if all changed files match at least one pattern
     for file in "${changed_files[@]}"; do
         matches_any=false
@@ -110,7 +110,7 @@ if [[ "$mode" == "only-in" ]]; then
         fi
     done
     echo "true"
-elif [[ "$mode" == "includes" ]]; then
+elif [[ "$mode" == "any-match" ]]; then
     # Check if any changed file matches any pattern
     for file in "${changed_files[@]}"; do
         for pattern in "${patterns[@]}"; do
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -29,11 +29,14 @@ fi
 mode=""
 patterns=()
 
+# Define error message for mutually exclusive options
+EXCLUSIVE_OPTIONS_ERROR="Error: either specify --only-in or --includes; cannot specify both"
+
 while [[ "$#" -gt 0 ]]; do
     case $1 in
         --only-in)
             if [[ -n "$mode" ]]; then
-                echo "Error: cannot specify both --only-in and --includes" >&2
+                echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
                 exit 1
             fi
             mode="only-in"
@@ -45,7 +48,7 @@ while [[ "$#" -gt 0 ]]; do
             ;;
         --includes)
             if [[ -n "$mode" ]]; then
-                echo "Error: cannot specify both --only-in and --includes" >&2
+                echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
                 exit 1
             fi
             mode="includes"
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -34,7 +34,7 @@
 #   → "false" if PR only changes `README.md` (none of files match any of the patterns)
 #
 #   # Check if only documentation files changed (to skip UI tests for example)
-#   $ pr_changed_files --all-match "*.md" "docs/**"
+#   $ pr_changed_files --all-match "*.md" "docs/*"
 #   → "true"  if PR only changes `docs/guide.md` and `README.md`
 #   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
 #   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
@@ -115,9 +115,7 @@ fi
 file_matches_any_pattern() {
     local file="$1"
     shift
-    local patterns=("$@")
-
-    for pattern in "${patterns[@]}"; do
+    for pattern in "$@"; do
         # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
         if [[ "$file" == ${pattern} ]]; then
             return 0
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -93,13 +93,9 @@ done
 
 # Get list of changed files as an array
 changed_files=()
-while IFS= read -r file; do
-    # Remove surrounding quotes if present and unescape special characters
-    file="${file#\"}"
-    file="${file%\"}"
-    file="${file//\\/}"
-    changed_files+=("$file")
-done < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
+while IFS= read -r -d '' file; do
+  changed_files+=("$file")
+done < <(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
 
 if [[ -z "$mode" ]]; then
     # No arguments = any change
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -99,18 +99,25 @@ if [[ ${#changed_files[@]} -eq 0 ]]; then
     exit 0
 fi
 
+# Returns 0 if the file matches any of the patterns, 1 otherwise
+file_matches_any_pattern() {
+    local file="$1"
+    shift
+    local patterns=("$@")
+
+    for pattern in "${patterns[@]}"; do
+        # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
+        if [[ "$file" == ${pattern} ]]; then
+            return 0
+        fi
+    done
+    return 1
+}
+
 if [[ "$mode" == "all-match" ]]; then
     # Check if all changed files match at least one pattern
     for file in "${changed_files[@]}"; do
-        matches_any=false
-        for pattern in "${patterns[@]}"; do
-            # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
-            if [[ "$file" == ${pattern} ]]; then
-                matches_any=true
-                break
-            fi
-        done
-        if [[ "$matches_any" == "false" ]]; then
+        if ! file_matches_any_pattern "$file" "${patterns[@]}"; then
             echo "false"
             exit 0
         fi
@@ -119,13 +126,10 @@ if [[ "$mode" == "all-match" ]]; then
 elif [[ "$mode" == "any-match" ]]; then
     # Check if any changed file matches any pattern
     for file in "${changed_files[@]}"; do
-        for pattern in "${patterns[@]}"; do
-            # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
-            if [[ "$file" == ${pattern} ]]; then
-                echo "true"
-                exit 0
-            fi
-        done
+        if file_matches_any_pattern "$file" "${patterns[@]}"; then
+            echo "true"
+            exit 0
+        fi
     done
     echo "false"
 fi
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -59,7 +59,8 @@ while [[ "$#" -gt 0 ]]; do
 done
 
 # Get list of changed files
-changed_files=$(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
+# Use -z to get null-terminated output and avoid escaping issues
+changed_files=$(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | tr '\0' '\n' | sort)
 
 if [[ -z "$mode" ]]; then
     # No arguments = any change
@@ -88,11 +89,14 @@ fi
 
 if [[ "$mode" == "only-in" ]]; then
     # Check if all changed files match at least one pattern
-    for file in "$changed_files"; do
+    while IFS= read -r file; do
+        [[ -z "$file" ]] && continue
+        # Remove any escaping of special characters
+        file="${file//\\/}"
         matches_any=false
         for pattern in "${patterns[@]}"; do
             # shellcheck disable=SC2053
-            if [[ "$file" == $pattern ]]; then # glob match with pattern
+            if [[ "$file" == ${pattern} ]]; then
                 matches_any=true
                 break
             fi
@@ -101,18 +105,21 @@ if [[ "$mode" == "only-in" ]]; then
             echo "false"
             exit 0
         fi
-    done
+    done <<< "$changed_files"
     echo "true"
 elif [[ "$mode" == "includes" ]]; then
     # Check if any changed file matches any pattern
-    for file in $changed_files; do
+    while IFS= read -r file; do
+        [[ -z "$file" ]] && continue
+        # Remove any escaping of special characters
+        file="${file//\\/}"
         for pattern in "${patterns[@]}"; do
             # shellcheck disable=SC2053
-            if [[ "$file" == $pattern ]]; then # glob match with pattern
+            if [[ "$file" == ${pattern} ]]; then
                 echo "true"
                 exit 0
             fi
         done
-    done
+    done <<< "$changed_files"
     echo "false"
 fi
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -95,7 +95,7 @@ if [[ "$mode" == "only-in" ]]; then
         file="${file//\\/}"
         matches_any=false
         for pattern in "${patterns[@]}"; do
-            # shellcheck disable=SC2053
+            # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
             if [[ "$file" == ${pattern} ]]; then
                 matches_any=true
                 break
@@ -114,7 +114,7 @@ elif [[ "$mode" == "includes" ]]; then
         # Remove any escaping of special characters
         file="${file//\\/}"
         for pattern in "${patterns[@]}"; do
-            # shellcheck disable=SC2053
+            # shellcheck disable=SC2053 # We don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
             if [[ "$file" == ${pattern} ]]; then
                 echo "true"
                 exit 0
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -68,6 +68,11 @@ while [[ "$#" -gt 0 ]]; do
             fi
             mode="${1#--}"
             shift
+            # Check if there are any patterns after the flag
+            if [[ "$#" -eq 0 || "$1" == "--"* ]]; then
+                echo "Error: must specify at least one file pattern" >&2
+                exit 1
+            fi
             while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
                 patterns+=("$1")
                 shift
@@ -97,21 +102,6 @@ done < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_B
 if [[ -z "$mode" ]]; then
     # No arguments = any change
     if [[ ${#changed_files[@]} -gt 0 ]]; then
-        echo "true"
-    else
-        echo "false"
-    fi
-    exit 0
-fi
-
-if [[ ${#patterns[@]} -eq 0 ]]; then
-    echo "Error: must specify at least one file pattern" >&2
-    exit 1
-fi
-
-# No files changed
-if [[ ${#changed_files[@]} -eq 0 ]]; then
-    if [[ "$mode" == "all-match" ]]; then
         echo "true"
     else
         echo "false"
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -58,13 +58,15 @@ while [[ "$#" -gt 0 ]]; do
     esac
 done
 
-# Get list of changed files
-# Use -z to get null-terminated output and avoid escaping issues
-changed_files=$(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | tr '\0' '\n' | sort)
+# Get list of changed files as an array
+changed_files=()
+while IFS= read -r file; do
+    changed_files+=("$file")
+done < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
 
 if [[ -z "$mode" ]]; then
     # No arguments = any change
-    if [[ -n "$changed_files" ]]; then
+    if [[ ${#changed_files[@]} -gt 0 ]]; then
         echo "true"
     else
         echo "false"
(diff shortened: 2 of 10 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
...
+        echo "true"
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ "$mode" == "only-in" ]]; then
+    # Check if all changed files match at least one pattern
+    for file in $changed_files; do
```
iangmaia: I couldn't escape from the nested loop. I tried @AliSoftware's suggestion from https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/issues/131 (using `diff` to compare the lists) but I got some errors in some of the scenarios. Also, the code was getting a bit harder to understand than a simple dumb nested loop.

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
...
+#   pr_changed_files --includes "*.swift"
+#
+# Returns:
+#   Prints "true" if the condition is met, "false" otherwise
+#   Always exits with code 0 to allow easy variable assignment
+
+if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
+    echo "Error: this tool can only be called from a Buildkite PR job" >&2
+    exit 1
+fi
```
mokagio, line 50: Given this script expects to run in Buildkite, would it make sense to add `buildkite` somewhere in the name?

On the other hand, we are within the a8c-ci-toolkit-**buildkite-plugin** so maybe it's fair to take dependency on Buildkite and its env var as a given. 🤔

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
...
+
+mode=""
+patterns=()
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --only-in)
+            if [[ -n "$mode" ]]; then
+                echo "Error: cannot specify both --only-in and --includes" >&2
```
mokagio: Double nitpick.

1. The message is duplicated below, so we could DRY it in a var
2. What do you think of starting with positive language? 

> Error: either specify --only-in or --includes; cannot specify both.

bin/pr_changed_files
mokagio: [I asked ChatGPT 4o](https://chatgpt.com/share/67a2f064-e6d8-8011-ae20-318d56d2a12f) for suggestion and it generated this, after a bit of back and forth where I asked about edge cases in file names and long lists

```bash
#!/bin/bash -eu

# This script checks if files are changed in a PR.
#
# Usage:
#   pr_changed_files                                  # Check if any files changed
#   pr_changed_files --includes <file-patterns...>    # Check if changes include files matching patterns
#   pr_changed_files --only-in <file-patterns...>     # Check if changes are limited to given patterns
#
# Examples:
#   # Check if any files changed
#   pr_changed_files
#
#   # Check if only documentation files changed
#   pr_changed_files --only-in "*.md" "docs/**"
#
#   # Check if any Swift files changed
#   pr_changed_files --includes "*.swift"
#
# Returns:
#   Prints "true" if the condition is met, "false" otherwise
#   Always exits with code 0 to allow easy variable assignment

if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
    echo "Error: this tool can only be called from a Buildkite PR job" >&2
    exit 1
fi

mode=""
patterns=()

while [[ "$#" -gt 0 ]]; do
    case $1 in
        --only-in)
            if [[ -n "$mode" ]]; then
                echo "Error: cannot specify both --only-in and --includes" >&2
                exit 1
            fi
            mode="only-in"
            shift
            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
                patterns+=("$1")
                shift
            done
            ;;
        --includes)
            if [[ -n "$mode" ]]; then
                echo "Error: cannot specify both --only-in and --includes" >&2
                exit 1
            fi
            mode="includes"
            shift
            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
                patterns+=("$1")
                shift
            done
            ;;
        --*) 
            echo "Error: unknown option $1" >&2
            exit 1
            ;;
        *)
            echo "Error: unexpected argument $1" >&2
            exit 1
            ;;
    esac
done

# Get list of changed files safely (handling spaces and special characters)
mapfile -t changed_files < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)

if [[ -z "$mode" ]]; then
    # No arguments = check if any files changed
    if [[ ${#changed_files[@]} -gt 0 ]]; then
        echo "true"
    else
        echo "false"
    fi
    exit 0
fi

if [[ ${#patterns[@]} -eq 0 ]]; then
    echo "Error: must specify at least one file pattern" >&2
    exit 1
fi

# No files changed
if [[ ${#changed_files[@]} -eq 0 ]]; then
    if [[ "$mode" == "only-in" ]]; then
        echo "true"
    else
        echo "false"
    fi
    exit 0
fi

# Escape patterns for safe regex matching
escaped_patterns=()
for pattern in "${patterns[@]}"; do
    escaped_patterns+=("$(printf '%s' "$pattern" | sed 's/[][.*^$()|?+{}\\]/\\&/g')")
done
pattern_regex=$(IFS="|"; echo "${escaped_patterns[*]}")

# Store changed files in a temp file for safe handling
temp_file=$(mktemp)
printf "%s\n" "${changed_files[@]}" > "$temp_file"

if [[ "$mode" == "only-in" ]]; then
    # Check if all changed files match at least one pattern
    if grep --extended-regexp --invert-match --quiet "^($pattern_regex)$" "$temp_file"; then
        echo "false"
        exit 0
    fi
    echo "true"
elif [[ "$mode" == "includes" ]]; then
    # Check if any changed file matches any pattern
    if grep --extended-regexp --quiet "^($pattern_regex)$" "$temp_file"; then
        echo "true"
        exit 0
    fi
    echo "false"
fi

rm -f "$temp_file" # Cleanup temporary file
```

I haven't tested it, so take it with a pinch of salt.

Assuming it works, I _think_ that implementation would be more efficient than the nested loop. However, I wonder how often we'd run into this being a performance bottleneck...

mokagio (commented):
I don't know whether the implementation ChatGPT suggested is genuinely better, but I do like how it accounts for spaces in the file names etc.

Regardless, I think we might either come up with a way to test this script at the repo level, or bite the bullet and start using it (maybe in the repos that our team owns) and see how it behaves. I _think_ it should be possible to devise a few integration tests, maybe with the help of Ruby, where we make commits in a dummy repo and then run the script and check what it returns.

bin/pr_changed_files
iangmaia: > we are within the a8c-ci-toolkit-buildkite-plugin so maybe it's fair to take dependency on Buildkite and its env var as a given

🎯 And that also applies to a few other commands that rely on Buildkite env vars.

bin/pr_changed_files
iangmaia: Though it does create a temp file. So I'd imagine using IO wouldn't be as efficient as an in-memory loop 🤔 

After a few attempts, Claude suggested to convert the glob patterns to grep patterns and I liked how it looks as the result is clean and still readable, but it felt a bit more fragile? Not sure how well this conversion would work and if it would work well in all scenarios.

```bash
# (...)

# Convert glob patterns to grep patterns
grep_pattern=$(printf '%s|' "${patterns[@]}" | sed 's/\*\*/.*/' | sed 's/\*/[^\/]*/' | sed 's/|$//')

if [[ "$mode" == "only-in" ]]; then
    # Check if all changed files match the pattern
    if echo "$changed_files" | grep -qvE "$grep_pattern"; then
        echo "false"
    else
        echo "true"
    fi
elif [[ "$mode" == "includes" ]]; then
    # Check if any changed file matches the pattern
    if echo "$changed_files" | grep -qE "$grep_pattern"; then
        echo "true"
    else
        echo "false"
    fi
fi
```

I think it's probably a good idea to write a few tests to potentially explore other implementations, I'll look into it.

AliSoftware: > ```
> # Get list of changed files safely (handling spaces and special characters)
> mapfile -t changed_files < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
> ```

`mapfile` bash built-in is only available in bash4 iinm. Which is not always the version available in all our agents (e.g. the default bash in macOS is 3.2.57)

bin/pr_changed_files
iangmaia: 👍 
Addressed in 4dd491c28c7402b964616910303fe9af84af369d

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
+#!/bin/bash -eu
+
+# This script checks if files are changed in a PR.
+#
+# Usage:
+#   pr_changed_files                                  # Check if any files changed
+#   pr_changed_files --includes <file-patterns...>    # Check if changes include files matching patterns
+#   pr_changed_files --only-in <file-patterns...>     # Check if changes are limited to given patterns
```
AliSoftware: I wonder if it'd be worth it to be a bit more explicit in the documentation of those flags, like

 - For `--includes`, be more clear/explicit about "Still returns true if the PR also includes other files in addition to including all the ones listed in file-patterns, but will return false if at least one of the file-patterns don't match any of the files in the PR"
 - For `--only-in`, be more clear/explicit about "Still returns true if the PR does not contain all of the files listed in the file-patterns but only matches some of the patterns, as long as the PR does NOT contain any file in the PR that would NOT match any pattern at all"

I'm not super happy with the wording I just suggested, so feel free to come up with something better, but I still think having the doc be extra clear on those could be useful.

Another approach to help be clearer about the behavior would be to provide examples. Either in the `Examples` section you wrote below, also explain what would be the output depending on the files present in the diff for each invocation, or provide some table with use cases, like this:

```
| Option used | Files in PR | Output | Why? |
| `--only-in "*.md" "docs/**"` | `docs/demo.png` | `true` | All files match at least one pattern |
| `--only-in "*.md" "docs/**"` | `Fastfile` | `false` | `Fastfile` does not match any pattern |
| `--only-in "*.md" "docs/**"` | `README.md`, `Fastfile` | `false` | `Fastfile` does not match any pattern |
| `--only-in "*.md" "docs/**"` | (empty diff) | `true` | There was no file that didn't match any pattern since there was no file at all |
| `--includes "*.swift"` | `README.md` | `false` | No file matching any of the patterns |
| `--includes "*.swift"` | `Foo.swift`, `README.md` | `true` | At least one file (`Foo.swift`) matches one of the patterns |
| `--includes "*.swift"` | `.swiftlint` | `false` | None of the files match any of the pattern |
| `--includes "*.swift" ".swiftlint.yml"` | `.swiftlint.yml`, `README.md` | `true` | At least one of the files (`.swiftlint.yml`) matches one of the patterns |
```

While writing those examples I actually got a doubt about the last use case of `--includes "*.swift" ".swiftlint.yml"` because I started to wonder if the patterns we provide to `--include` were acting as an AND or an OR. In practice that wouldn't make sense to make them be interpreted as an AND so I think that's correct, but it shows how providing concrete examples highlighting all the possible use cases of mode and input patterns vs the files in the PR can help thinking about all edge cases.

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
...
+#
+# Examples:
+#   # Check if any files changed
+#   pr_changed_files
+#
+#   # Check if only documentation files changed
+#   pr_changed_files --only-in "*.md" "docs/**"
```
AliSoftware:
```suggestion
#   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
#   pr_changed_files --only-in "*.md" "docs/**"
```

bin/pr_changed_files
```diff
@@ -0,0 +1,126 @@
...
+#
+#   # Check if any Swift files changed
+#   pr_changed_files --includes "*.swift"
```
AliSoftware: Since this one is likely to be used for deciding to run Swiftlint or not, I'd make sure the example suggests the proper patterns that we should use in that case, i.e. including the `.swiftlint.yml` file that would require linting as well if it were changed too

```suggestion
#   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
#   pr_changed_files --includes "*.swift" ".swiftlint.yml"
```

While I'm writing this, I am still a bit unsure how that reads, because intuitively if I see a command spelled `--includes "*.swift" ".swiftlint.yml"` I'd likely instinctively read it as "Includes at least one Swift file AND the `.swiftlint.yml` files". While in practice it's more of an OR in the implementation (at least I think it should be). So I wonder if there's a better wording to make this clearer (bike-shedding time!). `--includes-at-least-one-of` seems too wordy, `--one-of` makes it sound like an XOR, … so not sure what'd be best, open to suggestions.

Or maybe `--includes` is enough and OK and I am the only one who think it might be confusing… 🤷

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+                echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
+                exit 1
+            fi
+            mode="includes"
+            shift
+            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
+                patterns+=("$1")
+                shift
+            done
+            ;;
```
AliSoftware: In order to DRY things you can just combine the 2 cases in a single one, since all their code are the same except the `mode=` (and use `mode=${1#--}` to use `$1` without the `--` prefix for that assignment).

```suggestion
        --only-in | --includes)
            if [[ -n "$mode" ]]; then
                echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
                exit 1
            fi
            mode="${1%--}"
            shift
            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
                patterns+=("$1")
                shift
            done
            ;;
```

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+#
+#   # Check if only documentation files changed
+#   pr_changed_files --only-in "*.md" "docs/**"
+#
+#   # Check if any Swift files changed
+#   pr_changed_files --includes "*.swift"
+#
+# Returns:
+#   Prints "true" if the condition is met, "false" otherwise
+#   Always exits with code 0 to allow easy variable assignment
```
AliSoftware: To be fair, the script sometimes `exit 1`… but not because it printed `false`, but instead because of a misuse of the command, e.g. use of both `--only-in` and `--includes` in the invocation.

```suggestion
#   Exits with code 0 regardless of if the condition was met or not, to allow easy variable assignment.
#   Only exits with non-zero if the command invocation itself was incorrect (called outside of a PR context, incorrect arguments…)
```

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+            ;;
+        *)
+            echo "Error: unexpected argument $1" >&2
+            exit 1
+            ;;
+    esac
+done
+
+# Get list of changed files
+changed_files=$(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
```
AliSoftware: I wonder if we'd need to `git fetch origin $BUILDKITE_PULL_REQUEST_BASE_BRANCH` explicitly first to make sure the agent locally knows about that branch existing and not fail complaining about unknown reference… especially since we do a clean checkout (even if using `--reference <cache>`) on each job using a non-persistent agent (most of them), and Buildkite checks out a specific commit sha1 at the start of each job instead of checking out branches.

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ ${#patterns[@]} -eq 0 ]]; then
+    echo "Error: must specify at least one file pattern" >&2
+    exit 1
+fi
```
AliSoftware: Could be worth moving this inside the `case` on line 48, right after the `while` that parses the patterns (or instead testing for `$#` just _before_ starting the `while` that builds the `patterns` array) so that such an error is caught earlier in the script (before the run of `git diff` above)?

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+        echo "true"
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ "$mode" == "only-in" ]]; then
+    # Check if all changed files match at least one pattern
+    for file in $changed_files; do
```
AliSoftware: I think we might need quotes here to prevent issues with spaces in file names?
```suggestion
    for file in "$changed_files"; do
```

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+        matches_any=false
+        for pattern in "${patterns[@]}"; do
+            # shellcheck disable=SC2053
```
AliSoftware: Might be worth a comment explaining what the shellcheck is about and why we needed to disable it?

```suggestion
            # shellcheck disable=SC2053 # It's expected that we don't quote the rhs in the condition on the next line because we want to interpret pattern as a glob pattern
```

The [shellcheck doc](https://www.shellcheck.net/wiki/Ignore) is not super clear if it's possible to provide such a comment on the same line for inline magic comments like those… but probably easy to check?

And if not, maybe your `# glob match with pattern` comment on the next line was there to intend documenting exactly this in a different place instead; but I think in that case it might as well be more explicit about it being related to the shellcheck ignored warning, maybe as a comment on a dedicated line between line 105 and 106 then?

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+    else
+        echo "false"
+    fi
+    exit 0
+fi
+
+if [[ "$mode" == "only-in" ]]; then
+    # Check if all changed files match at least one pattern
+    for file in $changed_files; do
+        matches_any=false
```
AliSoftware: 💡  This choice of variable name made me realize that… this could be a viable alternative for the name of the flags:
 - `pr_changed_files --any "*.swift" ".swiftlint.yml"` (instead of `pr_changed_files --includes`) to return `true` if **_any_** of the changed files contain Swift files or swiftlint config, and false if none of the changed files match one of the provided pattern
 - `pr_changed_files --all "**/*.md" "docs/**"` (instead of `pr_changed_files --only-in`) to return `true` if **all** the changed files are matching one of the provided patterns

Maybe that would help lessen the ambiguity I discussed in my comment above around the documentation comments at the start of the script? 🤔

bin/pr_changed_files
```diff
@@ -0,0 +1,129 @@
...
+                matches_any=true
+                break
+            fi
+        done
+        if [[ "$matches_any" == "false" ]]; then
+            echo "false"
+            exit 0
+        fi
+    done
+    echo "true"
```
AliSoftware: One way to maybe make this more readable would be to extract the test of if `$file` matches at least one of the `$patterns`… in a dedicated function, like `file_matches_any_pattern` that would return `0` if match found and `1`  if not.

Then you'd be able to use it:
 - Here inside your `for file in "$changed_files"` loop for `only-in` to just say `if ! file_matches_any_pattern "$file" "$patterns[@]"; then echo "false"; exit 0; fi`, and still with the `echo "true"` outside the file loop for if all files passed the test
 - In the `includes` mode by using `if file_matches_any_pattern; then echo "false"; exit 0; fi", still with the `echo "false"` outside the file loop for if none of the files passed the test

bin/pr_changed_files
AliSoftware: > I don't know whether the implementation ChatGPT suggested is genuinely better, but I do like how it accounts for spaces in the file names etc.

Actually, the more I read the alternate script proposed by ChatGPT that @mokagio shared above, the more I find issues with it:
 - As already mentioned above, `mapfile` is not supported in the bash versions we plan to use this script in, so that would not work to begin with
 - The code that tries to convert our glob patterns to regex patterns (`escaped_patterns+=("$(printf '%s' "$pattern" | sed 's/[][.*^$()|?+{}\\]/\\&/g')")`) is quite unreadable. I'm not even sure it's correct.
 - The code creates a temp file while it'd be better to avoid it
 - The code feels less readable 

By contrast, I _think_ that just adding quotes around the `for file in "$changed_files"` would be enough to account for potential file names with spaces.

> After a few attempts, Claude suggested to convert the glob patterns to grep patterns and I liked how it looks as the result is clean and still readable, but it felt a bit more fragile? Not sure how well this conversion would work and if it would work well in all scenarios.

Yep, same concern with the conversion of the glob pattern to a grep pattern. Not only does it feel less readable to have that conversion logic in, but I'm also not confident that conversion would cover all cases.
If we decide that it'd be nicer to use a grep/regex pattern rather than a glob/fnmatch pattern then I'd instead suggest that we let the caller provide the input parameters as a grep/regex pattern directly to begin with (`pr_changed_files ".*\.swift"`).

But I don't think that'd be a good idea as that would make it harder to express glob patterns for subdirectories (like `docs/**` or `**/*.md`, while `docs/[^/]*` doesn't read as nice). And that also makes me confirm that I'm not confident that the `sed` command to try turning a glob pattern into a regex suggested by Claude would cover all cases without encountering any edge cases that would make the resulting regex not match what we initially intended…

---

To be honest, especially if we also (1) confirm that my suggestion to add quotes around "$changed_files` would solve the case of files with spaces and (2) go with [the code DRY-ing I've suggested below](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1942797360) for testing if a file matches any of the pattern, I think the end result would be readable enough, even if it involves nested for loops.

Also, better have an implementation that we as human are able to understand and reason about, and be confident about the behavior being implemented, than one of the implementations that ChatGPT or Claude suggested above which would require a lot more brainpower to decipher, or worse, to debug if we ended up noticing the script didn't behave as intended in some situations. If the AI suggested answers led to a more performant and efficient implementation _that would also be readable and compatible with all version of bash_ then I'd be happy to take their recommendations, but in this case it seems to me that those suggestions end up being worse or less readable and stable than ours.

bin/pr_changed_files
iangmaia: Hmm yeah I have the same thought when reading multiple parameters: "includes" might give the idea it should be all arguments 🤔

bin/pr_changed_files
iangmaia: Oh I think that might be a good alternative indeed! We could explore `--any-match` / `--only-matches` | `--all-matches`, but `--any` / `--all` already sounds good to me.

AliSoftware: I like `--any-match` / `--all-match` actually, because after thinking about it I think `--all` would have the same issue as `--includes` with multiple patterns: `--all "*.swift" ".swift.yml"` could be misread "All those patterns must match at least one file" (i.e. we need to have at least one Swift file AND the `.swiftlint.yml` file in the changed files) instead of "All files much match at least one pattern" (what we actually mean)

bin/pr_changed_files
```diff
@@ -0,0 +1,130 @@
...
+#
+# Examples:
+#   # Check if any files changed
+#   pr_changed_files
+#
+#   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
+#   pr_changed_files --only-in "*.md" "docs/**"
+#
+#   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
+#   pr_changed_files --includes "*.swift" ".swiftlint.yml"
```
AliSoftware: Those examples just made me realize:
 - Do we want `*.md` and `*.swift` syntaxes to mean to only match Markdown and Swift files at the root of the repo, and thus needing to use `**/*.md` and `**/*.swift` explicitly if we want to match those files from in any directory instead?
 - On the other hand, `**` is not supported in bash 3. For that pattern to work as expected (match subdirectories in a pathname string), we'd need to use `shopt -s globstar`. Except that this shell option was only added in `bash 4`

So maybe `*.swift` is indeed what we want to match "any swift file anywhere in the directory structure" after all given no support for `globstar` to begin with. But that also means that we probably want to change the `docs/**` to just a `docs/*` in this documentation comment, to avoid risk of confusion and making the user think it'd be supported.

```suggestion
#   # Check if only documentation files changed. Could be useful to decide to skip UI tests for example
#   pr_changed_files --only-in "*.md" "docs/*"
#
#   # Check if any Swift files changed, e.g. to decide if we should run Swiftlint
#   pr_changed_files --includes "*.swift" ".swiftlint.yml"
```

See https://www.gnu.org/software/bash/manual/html_node/Pattern-Matching.html—that we could actually consider mentioning in the docs as well… as long as we mention that neither `globstar` nor `extglob`[^1] shell options will be enabled when evaluating the patterns passed to this command.

[^1]: In theory we _could_ call `shopt -s extglob` in this script to support extended glob syntax even in bash 3… but this will result in the pattern matching being slower while I don't think we'll even need such advanced patterns in our use cases, so I don't think it's worth enabling that one.

bin/pr_changed_files
iangmaia: Exactly, shellcheck doesn't clearly say whether inline comments would work. (update: it seems it does: d698f4274aa2c7b203f0de31dd8e63896d5813a3)

bin/pr_changed_files
AliSoftware: Ok adding quotes in `for file in "$changed_files"` seems to not do what I'd expect—[thanks `shellcheck` for catching that](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/931#0194d747-3945-4fda-b00a-70f31c133eca/165-175)!
```
     for file in "$changed_files"; do
--
                 ^--------------^ SC2066 (error): Since you double quoted this, it will not word split, and the loop will only run once.
```

Which means maybe we instead need to turn `changed_files` into a bash array (similar to the `$patterns` array) to handle this properly. Which iinm can probably be handled using something `read`:

```bash
IFS=$'\n' read -r -d '' -a changed_files < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
```
And then at that point I think a `for file in "$changed_files[@]"` [EDIT]or maybe using `while read` for the loop here too?[/EDIT] would finally work as expected?

Alternatively I think if we replace this `for file in "$changed_files"` with a `IFS=$'\n' while read -r -d ''; do … done <<<"$changed_files"` this should also work to loop through the lines of the `$changed_files` variable.

.buildkite/pipeline.yml
```diff
@@ -92,3 +92,29 @@ steps:
         notify:
           - github_commit_status:
               context: "install_swiftpm_dependencies Tests: Xcode Workspace and Project - Explicit"
+
+  - group: ":git: pr_changed_files Tests"
```
AliSoftware: ❤️

bin/pr_changed_files
```diff
@@ -0,0 +1,125 @@
...
+        *)
+            echo "Error: unexpected argument $1" >&2
+            exit 1
+            ;;
+    esac
+done
+
+# Get list of changed files
+# Use -z to get null-terminated output and avoid escaping issues
+changed_files=$(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | tr '\0' '\n' | sort)
```
AliSoftware: Feels strange to use `-z` only to pipe the output to `tr '\0' '\n'`… wouldn't that end up doing the same as just letting `git diff` output the `\n` directly? Or I guess the idea is that it'd handle potential escapes in file names (e.g. a file strangely named `foo\bar.txt`) by having `-z` escaping those…? But then I see that you have to manually parse and remove them on L114–115, so…)

---

1️⃣  As mentioned in my [recent comment](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1943474741), I think at that point it'd be better to turn this variable into a bash array, to make sure the case of spaces in file names is better handled all while making the script more resilient (using arrays here for an array of files makes more sense than handling it as a scalar variable with newlines).

2️⃣  Alternatively, we could store the git command in an array, and only execute it as part of the `while read` loops that you used below?
In other words, instead of executing the git command then storing the list in `changed_files` to use `IFS= while read … <<< "$changed_files"`, you'd set `git_command=(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD)` then use `IFS= while read … < <("$git_command[@]")` when looping through the files.

I think both solutions (1️⃣  or 2️⃣ ) should be equally valid, so up to you whatever you prefer. Personally I think I prefer 1️⃣  even if it means two loops (a `while read` to turn the result of `git diff` into an array then later another one to loop through that array), as it feels more readable…
But in any case, working with bash arrays seems safer and also simpler, without the issues of escapes (as long as you use `IFS=` or `IFS=$'\n'`, and `-r -d''` when calling `read`)?

bin/pr_changed_files
iangmaia: Though we may exit early if there are no arguments (any change = true).

AliSoftware: Yes but if you put that check about patterns not being empty inside the `case --only-in | --includes)` (right after the `while` loop that parses the arguments and put them into `patterns+=($1)`), that would only run that check in the case where we use either `--only-in` or `--includes`, while not checking for that in the case where neither were used (e.g. `ANYCHANGE=$(pr_changed_files)` without any flag), right?

bin/pr_changed_files
iangmaia: I've updated the code on 9849d4db72e002c956736441e340b436bfb486e5 but some tests are failing (it's taking me some time to figure out why 😕)

bin/pr_changed_files
iangmaia: Good point (yes, it's implemented as an `OR`) 👍 some examples will definitely help.

.buildkite/pipeline.yml
```diff
@@ -92,3 +92,29 @@ steps:
         notify:
           - github_commit_status:
               context: "install_swiftpm_dependencies Tests: Xcode Workspace and Project - Explicit"
+
```
iangmaia, line 95: Hi @mokagio and @AliSoftware !

Gathering everything in a comment to make sure I'm not missing anything (and I'll continue tomorrow).

I've implemented a few updates, and the most relevant change was to add some tests.

These are my next steps in the PR:
- ~Fix the code to the array implementation (9849d4db72e002c956736441e340b436bfb486e5, see comments https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1943474741 / https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1943495921)~ (8d4cb986c7794f95968aae371498e93a7fe01970)
- ~Rename the arguments (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1943115056)~ (321ce30d2779be7b4251ad1c131c28de439f25af)
- ~Extract function if still applicable (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1942797360)~ (dbbe3d2079f1fa870094660136c97e037a36f689)
- ~Add a `git fetch origin` (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1942715413)~ (542703f47f2923a2959284d4064c0192c24ef395)
- ~Add more extensive usage examples (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1942664757)~ (80184b4b135e732e4900b44c6d6a9d27c8912746)
- ~Move pattern existence check https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148#discussion_r1943598793~ (fc315a6b687a5be8968d583a9d649942b99b7951)

.buildkite/pipeline.yml
```diff
@@ -92,3 +92,29 @@ steps:
...
+  - group: ":github: pr_changed_files Tests"
```
AliSoftware, line 96: 💡  What's a bit funny and meta is that once this new `pr_changed_files` helper lands in `trunk`, we could consider use it on itself to only run its unit tests if the script itself has changed, i.e.

```suggestion
  - group: ":github: pr_changed_files Tests"
    if: build.env('RUN_PR_CHANGED_FILES_TESTS')
```
After adding `RUN_PR_CHANGED_FILES_TESTS=$(pr_changed_files --includes "bin/pr_changed_files" "tests/pr_changed_files/*")` to the `shared-pipeline-vars` of this repo 😛

bin/pr_changed_files
iangmaia: Got it, I thought you meant before everything. Will update it!

.buildkite/pipeline.yml
iangmaia: YES! I was thinking exactly the same 😄

bin/pr_changed_files
```diff
@@ -0,0 +1,145 @@
...
+#   $ pr_changed_files --all-match "*.md" "docs/**"
+#   → "true"  if PR only changes docs/guide.md and README.md
+#   → "true"  if PR only changes docs/guide.md (not all patterns need to match)
+#   → "false" if PR changes docs/guide.md AND src/main.swift
+#
+#   # Check if any Swift files changed (to decide if we should run SwiftLint)
+#   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
+#   → "true"  if PR changes src/main.swift AND README.md
+#   → "true"  if PR changes .swiftlint.yml
+#   → "false" if PR only changes README.md
```
AliSoftware: Given the rest of the documentation always documents `--any-match` before `--all-match`, better keep the same order in the example to reduce the brainload of parsing those examples 😅 

```suggestion
#   # Check if any Swift files changed (to decide if we should run SwiftLint)
#   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
#   → "true"  if PR changes `.swiftlint.yml`
#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
#
#   # Check if only documentation files changed (to skip UI tests for example)
#   $ pr_changed_files --all-match "*.md" "docs/**"
#   → "true"  if PR only changes `docs/guide.md` and `README.md`
#   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
#   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
```

bin/pr_changed_files
```diff
@@ -0,0 +1,145 @@
...
+#   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
+#
+# Behavior:
+#   With no arguments:
+#     Returns "true" if the PR contains any changes, "false" otherwise
+#
+#   With --any-match:
+#     Returns "true" if ANY changed file matches ANY of the patterns
+#     Returns "false" if NONE of the changed files match ANY pattern
+#     Note: Will return "true" even if the PR includes other files not matching the patterns
```
AliSoftware:
```suggestion
#     Note: Will return "true" even if the PR includes other files not matching the patterns.
#           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
```

bin/pr_changed_files
```diff
@@ -0,0 +1,145 @@
...
+#
+#   With --all-match:
+#     Returns "true" if ALL changed files match AT LEAST ONE of the patterns
+#     Returns "false" if ANY changed file doesn't match ANY pattern
+#     Note: Will return "true" even if not all patterns are matched by the changed files
```
AliSoftware:
```suggestion
#     Note: Will return "true" even if not all patterns are matched by the changed files.
#           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
```

bin/pr_changed_files
```diff
@@ -0,0 +1,145 @@
...
+            shift
+            # Check if there are any patterns after the flag
+            if [[ "$#" -eq 0 || "$1" == "--"* ]]; then
+                echo "Error: must specify at least one file pattern" >&2
+                exit 1
+            fi
+            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
+                patterns+=("$1")
+                shift
+            done
```
AliSoftware: Opinionated: wouldn't it be simpler to put that test after the parsing of patterns, so that you don't have to repeat the logic about `!= "--"*` case but instead only test the size of the `patterns` array?

```suggestion
            while [[ "$#" -gt 0 && "$1" != "--"* ]]; do
                patterns+=("$1")
                shift
            done
            if [[ "${#patterns[@]}" -eq 0 ]]; then
                echo "Error: must specify at least one file pattern" >&2
                exit 1
            fi
```

bin/pr_changed_files
```diff
@@ -0,0 +1,145 @@
...
+
+# Get list of changed files as an array
+changed_files=()
+while IFS= read -r file; do
+    # Remove surrounding quotes if present and unescape special characters
+    file="${file#\"}"
+    file="${file%\"}"
+    file="${file//\\/}"
+    changed_files+=("$file")
+done < <(git --no-pager diff --name-only --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
```
AliSoftware: Instead of trying to post-process the quotes and escaped characters ourselves, I wonder if it wouldn't be easier to use the `-z` option of `git diff` and set the delimiter to NUL character in `read -r -d $'\0' file`?

.buildkite/pipeline.yml
iangmaia: @AliSoftware @mokagio 
I've updated all the items above (I've added the commit in which each change was implemented).

Now I've been trying to use `a8-ci-toolkit-buildkite-plugin` by pointing it directly to this branch in another repo, to make sure the general solution works as expected. See https://github.com/woocommerce/woocommerce-ios/pull/15086.

I'm facing an issue (that is a bit obvious, in retrospect) where the `pr_changed_files` command we're trying to run on `.buildkite/shared-pipeline-vars` isn't installed yet, as the version is defined in the `.buildkite/shared-pipeline-vars` itself ([see the reported error](https://buildkite.com/automattic/woocommerce-ios/builds/28036#0194dc6a-fca8-46fb-9612-8f03a3ed7fab/139-147)). 

One simple alternative I can think of: call `buildkite-agent plugin install $CI_TOOLKIT` directly in `shared-pipeline-vars`, but that would of course slow down the upload a bit 🤔 
What do you think? Any other ideas?

AliSoftware: > I'm facing an issue (that is a bit obvious, in retrospect) where the `pr_changed_files` command we're trying to run on `.buildkite/shared-pipeline-vars` isn't installed yet, as the version is defined in the `.buildkite/shared-pipeline-vars` itself

Oh shoot, didn't think about that… right…

> One simple alternative I can think of: call buildkite-agent plugin install $CI_TOOLKIT directly in shared-pipeline-vars, but that would of course slow down the upload a bit 🤔

Yeah. Tricky.

1️⃣  `buildkite-agent plugin install`

I guess it depends how much time it will add.

It will only add that time for pipelines that do want to use `pr_changed_files` and thus need to install the plugin in the uploader agent—without affecting other pipelines that don't use `pr_changed_files` in their `shared-pipeline-vars` and won't call `buildkite-agent plugin install`. And given the goal of this trick is to be able to skip e.g. UI Test jobs on builds where they're not needed, if the installation of the plugin takes only 1–2 seconds, then the time lost will be balanced tenfold with the benefit to avoid the build time of many skippable jobs…

2️⃣  Another option would be to download just that binary ourselves (`curl https://raw.githubusercontent.com/Automattic/a8c-ci-toolkit-buildkite-plugin/refs/heads/trunk/bin/pr_changed_files` + some `chmod +x`…) but at that point that would kinda be reinventing the wheel compared to what `buildkite-agent plugin install` does (sure it'd reduce the download to only that specific binary instead of installing the whole plugin, but not sure there'd be much of a difference, those `bin/` scripts in our plugin are not that big)

3️⃣  Bootstrap script on the uploader agent

The last option I could think of would be to install the script directly into the agents at the time we provision and deploy them.
 - Right now [our `upload` agents use the default Buildkite stack as our uploader agents as is](https://github.com/Automattic/buildkite-ci/blob/trunk/src/aws/stacks/clustered-upload.tf)
 - So that would require either create a custom AMI on top of that to install custom scripts into it on top of the stock Buildkite agent stack…
 - or use the ability provided by https://github.com/buildkite/elastic-ci-stack-for-aws to [use a bootstrap script](https://buildkite.com/docs/agent/v3/elastic-ci-aws/managing-elastic-ci-stack#customizing-instances-with-a-bootstrap-script) to install the `a8c-ci-toolkit` plugin inside the agent at the time they are booted

At that point that last option 3️⃣  feels a bit too convoluted just to avoid the simpler call to `buildkite-agent plugin install`… but it's still a possibility to be aware of…

AliSoftware (commented):
Dammit I just realized that I haven't submitted my pending comments of my last review… so posting them even if you asked for review only after I started this one 😅

.buildkite/pipeline.yml
iangmaia: Turns out something like `buildkite-agent plugin install` was how I remembered we did to install it in the agents, but in fact we just download them during the agent provisioning 😓
I don't think there's an easy command to dynamically install plugins, only use them in pipeline jobs.

Which makes me think of two other alternatives:
4️⃣ 
```yml
  - label: "Check Changes"
    key: check-changes
    command: |
      SKIP_UI_TESTS=$(pr_changed_files --all-match "*.md")
      buildkite-agent meta-data set "SKIP_UI_TESTS" "$SKIP_UI_TESTS"
    plugins: [$CI_TOOLKIT]
```
And make the more expensive jobs depend on it 🤔 (assuming I can check then `SKIP_UI_TESTS` in the pipeline, but I haven't tested it).

And 5️⃣ 
Just use it on a case by case basis either in the script itself (e.g. `.buildkite/commands/run-ui-tests.sh` or equivalent) or in the `command` in yml attribute.

---

3️⃣ even if it is a bit more complex to build, it's easier to use in the end. Alternatively, 4️⃣ or 5️⃣  could be a simple starting point to try the command.

AliSoftware: 4️⃣  [wouldn't work](https://buildkite.com/docs/pipelines/configure/conditionals#variable-and-syntax-reference-variables), at least not as pitched of using `SKIP_UI_TESTS` metadata in the `if:` attribute of UI Tests step.

image

But alternatives of that include:
 - 4️⃣ 🅰️  either use `buildkite-agent pipeline upload` to upload the UITests steps conditionally (after having extracted their steps in a dedicated `*.yml` file)
 - 4️⃣ 🅱️  or keep `buildkite-agent meta-data set` and let the UITests steps start a job in all cases (instead of the creation of the job itself being skipped by `if:` attribute), and check the metadata value at the start of the job's script. Which would mean we'd still have the UITests job request an agent and start the build, but it'd exit early.

🅰️  Would have the benefit of preventing to create the UITests job to begin with, reducing the need to assign a CI agent for nothing. But that also means it wouldn't generate the GitHub CI check notification, and so that the "Run UI Tests" check would never be marked as done in the PR. If that check is marked as required for the PR to be merged, that is going to be an issue.

🅱️  Has the drawback of requesting a CI agent even if it will exit early and have nothing to build, but has the benefit of ensuring the CI check notification will be sent to the PR if that check is required.

---

But if we go with 4️⃣ 🅱️  at that point we might as well just run the `pr_changed_files` directly in the UI Tests `command` (instead of running it in a separate step, set it as metadata, and read the metadata in the UI Tests `command`…), aka your option 5️⃣

iangmaia: Test with 4️⃣ 🅱️ /  5️⃣: https://github.com/woocommerce/woocommerce-ios/pull/15086 / https://buildkite.com/automattic/woocommerce-ios/builds/28092#0194e1ba-9cf5-4ccc-8cfd-9bd73e0b1494/286-288.
It's pretty straightforward, but another downside as you mention is the fact that it still uses some CI resources.

In any case, I believe we can move forward with this PR, as the final solution on how this is going to be used should be independent of the implementation.

bin/pr_changed_files
```diff
@@ -0,0 +1,147 @@
...
+#   $ pr_changed_files
+#
+#   # Check if any Swift files changed (to decide if we should run SwiftLint)
+#   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
+#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → "true"  if PR changes `.swiftlint.yml`
+#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
+#
+#   # Check if only documentation files changed (to skip UI tests for example)
+#   $ pr_changed_files --all-match "*.md" "docs/**"
```
AliSoftware:
```suggestion
#   $ pr_changed_files --all-match "*.md" "docs/*"
```

bin/pr_changed_files
AliSoftware: Up on this question/suggestion in case you missed it.

bin/pr_changed_files
```diff
@@ -0,0 +1,147 @@
...
+    exit 0
+fi
+
+# Returns 0 if the file matches any of the patterns, 1 otherwise
+file_matches_any_pattern() {
+    local file="$1"
+    shift
+    local patterns=("$@")
+
+    for pattern in "${patterns[@]}"; do
```
AliSoftware: Nit: given `patterns` is not used anywhere other than on the `for` line 120, we might as well skip the bash array creation?

```suggestion
    for pattern in "$@"; do
```

bin/pr_changed_files
```diff
@@ -0,0 +1,147 @@
...
+elif [[ "$mode" == "any-match" ]]; then
+    # Check if any changed file matches any pattern
+    for file in "${changed_files[@]}"; do
+        if file_matches_any_pattern "$file" "${patterns[@]}"; then
+            echo "true"
+            exit 0
+        fi
+    done
+    echo "false"
+fi
```
AliSoftware, line 141: Nitpick, similar to the comment I made some days ago about the order we list the modes for consistency: might be nice to put the implementation of `any-match` first and `all-match` next, since that's the order in which those are listed in the rest of the script.

```suggestion
if [[ "$mode" == "any-match" ]]; then
    # Check if at least one changed file matches any of the patterns
    for file in "${changed_files[@]}"; do
        if file_matches_any_pattern "$file" "${patterns[@]}"; then
            echo "true"
            exit 0
        fi
    done
    echo "false"
elif [[ "$mode" == "all-match" ]]; then
    # Check if all changed files match at least one of the patterns
    for file in "${changed_files[@]}"; do
        if ! file_matches_any_pattern "$file" "${patterns[@]}"; then
            echo "false"
            exit 0
        fi
    done
    echo "true"
else
    echo "Internal error: unexpected mode $mode"
    exit 255 # Should never happen, but just for belt-and-braces
fi
```

bin/pr_changed_files
```diff
@@ -0,0 +1,147 @@
...
+
+mode=""
+patterns=()
+
+# Define error message for mutually exclusive options
+EXCLUSIVE_OPTIONS_ERROR="Error: either specify --all-match or --any-match; cannot specify both"
+
+while [[ "$#" -gt 0 ]]; do
+    case $1 in
+        --all-match | --any-match)
```
AliSoftware, line 66: Another one for `$mode` order consistency throughout the script
```suggestion
        --any-match | --all-match)
```

bin/pr_changed_files
AliSoftware: After some testing in my `/bin/bash` (version 3), I think this should work, and would things not only simpler, but also safer (as we won't need to do some manual post-processing of the file names to remove quotes or escaped characters ourselves):

```suggestion
while IFS= read -r -d '' file; do
  changed_files+=("$file")
done < <(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
```

Explanations:
 - `-d ''` asks `read` to read until it encounters a `NUL` character (`''` is an empty-and-NUL-terminated bash string; you can also use `$'\0'` if you wanted to be more explicit, but using `-d ''` to use a `NUL` delimiter is exactly equivalent and simpler + more commonly used).
 - And since `git … diff --name-only -z …` will print the file names raw (without added quotes nor escaping) and each terminated by a `NUL` char, that will do exactly what we want

In fact this `while IFS= read -r -d '' …` is a typical usage found all over the internet when coupled with a command that prints entries separated by a NUL—typically `find … -print0` in most examples I've found, but `git diff --name-only -z` should be the same.

bin/pr_changed_files
mokagio: > might be nice to put the implementation of `any-match` first and `all-match` next, since that's the order in which those are listed in the rest of the script.

Nitpick on the nitpick. 

I haven't followed the past couple of iterations, so I might be missing something here. I operate under the assumption the order is arbitrary, after all the two modes are mutually exclusive. With that in mind, I would suggest following the alphabetical order, `all-match` before `any-match` 🤷‍♂️

bin/pr_changed_files
iangmaia: I've tested it here and it fails in one of the tests, but the rest succeeds 🤔

```bash
# [Test] All changes in docs with explicit patterns including spaces and special chars
result=$(pr_changed_files --all-match "docs/read me.md" "docs/guide with spaces.md" "docs/special\!\@\#\$chars.md")
assert_output "true" "$result" "Should return true when all changes match patterns with spaces and special chars"
```

bin/pr_changed_files
iangmaia: After @mokagio 's [suggestion](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/148/files#r1948307210) (use the alphabetical order) I've updated the code on 13d4dfbafdfb0d9e0d0f8ec39f240b5173a79837.

bin/pr_changed_files
iangmaia: Thanks -- good idea to keep the order consistent, also following the alphabetical order. Updated on 13d4dfbafdfb0d9e0d0f8ec39f240b5173a79837.

bin/pr_changed_files
AliSoftware: I wonder if that failure you have here is not because of the script itself but the invocation and the shell interpreting the escapes before passing the arguments to the script.

When using `"docs/special\!\@\#\$chars.md"` with double quotes as a parameter in the bash shell invocation, the shell interprets the evalid scapes in the quoted string before passing it as parameters to the `pr_changed_files` script:
```
$ echo "docs/special\!\@\#\$chars.md"
docs/special!\@\#$chars.md
```

Similarly, when in your test script you create the test file using:
```
echo "doc3" > "docs/special\!\@\#\$chars.md"
```
The actual name of the file being created is `special!\@\#$chars.md` (you can see that either with a `ls "folder with spaces"` or a `find . -print0`). In particular, the backslash before the `!` and `$` were interpreted by the shell as escaping those characters—because indeed `!` and `$` are special characters in shell prompt—but all the other characters you prefixed with a backslash, like `\@` and `\#`, are not valid escape sequences—as `@` and `#` are not special characters in this context in shell, so the backslash was kept for those ones.

It'd be interesting to debug-print the content of `changed_files` array after that parsing (e.g. using `declare -p changed_files` to dump the declaration), as well as printing the `$patterns` that were parsed from the command line invocation and passed to the script, and compare them.

---

My bet would be that the issue might not necessarily be with our `read …` commands there and how we parse the list of files and patterns into bash arrays when they have special characters, but instead maybe an issue with how we use them later in the script, especially the behavior when using it in the `[[ "$file" == $pattern ]]` expression.

Indeed, once the shell has parsed your 3rd parameter `"docs/special\!\@\#\$chars.md"` in your call to `pr_changed_files` invocation, the string the shell passes to your script (as `$4`) has been processed by the shell and ends up being the string `docs/special!\@\#$chars.md`. Then when you use that `$pattern` in the `[[ "$file" == $pattern ]]` expression, it will interpret the now-unescaped `!` and `$` characters of the pattern string in that test.

It's a tricky balance to find between having the shell interpret special characters when you pass a string with special characters to your command invocation, but not having it interpret special characters you don't want to when you use it in the `[[ … ]]` condition as an unquoted `$pattern`…

```
file="docs/special\!\@\#\$chars.md"
# If we don't quote the $pattern we have to escape all the special we don't want to be interpreted as such
[[ "$file" == docs/special\!\\@\\#\$chars.md ]] # true

# If we quote the pattern, we don't have this problem.
# But that also means that if the $pattern contains a `*` it will interpreted as a literal character, not as a wildcard
[[ "$file" == 'docs/special!\@\#$chars.md' ]] # true
```

In a way, you passing `pr_changed_files "docs/special\!\@\#\$chars.md"` (or `pr_changed_files 'docs/special!\@\#$chars.md'`, with single quotes but no backslash in characters that need escaping, which should pass the same string to the script ultimately) does exactly what you ask it to: compare the file name with the pattern `'docs/special!\@\#$chars.md'`. Except that it happens that in this pattern, `!` and `$` are special character just like a `*` wildcard would be, so it interprets it when evaluating the `==` with the pattern, just like you'd want it to interpret a `*` if there was one, I think.

I also think that if you used single-quotes for the paramater (while keeping the current escapes), i.e. `pr_changed_files 'docs/special\!\@\#\$chars.md`, it would work and make the test pass, because then the test would be using `[[ "$file" == docs/special\!\@\#\$chars.md ]]` (un-escaping the `\!` and `\$` only at that stage instead of already being unescaped at the shell-interpreter stage) and would do what you want?

---

PS: I wonder if it would be worth checking amongst the various `shopt` available to bash 3 if there is one of the options that would tell pattern globbing to only consider `*` (and maybe `{…}` and `?` too) but not `!` and `$` and other chars… or alternatively a different way to do pattern globbing that would be safer than using the pattern unquoted and having the disable `SC2053`? Though I'm not sure there's a perfect solution for what we'd want here, that would only interpret some characters like `*` but not `!` during the call to `[[`… 🤔 

But ultimately I think that's probably not worth it (who uses `!` and `$` in file names anyway? 😅 ), + the fact that those `!` and `$` are interpreted as they are when used in unquoted `$pattern` in `[[ … ]]` test is actually the expected behavior after all… I think it's probably not worth spending more time on this, and instead update your test use single-quotes when passing the pattern to the `pr_changed_files` call would be enough to move on…

iangmaia: > It'd be interesting to debug-print the content of changed_files array after that parsing 

Good idea!
The output was `changed_files='([0]="docs/guide with spaces.md" [1]="docs/read me.md" [2]="docs/special\\!\\@\\#\$chars.md")'`.

> I also think that if you used single-quotes for the paramater (while keeping the current escapes), i.e. pr_changed_files 'docs/special\!\@\#\$chars.md, it would work and make the test pass

That itself didn't make it pass, though using `'docs/special\\!\\@\\#\$chars.md'` (following what we got for the output of `changed_files`).

That's how I updated it finally at 4cc76b299fa894747fba4b15f66e319bfd267f43 and everything is passing now. In the end this was moving the burden to escaping on the caller 🤷 though it probably makes sense? Anyway, I'm also feeling like it's not worth spending more time on this.

AliSoftware: >  In the end this was moving the burden to escaping on the caller 🤷 though it probably makes sense?

Yep makes sense to me because the caller might want to escape some characters of the pattern (like `!` and `$`)… but might not want to escape `*`. So if we want to pass a pattern that would match the literal string 'ab*cd' followed by arbitrary characters, followed by literal string `ef`, it makes sense that it'd be up to the called to escape `'ab\*cd*ef'` to escape the first `*` so it's matched as a literal, but let the second one be interpreted as a wildcard, right?

> Anyway, I'm also feeling like it's not worth spending more time on this.
👍

AliSoftware: Ah actually while looking around and playing with your test I think the test is not exactly doing what we think it's doing.

In particular, in zsh—in which I think I accidentally used while testing some of the above instead of testing it inside a `bash` 3.x shell—`!` has the special meaning of recalling a shell history entry… but inside a `/bin/bash` script it does not seem to.

Which means for example that your `echo "doc3" > "docs/special\!\@\#\$chars.md"` from line 24 for `test_all_match_patterns.sh` creates a file whose name contains literal baskslahes before each of `!`, `@` and `#` (so, including the `!`… that I previously said was a special character in this context and thus `\!` being a valid escape… while in `bash` it's not). Only the `$` would be interpreted if not escaped, and `\$` is the only valid escape.

This results in a file named `special\!\@\#$chars.md`, with backslashes kept everywhere except in front of the `$`. So no wonder that when when you then make your test call with `pr_changed_files --all-match` you had to pass `'docs/special\\!\\@\\#\$chars.md'`, with the double-backslashes, which would pass exactly that string literally as a pattern… and once used unquoted in `[[ "$file" == docs/special\\!\\@\\#\$chars.md ]]` would only match the file because the double-backslashes in the unescaped pattern indeed would match the literal backslashes in the file name.

---

In other words:
 - This works exactly as expected…
 - …but the way the `test_all_match_patterns` is written makes it seems like you wanted to create files whose names didn't contain those literal backslashes—but instead that you used those backslashes to escape special characters… while they weren't—while in reality those backslashes ended up being literal characters in the names of the created files

It would probably be clearer when reading the tests to use simple quotes when creating your test files instead, e.g.:
```bash
# File name containing a literal backslash, then all literal !, @, * and # and $ characters
# Note: we're using single quotes, so none of those characters are interpreted as special character in that call
echo "doc3" > 'docs/special\!@*#$chars.md'
```
Then your invocation to `pr_changed_files` would still need to escape the special characters that are the backslash and the `*`—given the pattern would be used unquoted while you want all those characters to be matched literally and not as pattern wildcards:
```bash
result=$(pr_changed_files --all-match 'docs/read me.md' 'docs/guide with spaces.md' 'docs/special\\!@\*#$chars.md')
```

And in addition to that, it would be nice to add a test case that actually makes use of the `*` as a wildcard in the pattern, e.g.:
```bash
result=$(pr_changed_files --all-match 'docs/read me.md' 'docs/guide with spaces.md' 'docs/special\\!*.md')
```

AliSoftware: @iangmaia I took the liberty to push a commit 40cac254efda7e175b42f9ad1e0fb66d550236e6 with my suggested changes from the comment above, as I figured it'd be clearer for you to see the diff done than figuring out what I was trying to mean 😅

AliSoftware: @iangmaia  Also just pushed ee11f58 to do similar changes (of using single-quotes everywhere to make it clearer which special characters we were playing with in those tests, without having to escape them every time they were used in double-quote contexts in the test script) to all other test files.

iangmaia: > Then your invocation to pr_changed_files would still need to escape the special characters that are the backslash

Aah, that's what I missed! In one of my attempts today I did exactly that change to use `echo "doc3" > 'docs/special\!@*#$chars.md'` but it still didn't work when only using single quotes in the arguments, as it needed to adapt the escaping as well.

So much escaping-chaos  😅 thanks for the commit; the intention is also clearer now.

AliSoftware (approved):
:shipit: 🍾

bin/pr_changed_files
AliSoftware: > So much escaping-chaos 😅

Yep, hence why we should have switched to use single-quotes from the beginning to avoid the need for all that escaping done by the shell invocations in the test files themselves and avoid us all those headaches!

In the end the issue was not in the implementation of the `pr_changed_files` script doing incorrect escaping or whatnot… but in the invocation of those in your test scripts after all 😅

PR #149, opened 2025-02-11
Add libc version to Gems cache key

mokagio:
Internal ref p1739222874153819-slack-C02KLTL3MKM

The comments in `install_gems` should explain the rationale for this PR.

You can see it in action in [this Day One Android demo build](https://buildkite.com/automattic/dayone-android/builds/19680)

Notice how the two steps use different gem caches because the `Gemfile.lock` includes `nokogiri` and the `android` and `tumblr-metal` queue have different `libc` versions

image

image

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

tests/validate_get_glibc_version_format
```diff
@@ -0,0 +1,30 @@
...
+    os_exp="linux"
+    ;;
+  darwin*)
+    os_exp="macos"
+    ;;
+  msys*|cygwin*|Windows_NT*)
+    os_exp="win"
+    ;;
+  *)
+    echo "❌ Unknown OS type: $OSTYPE"
```
mokagio:
```suggestion
    echo "Unknown OS type: $OSTYPE"
```

.buildkite/pipeline.yml
```diff
@@ -37,6 +37,27 @@ steps:
...
+    steps:
+      - label: ":c: Test get_libc_version on {{matrix}}"
+        command: ./tests/validate_get_libc_version_format
+        agents:
+          queue: "{{matrix}}"
+        matrix:
+          - default
+          - android
+          - tumblr-metal
+          - mac
```
mokagio:
```suggestion
          - android
          - default
          - mac
          - tumblr-metal
```

bin/get_libc_version
```diff
@@ -0,0 +1,40 @@
+#!/bin/bash -eu
+
+VERBOSE=${1:-false}
```
mokagio, line 8:
```suggestion
#
# Returns the system's libc version, or a close proxy of it.
# Useful for commands like install_gems which use a number of system information to compute the key for the cache they lookup.
#

VERBOSE=${1:-false}
```

tests/validate_get_libc_version_format
```diff
@@ -0,0 +1,30 @@
...
+    os_exp="linux"
+    ;;
+  darwin*)
+    os_exp="macos"
+    ;;
+  msys*|cygwin*|Windows_NT*)
+    os_exp="win"
+    ;;
+  *)
+    echo "❌ Unknown OS type: $OSTYPE"
```
mokagio:
```suggestion
    echo "Unknown OS type: $OSTYPE"
```

bin/install_gems
```diff
@@ -4,7 +4,16 @@ PLATFORM=$(uname -s)
 ARCHITECTURE=$(uname -m)
 RUBY_VERSION=$(rbenv version-name 2>/dev/null || cat .ruby-version)
 GEMFILE_HASH=$(hash_file Gemfile.lock)
-CACHEKEY="$BUILDKITE_PIPELINE_SLUG-$PLATFORM-$ARCHITECTURE-ruby$RUBY_VERSION-$GEMFILE_HASH"
+# Many of our projects require nokogiri, which may result in issues if the
+# cache is built on an environment with a different libc version than the one
+# where it is downloaded. To avoid issues, add the libc version to the cache
+# key if the Gemfile.lock includes nokogiri.
+if grep -q "nokogiri" Gemfile.lock; then
```
iangmaia, line 11: I wonder if we should worry about the additional "cache miss" that this is going to introduce. 
For example, if we think this is worth it, maybe we can implement something in `restore_cache` so that:
- A system with libc `2.34` can use cache from `2.17`
- A system with libc `2.17` cannot use cache from `2.34`

And I'm thinking under the assumption that those would be backward compatible, as we've seen (which I think should be the case as long as they have the same major version 🤔)

Considering the family of agents we have and they're all using the same version, maybe this won't be a big issue. Perhaps we'd have a few unwanted "cache miss" (e.g. a cache built using `tumblr-metal` that won't be used by `default` or `android`, while it could be used) but not such a big deal 🤔

AliSoftware: Yeah, that's a good point… but tbh I'm not sure it's worth trying to implement some custom logic to compare the versions. Including because:
 - The logic for it can be tricky and we risk getting that version comparison logic wrong…
 - Or even if we get it right, maybe it might happen that a future minor version of libc would still not be backwards-compatible despite not being a major bump (i.e. if libc don't fully respect semver as we hoped it would)
 - But also, as shown in the implementation of `get_libc_version` helper, we'll not always have an actual version of libc per se, but sometimes will have to fallback to a version of the OS to differentiate… in which case such semver-aware version comparison won't be the right fit (e.g. I can totally imagine macOS not waiting for a major version bump of the OS itself to upgrade libc to a next major version one day…)
 - And finally, even if we implemented such logic, there's still a risk (and actually quite a likely case) that the cache miss would _first_ happen on an `android` or `default` agent that use a newer version of `libc` (e.g. `2.34`), because it'd be triggered by a PR updating the `Gemfile.lock` and the CI checks running on that PR mostly running on `android`/`default` and not `tumblr-metal`… and it's only when we'll _later_ have one of our `.buildkite/release-pipelines/*.yml` (which would typically need `tumblr-metal` trusted agents to be able to `git push`) that we'll need a cache that is compatible with an older version of `libc` (e.g. `2.17`), in which case we'll still have a cache miss and a new cache being built (even if we had that version logic) for that older libc version. So overall I'm not even sure we'd ever encounter a case where the cache would be rebuild with `libc 2.17` first and then be able to be recycled for agents that have `2.34` anyway…

So overall I think it's not only simpler but also less risky to just not try to implement such version backward-compat-check logic, even if that means that we might have more cache miss (and more `.tar.gz` cache files created in our S3 bucket).

AliSoftware, line 11: Clearly `nokogiri` is the main and most common culprit there. But I wonder if there might be cases of some other gems than `nokogiri` that would create the same situation. Especially I'm thinking _any_ gem that includes "native extensions" might be a candidate.

That being said, I don't know of a way to know _in advance_ (e.g. just from checking an existing `Gemfile.lock` _before_ actually installing any gem) which gems _will_ require native extensions when you will install them. 

Which is really the test we'd want to do (only use the libc version in the cache key if _any_ of the gems listed in `Gemfile.lock` use native extensions and will require to compile something to be installed)… but not sure is something we _can_ technically do… 🤔

AliSoftware: While I still think that this approach of including the `libc` version in the cache key is the best solution (and also quick to implement all while being confident it'll be reliable), maybe in a future iteration we could check if the `bundle cache` command can be helpful to improve our implementation of `install_gems` overall.

I'm not convinced that `bundle cache` would solve the libc issue, because from what I understand from `bundle cache --help`, it mostly differentiate the cache entries per `platform` (those `<arch>-darwin-<n>` triples) while I'm not sure if the `android` vs `tumblr-metal` CI agent would have different triples (but maybe same triple but not same version of libc installed still?)… So maybe even if we started using `bundle cache` we'd still have the issue of `GLIBC` compatibilities on different agents… But still worth noting as an idea to experiment in the future for potential improvements and testing?

AliSoftware (approved):
Given this is happening on more projects, I'd say let's merge this and start adopting it, and we can always iterate on potential improvements later

bin/install_gems
iangmaia: > So overall I think it's not only simpler but also less risky to just not try to implement such version backward-compat-check logic, even if that means that we might have more cache miss

Agreed! 👍

PR #151, opened 2025-02-12
Enhance pr_changed_files with exit codes and string output

iangmaia:
Follow up to the discussion started on https://github.com/woocommerce/woocommerce-ios/pull/15086#discussion_r1953247300.

This PR adds both exit codes (default mode) and `"true"` `"false"` output (using `--stdout` flag) to the `pr_changed_files` command, updating documentation and tests accordingly.

---

### DIFF
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -12,39 +12,53 @@
 #   1: Condition is not met (no files changed / patterns not matched)
 # 255: Error in command invocation (bad parameters, not in PR context, etc.)
 #
+# Output:
+#   In addition to exit codes, the script also outputs "true" or "false" to stdout
+#   when the condition is met or not met respectively.
+#
+# Typical usage patterns:
+#   # To assign the result to a variable — use `|| true` if your shell has `set -e` enabled
+#   $ DOC_UPDATED=$(pr_changed_files --any-match docs/* *.md || true)
+#
+#   # To use in an `if` condition — use `>/dev/null` to avoid the true/false output
+#   $ if pr_changed_files --any-match docs/* *.md >/dev/null; then
+#       echo "Documentation was updated"
+#     fi
+#
 # Behavior:
 #   With no arguments:
-#     Returns 0 if the PR contains any changes, 1 otherwise
+#     Returns exit code 0 (outputs "true") if the PR contains any changes
+#     Returns exit code 1 (outputs "false") if no changes
 #
 #   With --all-match:
-#     Returns 0 if ALL changed files match AT LEAST ONE of the patterns
-#     Returns 1 if ANY changed file doesn't match ANY pattern
+#     Returns exit code 0 (outputs "true") if ALL changed files match AT LEAST ONE of the patterns
+#     Returns exit code 1 (outputs "false") if ANY changed file doesn't match ANY pattern
 #     Note: Will return 0 even if not all patterns are matched by the changed files.
 #           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
 #
 #   With --any-match:
-#     Returns 0 if ANY changed file matches ANY of the patterns
-#     Returns 1 if NONE of the changed files match ANY pattern
+#     Returns exit code 0 (outputs "true") if ANY changed file matches ANY of the patterns
+#     Returns exit code 1 (outputs "false") if NONE of the changed files match ANY pattern
 #     Note: Will return 0 even if the PR includes other files not matching the patterns.
 #           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
 #
 # Examples with expected outputs:
 #   # Check if any files changed
 #   $ pr_changed_files
-#   → exit code 0 if PR has changes
-#   → exit code 1 if PR has no changes
+#   → exit code 0 (outputs "true") if PR has changes
+#   → exit code 1 (outputs "false") if PR has no changes
 #
 #   # Check if only documentation files changed (to skip UI tests for example)
 #   $ pr_changed_files --all-match "*.md" "docs/*"
-#   → exit code 0 if PR only changes `docs/guide.md` and `README.md`
-#   → exit code 0 if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
-#   → exit code 1 if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
+#   → exit code 0 (outputs "true") if PR only changes `docs/guide.md` and `README.md`
+#   → exit code 0 (outputs "true") if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
+#   → exit code 1 (outputs "false") if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
 #
 #   # Check if any Swift files changed (to decide if we should run SwiftLint)
 #   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
-#   → exit code 0 if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
-#   → exit code 0 if PR changes `.swiftlint.yml`
-#   → exit code 1 if PR only changes `README.md` (none of files match any of the patterns)
+#   → exit code 0 (outputs "true") if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → exit code 0 (outputs "true") if PR changes `.swiftlint.yml`
+#   → exit code 1 (outputs "false") if PR only changes `README.md` (none of files match any of the patterns)
 
 if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
     echo "Error: this tool can only be called from a Buildkite PR job" >&2
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -6,6 +6,7 @@
 #   pr_changed_files                                  # Check if any files changed
 #   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
 #   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
+#   pr_changed_files [--stdout] ...                   # Add --stdout to output "true"/"false" to stdout
 #
 # Exit codes:
 #   0: Condition is met (files changed / patterns matched)
@@ -13,17 +14,17 @@
 # 255: Error in command invocation (bad parameters, not in PR context, etc.)
 #
 # Output:
-#   In addition to exit codes, the script also outputs "true" or "false" to stdout
-#   when the condition is met or not met respectively.
+#   If the flag --stdout is used, the script will output "true" or "false" to stdout
+#   when the condition is met or not met respectively, and will return exit code 0.
 #
 # Typical usage patterns:
-#   # To assign the result to a variable — use `|| true` if your shell has `set -e` enabled
-#   $ DOC_UPDATED=$(pr_changed_files --any-match docs/* *.md || true)
-#
-#   # To use in an `if` condition — use `>/dev/null` to avoid the true/false output
-#   $ if pr_changed_files --any-match docs/* *.md >/dev/null; then
+#   # Using exit codes
+#   $ if pr_changed_files --any-match docs/* *.md; then
 #       echo "Documentation was updated"
 #     fi
+#
+#   # Using stdout output
+#   $ DOCS_UPDATED=$(pr_changed_files --stdout --any-match docs/* *.md)
 #
 # Behavior:
 #   With no arguments:
@@ -73,12 +74,17 @@ git fetch origin "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" >/dev/null 2>&1 || {
 
 mode=""
 patterns=()
+use_stdout="false"
 
 # Define error message for mutually exclusive options
 EXCLUSIVE_OPTIONS_ERROR="Error: either specify --all-match or --any-match; cannot specify both"
 
 while [[ "$#" -gt 0 ]]; do
     case $1 in
+        --stdout)
+            use_stdout="true"
+            shift
+            ;;
         --all-match | --any-match)
             if [[ -n "$mode" ]]; then
                 echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
@@ -113,14 +119,26 @@ while IFS= read -r -d '' file; do
   changed_files+=("$file")
 done < <(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
 
+exit_result() {
+    local result=$1
+    if [[ "$use_stdout" == "true" ]]; then
+        if [[ "$result" -eq 0 ]]; then
+            echo "true"
+        else
+            echo "false"
+        fi
+        exit 0
+    fi
+
+    exit $result
+}
+
 if [[ -z "$mode" ]]; then
     # No arguments = any change
     if [[ ${#changed_files[@]} -gt 0 ]]; then
-        echo "true"
-        exit 0
+        exit_result 0
     else
-        echo "false"
-        exit 1
+        exit_result 1
     fi
 fi
 
@@ -141,20 +159,16 @@ if [[ "$mode" == "all-match" ]]; then
     # Check if all changed files match at least one pattern
     for file in "${changed_files[@]}"; do
         if ! file_matches_any_pattern "$file" "${patterns[@]}"; then
-            echo "false"
-            exit 1
+            exit_result 1
         fi
     done
-    echo "true"
-    exit 0
+    exit_result 0
 elif [[ "$mode" == "any-match" ]]; then
     # Check if any changed file matches any pattern
     for file in "${changed_files[@]}"; do
         if file_matches_any_pattern "$file" "${patterns[@]}"; then
-            echo "true"
-            exit 0
+            exit_result 0
         fi
     done
-    echo "false"
-    exit 1
+    exit_result 1
 fi
--- a/bin/pr_changed_files
+++ b/bin/pr_changed_files
@@ -7,52 +7,54 @@
 #   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
 #   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
 #
+# Exit codes:
+#   0: Condition is met (files changed / patterns matched)
+#   1: Condition is not met (no files changed / patterns not matched)
+# 255: Error in command invocation (bad parameters, not in PR context, etc.)
+#
 # Behavior:
 #   With no arguments:
-#     Returns "true" if the PR contains any changes, "false" otherwise
+#     Returns 0 if the PR contains any changes, 1 otherwise
 #
 #   With --all-match:
-#     Returns "true" if ALL changed files match AT LEAST ONE of the patterns
-#     Returns "false" if ANY changed file doesn't match ANY pattern
-#     Note: Will return "true" even if not all patterns are matched by the changed files.
+#     Returns 0 if ALL changed files match AT LEAST ONE of the patterns
+#     Returns 1 if ANY changed file doesn't match ANY pattern
+#     Note: Will return 0 even if not all patterns are matched by the changed files.
 #           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
 #
 #   With --any-match:
-#     Returns "true" if ANY changed file matches ANY of the patterns
-#     Returns "false" if NONE of the changed files match ANY pattern
-#     Note: Will return "true" even if the PR includes other files not matching the patterns.
+#     Returns 0 if ANY changed file matches ANY of the patterns
+#     Returns 1 if NONE of the changed files match ANY pattern
+#     Note: Will return 0 even if the PR includes other files not matching the patterns.
 #           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
 #
 # Examples with expected outputs:
-#   # Check if any files changed, returning "true" if PR has changes, "false" otherwise
+#   # Check if any files changed
 #   $ pr_changed_files
+#   → exit code 0 if PR has changes
+#   → exit code 1 if PR has no changes
 #
 #   # Check if only documentation files changed (to skip UI tests for example)
 #   $ pr_changed_files --all-match "*.md" "docs/*"
-#   → "true"  if PR only changes `docs/guide.md` and `README.md`
-#   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
-#   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
+#   → exit code 0 if PR only changes `docs/guide.md` and `README.md`
+#   → exit code 0 if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
+#   → exit code 1 if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
 #
 #   # Check if any Swift files changed (to decide if we should run SwiftLint)
 #   $ pr_changed_files --any-match "*.swift" ".swiftlint.yml"
-#   → "true"  if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
-#   → "true"  if PR changes `.swiftlint.yml`
-#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
-#
-# Returns:
-#   Prints "true" if the condition is met, "false" otherwise
-#   Exits with code 0 regardless of if the condition was met or not, to allow easy variable assignment.
-#   Only exits with non-zero if the command invocation itself was incorrect (called outside of a PR context, incorrect arguments…)
+#   → exit code 0 if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → exit code 0 if PR changes `.swiftlint.yml`
+#   → exit code 1 if PR only changes `README.md` (none of files match any of the patterns)
 
 if [[ ! "${BUILDKITE_PULL_REQUEST:-invalid}" =~ ^[0-9]+$ ]]; then
     echo "Error: this tool can only be called from a Buildkite PR job" >&2
-    exit 1
+    exit 255
 fi
 
 # Ensure we have the base branch locally
 git fetch origin "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" >/dev/null 2>&1 || {
     echo "Error: failed to fetch base branch '$BUILDKITE_PULL_REQUEST_BASE_BRANCH'" >&2
-    exit 1
+    exit 255
 }
 
 mode=""
@@ -66,7 +68,7 @@ while [[ "$#" -gt 0 ]]; do
         --all-match | --any-match)
             if [[ -n "$mode" ]]; then
                 echo "$EXCLUSIVE_OPTIONS_ERROR" >&2
-                exit 1
+                exit 255
             fi
             mode="${1#--}"
             shift
@@ -77,16 +79,16 @@ while [[ "$#" -gt 0 ]]; do
             done
             if [[ "${#patterns[@]}" -eq 0 ]]; then
                 echo "Error: must specify at least one file pattern" >&2
-                exit 1
+                exit 255
             fi
             ;;
         --*) 
             echo "Error: unknown option $1" >&2
-            exit 1
+            exit 255
             ;;
         *)
             echo "Error: unexpected argument $1" >&2
-            exit 1
+            exit 255
             ;;
     esac
 done
@@ -100,11 +102,10 @@ done < <(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUES
 if [[ -z "$mode" ]]; then
     # No arguments = any change
     if [[ ${#changed_files[@]} -gt 0 ]]; then
-        echo "true"
+        exit 0
     else
-        echo "false"
+        exit 1
     fi
-    exit 0
 fi
 
 # Returns 0 if the file matches any of the patterns, 1 otherwise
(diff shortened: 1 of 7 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/pr_changed_files
```diff
@@ -7,52 +7,54 @@
...
 #
+# Exit codes:
+#   0: Condition is met (files changed / patterns matched)
+#   1: Condition is not met (no files changed / patterns not matched)
+# 255: Error in command invocation (bad parameters, not in PR context, etc.)
+#
 # Behavior:
 #   With no arguments:
-#     Returns "true" if the PR contains any changes, "false" otherwise
+#     Returns 0 if the PR contains any changes, 1 otherwise
```
AliSoftware:
```suggestion
#     Prints "true" and exit 0 if the PR contains any changes, prints "false" and exit 1 otherwise
```
(and obviously same idea for all others below)

bin/pr_changed_files
```diff
@@ -7,52 +7,54 @@
 #   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
 #   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
 #
+# Exit codes:
+#   0: Condition is met (files changed / patterns matched)
+#   1: Condition is not met (no files changed / patterns not matched)
```
AliSoftware: We should probably document that, in addition to exiting with code `0` or `1`… we still continue to also echo `true` or `false` to `stdout` (as that information can be important to know for people who might want to call it with `SKIP_UI_TESTS=$(pr_changed_files … || true)`

And speaking of, we could probably include example call sites of variable-assignment vs `if` too, probably near that line in the documentation comment (i.e. after Exit codes documentation but before behavior-specific examples), like:

```ruby
# Typical syntax for call sytes
# # To assign the result to a variable — be sure to use `|| true` if your shell has `set -e` enabled to avoid early exit
# $ DOC_UPDATED=$(pr_changed_files --any-match docs/* *.md || true)
# # To use it in an `if` condition — use `>/dev/null` to avoid the "true"/"false" output to pollute your logs
# $ if pr_changed_files --any-match docs/* *.md >/dev/null; then … fi
#   
```

tests/pr_changed_files/test_helpers.sh
```diff
@@ -76,3 +76,30 @@ assert_output() {
...
+assert_equal() {
+    local expected="$1"
+    local actual="$2"
+    local message="${3:-}"
+    
+    if [[ "$expected" != "$actual" ]]; then
+        echo "❌ Assertion failed: Expected '$expected' but got '$actual' ${message:+($message)}" >&2
+        return 1
+    fi
+}
```
AliSoftware: I don't see this `assert_equal()` helper used anywhere?

Plus, there's a missing newline at end of file.

tests/pr_changed_files/test_all_match_patterns.sh
```diff
@@ -26,34 +26,52 @@ git add .
 git commit -m "Add doc files"
 
 # [Test] All changes in docs
-result=$(pr_changed_files --all-match "docs/*")
-assert_output "true" "$result" "Should return true when all changes match patterns"
+pr_changed_files --all-match "docs/*"
+result=$?
+assert_return_code 0 $result "Should return 0 when all changes match patterns"
```
AliSoftware: We probably still want to assert that `pr_changed_files` also echos true/false, in addition to testing the return code…

```suggestion
result=$(pr_changed_files --all-match "docs/*")
assert_return_code 0 $? "Should return 0 when all changes match patterns"
assert_output "true" "$result" "Should print true when all changes match patterns"
```

Though at that point it'll probably be worth it to create a combined assertion helper instead of 2 separate ones (given you'll always want to assert both in all test cases, I think), something like this maybe?
```
assert_result 0 "$?" "$result" "when all changes match patterns"
```

tests/pr_changed_files/test_helpers.sh
iangmaia: Updated on b32c6b90529c09fdb7caf3a7877d315dcf328a55

tests/pr_changed_files/test_all_match_patterns.sh
iangmaia: > it'll probably be worth it to create a combined assertion helper instead of 2 separate ones 

Great idea 👍 implemented on b32c6b90529c09fdb7caf3a7877d315dcf328a55.

bin/pr_changed_files
iangmaia: Expanded documentation on 34c4eac034cc51bb0796588d7573f3fc95d627d1

bin/pr_changed_files
iangmaia: Updated on 34c4eac034cc51bb0796588d7573f3fc95d627d1

bin/pr_changed_files
```diff
@@ -7,52 +7,68 @@
...
+# 255: Error in command invocation (bad parameters, not in PR context, etc.)
+#
+# Output:
+#   In addition to exit codes, the script also outputs "true" or "false" to stdout
+#   when the condition is met or not met respectively.
+#
+# Typical usage patterns:
+#   # To assign the result to a variable — use `|| true` if your shell has `set -e` enabled
+#   $ DOC_UPDATED=$(pr_changed_files --any-match docs/* *.md || true)
```
AliSoftware: I know I keep flip-flopping on this, so sorry in advance… but seeing this invocation written plainly in this comment makes me wonder if this might not be confusing.

In particular, in scripts where we'll want to use this, seeing `pr_changed_files --any-match docs/* *.md || true` will likely make anyone think at first read that the expression kinda means "if any docs changed in the PR… or true if not`, like reading the whole expression as an OR logic condition (using the `||` like if you wanted to write an OR logic like `pr_changed_files --any-match docs/* *.md || "$BUILDKITE_MESSAGE" == *Docs*`)

Of course we could add a comment on those kind of call sites, to mention that the use of `|| true` in that `VAR=$(… || true)` is to "absorb" the exit code and ignore the exit code to avoid the script from exiting due to `set -e` but instead only rely on the stdout…
But having to add such a comment on all of our call sites that are using such syntax just to explain/disambiguate might be a sign that the way our CLI works is not clear enough to begin with…

iangmaia: Yeah no worries, I'm also not fully confident on what's the best strategy here. Sometimes it takes seeing a few examples and discussion to mature the ideas 🙂 

And I agree seeing `pr_changed_files --any-match docs/* *.md || true` makes it confusing, specially given the `-e` semantics aren't immediately obvious (specially for readers not very well versed in bash).
Not sure what next steps could come from this realization -- Continue on the current direction regardless? Move this issue aside for a while and let it sit until we see more actual usages of `pr_changed_files`? Close this PR? Revert the error code part of the PR (not sure if there will be much left, but maybe)? All valid options, I guess.

AliSoftware: What about:
 - Default behavior = exit code + stdout
 - With `--ok` or `--stdout`  flag = stdout only but always exit 0

iangmaia: So other options / combinations:
1. Default: exit code + stdout, using flag (`--ok` or `--stdout`) to use stdout only and always exit 0
2. Default: only stdout (previous behavior), adding a flag to suppress the output (like `--silent`) and use only the exit codes instead
3. Default: only exit code, adding a flag to use the stdout instead (`--stdout`)
...

AliSoftware: 2️⃣  I am wary of 2. —especially the fact it does not exit non-zero on error—because it'd be too easy to use in an `if` condition and forgetting to use the `--silent` flag (or forgetting to compare the stdout to `== "true"`). Which is exactly what happened to you with the WooCommerce PR after all.

Also, a flag like `-q` or `--quiet` or `--silent` reminds me of how those kind of flags are used in POSIX commands like `grep`… for which `grep -q` suppresses the output indeed… but `grep` _without_ `-q` would still exit 1 on no match. While with proposition 2. here using `pr_changed_files` _without_ `--silent` (i.e. default behavior) you are proposing to exit 0. So that would be an inconsistency compared to how the flag with similar name is used in a POSIX tool (`grep`) that is doing similar stuff (searching files for a pattern)… which I think would thus add to the confusion.

---

3️⃣ I like 3. actually.

The only drawback being that we'd have to remember to use `--stdout` when we use `pr_changed_files` to assign a variable (while with 1️⃣ assignment would still work even if we forget that… as long as the shell does not use `set -e` at least), otherwise not only the `X=$(pr_changed_files …)` will always be set to the empty string, but also it might make the script exit early (if `set -e`).
But at that point I'm now hoping that we'll pretty quickly notice that the assignment didn't work as expected in such cases (i.e. the `if: build.env('X')` in our pipeline would not work as expected and `if:` condition would always evaluate to the same boolean value[^1] regardless of PR content) and realize we'd need to use `--stdout` (or use a `if` in our `shared-pipeline-vars` to manually/explicitly decide the value to assign to the var anyway). So I'm ok with that
Especially since I'm hoping that when we'll make PRs to start adopting `pr_changed_files` in repos, we'll test that it it works as expected—by creating dummy commits changing some files and not others to validate job is skipped or not as expected—instead of just approving the PR introducing it without testing it, and we'd then notice if the variable was empty regardless of outcome.

[^1]: I actually have no idea how Buildkite interprets the empty strings as boolean conditions. Does it error if we just use `if: build.env('X')` because it doesn't consider a String to be truthy/falsy, and would require us to be explicit like `if: build.env('X') == 'true'`, or does it interpret empty strings and undefined env vars as falsy but any non-empty string as truthy (which would cause issues with `"false"` being non-empty and thus considered `true`…), or does it have basic attempt at parsing the string to consider `0`, `false`, `no` and empty string as false, `true`, `yes` and `1` as true, and any other string as an error…? Couldn't find a clear explicit mention in Buildkite doc about that, so would be worth experimenting with that to help inform our decision here? 🤔

iangmaia: > 2️⃣ I am wary of 2. —especially the fact it does not exit non-zero on error—because it'd be too easy to use in an if condition and forgetting to use the --silent flag (or forgetting to compare the stdout to == "true"). Which is exactly what happened to you with the WooCommerce PR after all.

Good argument 👍 I was actually tempted to go with 2️⃣  (more also to keep it compatible with the first implementation) but indeed, it's very easy to use in an `if` and miss how it actually works.

> 3️⃣ [...] we'd have to remember to use `--stdout` when we use `pr_changed_files` to assign a variable
> I'm now hoping that we'll pretty quickly notice that the assignment didn't work as expected in such cases

I agree. And accepting the fact that most use will be, at least for a while, without using the output directly in Buildkite, I don't think this will be a big issue.
I'm now on team 3️⃣ 😄

iangmaia: I've implemented 3️⃣ (only exit code by default, `--stdout` flag to use the "true"/"false" output instead): 9c26168873b213291604003dfdaebc41be1e42ac (tests updated on 0b0cedcddd4cadd937bd0b4f34b7b6227605f1d9).

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
 #   pr_changed_files                                  # Check if any files changed
 #   pr_changed_files --all-match <file-patterns...>   # Check if changes are limited to given patterns
 #   pr_changed_files --any-match <file-patterns...>   # Check if changes include files matching patterns
+#   pr_changed_files [--stdout] ...                   # Add --stdout to output "true"/"false" to stdout
```
AliSoftware:
```suggestion
#   pr_changed_files [--stdout] ...                   # Add --stdout to output "true"/"false" to stdout instead of using exit code
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
+#
+# Exit codes:
+#   0: Condition is met (files changed / patterns matched)
+#   1: Condition is not met (no files changed / patterns not matched)
+# 255: Error in command invocation (bad parameters, not in PR context, etc.)
+#
+# Output:
+#   If the flag --stdout is used, the script will output "true" or "false" to stdout
+#   when the condition is met or not met respectively, and will return exit code 0.
```
AliSoftware: Given the new behavior, I'm not sure if splitting the doc in Exit codes vs Output like that makes it clear enough / easy to understand. What about:

```suggestion
# Result
#  - By default, the script exits with code 0 if the condition is met (files changed / patterns matched), 1 if not.
#  - If you use the `--stdout` flag, it will output "true" or "false" instead of using the exit code (and exit 0 in both cases)
#  - The script will exit with code 255 if there is an error in the command invocation (bad parameters, not in PR context, etc.)
```


Or:

```suggestion
# Result
#  - By default, the script exits with code 0 if the condition is met (files changed / patterns matched), 1 if not. This is typically ideal to use in `if` conditions.
#  - If you use the `--stdout` flag, it will output "true" or "false" instead of using the exit code (and exit 0 in both cases). Useful to assign the result to an environment variable for example.
#  - The script will exit with code 255 if there is an error in the command invocation (bad parameters, not in PR context, etc.)
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
+#     fi
+#
+#   # Using stdout output
+#   $ DOCS_UPDATED=$(pr_changed_files --stdout --any-match docs/* *.md)
 #
 # Behavior:
 #   With no arguments:
-#     Returns "true" if the PR contains any changes, "false" otherwise
+#     Returns exit code 0 (outputs "true") if the PR contains any changes
+#     Returns exit code 1 (outputs "false") if no changes
```
AliSoftware:
```suggestion
#     Returns exit code 0 (or outputs "true" with `--stdout`) if the PR contains any changes
#     Returns exit code 1 (or outputs "false" with `--stdout`) if no changes
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
 #
 #   With --all-match:
-#     Returns "true" if ALL changed files match AT LEAST ONE of the patterns
-#     Returns "false" if ANY changed file doesn't match ANY pattern
-#     Note: Will return "true" even if not all patterns are matched by the changed files.
+#     Returns exit code 0 (outputs "true") if ALL changed files match AT LEAST ONE of the patterns
+#     Returns exit code 1 (outputs "false") if ANY changed file doesn't match ANY pattern
+#     Note: Will return 0 even if not all patterns are matched by the changed files.
```
AliSoftware:
```suggestion
#     Returns exit code 0 (or outputs "true" with `--stdout`) if ALL changed files match AT LEAST ONE of the patterns
#     Returns exit code 1 (or outputs "false" with `--stdout`) if ANY changed file doesn't match ANY pattern
#     Note: Will report success even if not all patterns are matched by the changed files.
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
 #           This mode is especially useful to check if the PR _only_ touches a particular subset of files/folders (but nothing else)
 #
 #   With --any-match:
-#     Returns "true" if ANY changed file matches ANY of the patterns
-#     Returns "false" if NONE of the changed files match ANY pattern
-#     Note: Will return "true" even if the PR includes other files not matching the patterns.
+#     Returns exit code 0 (outputs "true") if ANY changed file matches ANY of the patterns
+#     Returns exit code 1 (outputs "false") if NONE of the changed files match ANY pattern
+#     Note: Will return 0 even if the PR includes other files not matching the patterns.
```
AliSoftware:
```suggestion
#     Returns exit code 0 (or outputs "true" with `--stdout`) if ANY changed file matches ANY of the patterns
#     Returns exit code 1 (or outputs "false" with `--stdout`) if NONE of the changed files match ANY pattern
#     Note: Will report success even if the PR includes other files not matching the patterns.
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
 #           This mode is especially useful to check if the PR _includes_ (aka _contains at least_) particular files/folders
 #
 # Examples with expected outputs:
-#   # Check if any files changed, returning "true" if PR has changes, "false" otherwise
+#   # Check if any files changed
 #   $ pr_changed_files
+#   → exit code 0 (outputs "true") if PR has changes
+#   → exit code 1 (outputs "false") if PR has no changes
```
AliSoftware:
```suggestion
#   → exit code 0 if PR has changes
#   → exit code 1 if PR has no changes
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
 #
 #   # Check if only documentation files changed (to skip UI tests for example)
 #   $ pr_changed_files --all-match "*.md" "docs/*"
-#   → "true"  if PR only changes `docs/guide.md` and `README.md`
-#   → "true"  if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
-#   → "false" if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
+#   → exit code 0 (outputs "true") if PR only changes `docs/guide.md` and `README.md`
+#   → exit code 0 (outputs "true") if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
+#   → exit code 1 (outputs "false") if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
```
AliSoftware:
```suggestion
#   → exit code 0 if PR only changes `docs/guide.md` and `README.md`
#   → exit code 0 if PR only changes `docs/image.png` (not all patterns need to match, ok if no *.md)
#   → exit code 1 if PR changes `docs/guide.md` and `src/main.swift` (ALL files need to match at least one pattern)
```

bin/pr_changed_files
```diff
@@ -6,67 +6,89 @@
...
-#   → "true"  if PR changes `.swiftlint.yml`
-#   → "false" if PR only changes `README.md` (none of files match any of the patterns)
-#
-# Returns:
-#   Prints "true" if the condition is met, "false" otherwise
-#   Exits with code 0 regardless of if the condition was met or not, to allow easy variable assignment.
-#   Only exits with non-zero if the command invocation itself was incorrect (called outside of a PR context, incorrect arguments…)
+#   → exit code 0 (outputs "true") if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
+#   → exit code 0 (outputs "true") if PR changes `.swiftlint.yml`
+#   → exit code 1 (outputs "false") if PR only changes `README.md` (none of files match any of the patterns)
```
AliSoftware:
```suggestion
#   → exit code 0 if PR changes `src/main.swift` and `README.md` (AT LEAST one file matches one of the patterns)
#   → exit code 0 if PR changes `.swiftlint.yml`
#   → exit code 1 if PR only changes `README.md` (none of files match any of the patterns)
```

bin/pr_changed_files
```diff
@@ -97,14 +119,27 @@ while IFS= read -r -d '' file; do
...
 done < <(git --no-pager diff --name-only -z --merge-base "$BUILDKITE_PULL_REQUEST_BASE_BRANCH" HEAD | sort)
 
+exit_result() {
+    local result=$1
+    if [[ "$use_stdout" == "true" ]]; then
+        if [[ "$result" -eq 0 ]]; then
+            echo "true"
+        else
+            echo "false"
+        fi
```
AliSoftware: Totally debatable, but could use a more compact notation of `[[ "$result" -eq 0 ]] && echo "true" || echo "false"` to reduce the indentation of nested if. Though I'm also ok if you prefer the explicitness of the full `if` nested statement here.

tests/pr_changed_files/test_helpers.sh
```diff
@@ -61,16 +61,40 @@ cleanup_git_repo() {
...
 }
 
-# Helper to assert expected output
-assert_output() {
+# Helper to assert both return code and output
+# Arguments:
+#   $1 - Expected return code
+#   $2 - Actual return code
+#   $3 - Actual output
+#   $4 - Expected output
```
AliSoftware: Order of arguments feels a bit strange, with Expected and Actual being intertwined like that.
I'd suggest `[Expected return code, Actual return code, Expected output, Actual output]` instead?

iangmaia: Indeed -- updated on feb51f0066bac3a2e2116e908693187a2b0dba9c

bin/pr_changed_files
iangmaia: Hmm yeah I'd rather keep it more explicit 👍 the compact form is fine but I noticed it took me a few milliseconds more to parse 😅

bin/pr_changed_files
```diff
@@ -6,67 +6,85 @@
...
+#  - The script will exit with code 255 if there is an error in the command invocation (bad parameters, not in PR context, etc.)
+#
+# Typical usage patterns:
+#   # Using exit codes
+#   $ if pr_changed_files --any-match docs/* *.md; then
+#       echo "Documentation was updated"
+#     fi
+#
+#   # Using stdout output
+#   $ DOCS_UPDATED=$(pr_changed_files --stdout --any-match docs/* *.md)
```
AliSoftware, line 23: 👌

tests/pr_changed_files/test_helpers.sh
AliSoftware: Mmmh I see you updated some call sites… but not others 😅 

e.g. https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/6e4e1dda7589c8fc437ea059d3f62249456b4bd6/tests/pr_changed_files/test_edge_cases.sh#L68

should be `assert_result 0 $result "true" "$output" …` instead of `assert_result 0 $result "$output" "true" …` for consistency, right?

---

PS: If I'm being nickpicky, after seeing the call sites with that new order, and especially given the "expected output" argument can be long, making the `"$output"` one after it being quite far in the line… I'm now thinking it'd be more readable as `[Actual return code, Expected return code, Actual output, Expected output]` order, so that it looks like `assert_result $result 255 "$output" "Error: this tool can only be called from a Buildkite PR job" …` 🙃 
But feel free to ignore if you're too tired of me being too nitpicky 😅 , it's really a picky detail at that point.

iangmaia: After seeing it, I actually also think `actual, expected` also reads a bit better (looks closer to `assertEquals(myVar, true)`). Updated on ace481391a991033c403993741edba9a25f1bdb6.

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match '*.swift')
+result=$?
+assert_result $result 0 "$output" "true" "Should match .swift files with --stdout"
+
+# [Test] Match multiple patterns - exit code only
+output=$(pr_changed_files --any-match 'docs/*.md' '*.rb')
+result=$?
+assert_result 0 $result "$output" "" "Should match multiple patterns"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "" "Should match multiple patterns"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match 'docs/*.md' '*.rb')
+result=$?
+assert_result 0 $result "$output" "true" "Should match multiple patterns with --stdout"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "true" "Should match multiple patterns with --stdout"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# [Test] Match files with spaces and special characters - exit code only
+output=$(pr_changed_files --any-match 'docs/read me.md' 'docs/special!@*#$chars.md')
+result=$?
+assert_result 0 $result "$output" "" "Should match files with spaces and special characters"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "" "Should match files with spaces and special characters"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match 'docs/read me.md' 'docs/special!@*#$chars.md')
+result=$?
+assert_result 0 $result "$output" "true" "Should match files with spaces and special characters with --stdout"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "true" "Should match files with spaces and special characters with --stdout"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# [Test] No matches - exit code only
+output=$(pr_changed_files --any-match '*.js')
+result=$?
+assert_result 1 $result "$output" "" "Should not match non-existent patterns"
```
AliSoftware:
```suggestion
assert_result $result 1 "$output" "" "Should not match non-existent patterns"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match '*.js')
+result=$?
+assert_result 0 $result "$output" "false" "Should not match non-existent patterns with --stdout"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "false" "Should not match non-existent patterns with --stdout"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# [Test] Directory pattern - exit code only
+output=$(pr_changed_files --any-match 'docs/*')
+result=$?
+assert_result 0 $result "$output" "" "Should match directory patterns"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "" "Should match directory patterns"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match 'docs/*')
+result=$?
+assert_result 0 $result "$output" "true" "Should match directory patterns with --stdout"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "true" "Should match directory patterns with --stdout"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+# [Test] Exact pattern matching - exit code only
 echo "swiftfile" > swiftfile.txt
 git add swiftfile.txt
 git commit -m "Add file with swift in name"
 
-result=$(pr_changed_files --any-match '*.swift')
-assert_output "true" "$result" "Should only match exact patterns"
+output=$(pr_changed_files --any-match '*.swift')
+result=$?
+assert_result 0 $result "$output" "" "Should only match exact patterns"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "" "Should only match exact patterns"
```

tests/pr_changed_files/test_any_match_patterns.sh
```diff
@@ -26,36 +26,68 @@ echo "ruby" > 'src/ruby/main.rb'
...
+
+# Test with stdout
+output=$(pr_changed_files --stdout --any-match '*.swift')
+result=$?
+assert_result 0 $result "$output" "true" "Should only match exact patterns with --stdout"
```
AliSoftware:
```suggestion
assert_result $result 0 "$output" "true" "Should only match exact patterns with --stdout"
```

bin/pr_changed_files
```diff
@@ -97,14 +115,27 @@ while IFS= read -r -d '' file; do
...
+            echo "true"
+        else
+            echo "false"
+        fi
+        exit 0
+    fi
+
+    exit "$result"
+}
+
```
AliSoftware: 💭 Future improvement idea I just had while re-reading the whole code: I wonder if the implementation wouldn't read better (especially the call site of this internal function) if we had 2 separate methods instead:

```bash
exit_true() {
    if [[ "$use_stdout" == "true" ]]; then
        echo "true"
    fi
    exit 0
}

exit_false() {
    if [[ "$use_stdout" == "true" ]]; then
        echo "false"
        exit 0
    fi
    exit 1
}
```

Then the call sites would read `exit_false()` instead of `exit_result 1`, which arguably makes it easier to understand when reading the implementation… 🤔 🤷 

That's just a nitpick though, and definitively not a blocker.

tests/pr_changed_files/test_helpers.sh
```diff
@@ -61,16 +61,40 @@ cleanup_git_repo() {
...
+    local actual_code="$1"
+    local expected_code="$2"
+    local actual_output="$3"
+    local expected_output="$4"
+    local message="$5"
+
+    assert_equal "$actual_code" "$expected_code" "Exit code - $message"
+    if [[ -n "$expected_output" ]]; then
+        assert_equal "$actual_output" "$expected_output" "Output - $message"
+    fi
```
AliSoftware: ⚠️ This just occurred to me: what if we expect an empty output… but we end up getting an unexpected one? The `if [[ -n "$expected_output" ]]` here will skip calling `assert_equal` and we won't detect that the `"$actual_output"` is not empty when we expected it to be… 🤔 

So shouldn't we just not have that `if` at all and always call `assert_equal` in all cases?

```suggestion
    assert_equal "$actual_output" "$expected_output" "Output - $message"
```

AliSoftware (approved):
Apart from the test missing to assert expected empty output (which I think is worth fixing before merging), the rest looks good, so I'm approving to unblock, supposing you'll fix that test helper.

I've also left a nitpick for implementation of `exit_result`, but up to you if you want to adopt it, and not a blocker.

PS: 🎗️ Reminder to update your P2 post accordingly once this ships, to provide up-to-date call examples.

tests/pr_changed_files/test_helpers.sh
iangmaia: Hmm good point, agreed!

bin/pr_changed_files
iangmaia: Great idea though, makes it indeed more readable than `exit_result 1`! Implemented on ecd08ee73b65bd6c746928bad75dcbbe0f023fd9

AliSoftware (approved):

PR #153, opened 2025-02-19
Windows 10 SDK install tests

mokagio:
Builds on top of #144. Initially, this was a dedicated PR because I expected a lot of Git noise by trial and error in CI. But as I kept refining the implementation for the Windows 10 SDK version file validation, I also introduced what I hope are other improvements to the Windows scripts.

- Adds validation of the `.windows-10-sdk-version` content before attempting to use said content as the version to install, including comparing against an allow-list of valid versions
- Replaces all `Write-Host` with `Write-Output`, because `Write-Host` is meant for messages such as user prompts
- Adds a `-DryRun` option to the Windows 10 SDK install script, mostly so it can be tested without doing a full install
- Adds a `SkipWindows10SDKInstallation` to the host setup script, just in case

Together with #144, I think this should be enough for now to ship a new plugin (major) version and begin DRYing the Electron apps.

## Next steps

- Take a look at the CI setup time and consider moving the tests into a single script if the overhead is considerable
- Keep working on #133 to make sure the script properly handle errors

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. — N.A.

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -121,19 +121,19 @@ steps:
 
   - group: ":windows: install_windows_10_sdk Tests"
     steps:
-      - label: ":windows: install_windows_10_sdk Tests - Standard version file"
+      - label: ":windows: install_windows_10_sdk Tests - Version file with valid format and version"
         command: |
-          "123" | Out-File .windows-10-sdk-version
+          "20348" | Out-File .windows-10-sdk-version
           .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
         agents:
           queue: windows
         notify:
           - github_commit_status:
-              context: "install_windows_10_sdk Tests - Standard version file"
+              context: "install_windows_10_sdk Tests - Version file with valid format and version"
 
       - label: ":windows: install_windows_10_sdk Tests - Version file with one new line"
         command: |
-          "123`n" | Out-File .windows-10-sdk-version
+          "20348`n" | Out-File .windows-10-sdk-version
           .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
         agents:
           queue: windows
@@ -143,7 +143,7 @@ steps:
 
       - label: ":windows: install_windows_10_sdk Tests - Version file with more than one new line"
         command: |
-          "123`n`n" | Out-File .windows-10-sdk-version
+          "20348`n`n" | Out-File .windows-10-sdk-version
           .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
         agents:
           queue: windows
@@ -153,7 +153,7 @@ steps:
 
       - label: ":windows: install_windows_10_sdk Tests - Version file with leading whitespaces"
         command: |
-          "  123" | Out-File .windows-10-sdk-version
+          "  19041" | Out-File .windows-10-sdk-version
           .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
         agents:
           queue: windows
@@ -163,7 +163,7 @@ steps:
 
       - label: ":windows: install_windows_10_sdk Tests - Version file with trailing whitespaces"
         command: |
-          "123   " | Out-File .windows-10-sdk-version
+          "18362   " | Out-File .windows-10-sdk-version
           .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
         agents:
           queue: windows
@@ -193,3 +193,15 @@ steps:
         notify:
           - github_commit_status:
               context: "install_windows_10_sdk Tests - Version file with a word"
+
+      - label: ":windows: install_windows_10_sdk Tests - Version file with version number that is not in the allowed list"
+        command: |
+          "12345" | Out-File .windows-10-sdk-version
+          .\tests\test-install-windows-10-sdk.ps1 `
+            -ExpectedExitCode 1 `
+            -ExpectedErrorKeyphrase "Invalid Windows 10 SDK version: 12345"
+        agents:
+          queue: windows
+        notify:
+          - github_commit_status:
+              context: "install_windows_10_sdk Tests - Version file with version number that is not in the allowed list"
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -151,6 +151,16 @@ steps:
           - github_commit_status:
               context: "install_windows_10_sdk Tests - Version file with more than one new line"
 
+      - label: ":windows: install_windows_10_sdk Tests - Version file with trailing whitespace"
+        command: |
+          "123   " | Out-File .windows-10-sdk-version
+          .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
+        agents:
+          queue: windows
+        notify:
+          - github_commit_status:
+              context: "install_windows_10_sdk Tests - Version file with trailing whitespace"
+
       - label: ":windows: install_windows_10_sdk Tests - Version file with a word"
         command: |
           "not an integer" | Out-File .windows-10-sdk-version
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -151,7 +151,17 @@ steps:
           - github_commit_status:
               context: "install_windows_10_sdk Tests - Version file with more than one new line"
 
-      - label: ":windows: install_windows_10_sdk Tests - Version file with trailing whitespace"
+      - label: ":windows: install_windows_10_sdk Tests - Version file with leading whitespaces"
+        command: |
+          "  123" | Out-File .windows-10-sdk-version
+          .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
+        agents:
+          queue: windows
+        notify:
+          - github_commit_status:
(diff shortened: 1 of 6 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bin/install_windows_10_sdk.ps1
```diff
@@ -1,18 +1,41 @@
...
 
 $windowsSDKVersionFile = ".windows-10-sdk-version"
 if (-not (Test-Path $windowsSDKVersionFile)) {
   Write-Output "[!] No Windows 10 SDK version file found at $windowsSDKVersionFile."
   exit 1
 }
 
-$windows10SDKVersion = Get-Content $windowsSDKVersionFile
+$windows10SDKVersion = Get-Content $windowsSDKVersionFile `
+  | Where-Object { $_ -match '\S' }  # Remove empty lines
```
AliSoftware: Worth testing if this other (and arguably more idiomatic) implementation would work too:
```suggestion
$windows10SDKVersion = (Get-Content -TotalCount 1 $windowsSDKVersionFile).Trim()
```

Though tbh not even sure the `.Trim()` is strictly needed

bin/install_windows_10_sdk.ps1
```diff
@@ -1,18 +1,41 @@
...
+
+if ($windows10SDKVersion.Count -ne 1) {
+  Write-Output "[!] Invalid version file format."
+  Write-Output "Expected exactly one non-empty line, got:"
+  Write-Output ($windows10SDKVersion -join "`n")  # The join poperly formats multiple lines
```
```suggestion
  Write-Output ($windows10SDKVersion -join "`n")  # The join properly formats multiple lines
```

bin/install_windows_10_sdk.ps1
```diff
@@ -1,18 +1,41 @@
...
+  exit 1
+}
+
+if ($windows10SDKVersion -notmatch '^\d+$') {
+  Write-Output "[!] Invalid version file format."
+  Write-Output "Expected an integer, got: '$windows10SDKVersion'"
```
AliSoftware: I'm not well versed in how Windows10 SDK versioning works, so completely open question: are we sure it's guaranteed to be an integer? As opposed to versions like `13b4` or `12.7` or `123a` also being a possibility?

bin/install_windows_10_sdk.ps1
mokagio: Before accepting this suggestion, I'll test [the one above](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153#discussion_r1961399525), because if it works I'd say we can remove this check altogether.

I mean, it might be good to let the user know the added rubbish in `.windows-10-sdk-version` but  it might also be an annoying build failure given we were able to parse it anyway.

bin/install_windows_10_sdk.ps1
mokagio: As far as I understand, yes. That's because the integer used in the file is then passed to the SDK downloader and used to identify the version according to the spec in the screenshot below, sourced from [here](https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022):

image

Having said that, when I tried 18362, nothing happened:

image

I wonder whether that's because the version is 10.0.18362.1, there's something wrong with that version, or my command.

For the record, I tried passing 18362.1 and it still did not result in the SDK being installed.

This all makes me wonder whether we might be better off using a default SDK version in the installer, and only falling back to the version file and/or a parameter if the user provides one. After all, I implemented the version file because other tools use it and because the app I was working on had a check for the SDK being installed and so I wanted a single source of truth. But I'd argue now that our tooling takes care of the installing, that check is redundant: the build script should trust the SDK is available.

bin/install_windows_10_sdk.ps1
mokagio: > Though tbh not even sure the `.Trim()` is strictly needed

Turned out to be, but to account for leading and trailing whitespaces, not newlines. See 48410dbc98907c58e8c86cc1280c999ae5b35a24 and 85f7bc24116592482d15a9f6ddd8fc0b43975267

bin/install_windows_10_sdk.ps1
AliSoftware: I like the version file to enforce a specific version instead of always installing the latest though, especially for build reproducibility and idempotence (ie similar to a lockfile ensuring every rerun of a build on a same commits days later will still use the same libs and tools)

mokagio: Ah, good point 👍 

Given the timezone schedule that basically leaves me with a whole workday before the next round of feedback, I'll work on adding an allow-list for the valid versions. Otherwise, I would have suggested to merge this as is and ship the allow-list as a followup improvement.

bin/install_windows_10_sdk.ps1
```diff
@@ -1,17 +1,64 @@
+# Install the Windows 10 SDK and Visual Studio Build Tools using the value in .windows-10-sdk-version.
+#
+# The expected .windows-10-sdk-version format is a integer representing a valid SDK component id.
+# The list of valid component ids can be found at
+# https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
+#
+# Example:
+#
+#   20348
+
```
mokagio, line 10: See https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/144#discussion_r1961889583 by @iangmaia 

I added the documentation here rather than in the prepare script as per @iangmaia 's comment because it seemed like a better place where to explain the expectation.

Also, I am aware we have two PRs that add docs here, but it felt best to add a manual version in this context to get it over the line while waiting for the LLM-generated doc approach to be discussed, see #159 #160

mokagio:
@iangmaia @AliSoftware I'd appreciate a second (third, fourth?) look after the latest followup, so ideally we can merge in #144 and keep going.

Thanks! 🙇‍♂️

.buildkite/pipeline.yml
```diff
@@ -118,3 +118,98 @@ steps:
         notify:
           - github_commit_status:
               context: "pr_changed_files Tests: Edge Cases"
+
+  - group: ":windows: install_windows_10_sdk Tests"
```
AliSoftware, line 122: What about making use of our new `pr_changed_files` to run this group conditionally?

The nice thing is that this `a8c-ci-toolkit` repo is the only one for now where it'd actually be easy to use that helper in `shared-pipeline-vars` to set up an env var you could then use as an `if: build.env('HAS_INSTALL_WINDOWS_10_SDK_CHANGES') == 1`.

This is because unlike all other repos for which the `a8c-ci-toolkit` is not loaded on the `uploader` agent when the main step that does `pipeline upload` is run on it, our default steps don't include it as a plugin… in this specific repo we don't even need to load it as a plugin because it's already a binary in the working copy of the repo it's trying to run CI on 😛 

Which means you should be able to use `bin/pr_change_files` in this repo directly (instead of having the plugin loaded and having its `bin/` folder added to the `$PATH`) to `export HAS_INSTALL_WINDOWS_10_SDK_CHANGES=$(bin/pr_changed_files --stdout --any-match './bin/install_windows_10_sdk.ps1' './tests/test-install-windows-sdk.ps1')` 😉 

Same for the other one btw.

.buildkite/pipeline.yml
```diff
@@ -118,3 +118,98 @@ steps:
...
+    steps:
+      - label: ":windows: install_windows_10_sdk Tests - Version file with valid format and version"
+        command: |
+          "20348" | Out-File .windows-10-sdk-version
+          .\tests\test-install-windows-10-sdk.ps1 -ExpectedExitCode 0
```
AliSoftware, line 127: I just realized the inconsistency we have here between `snake_case` used for `bin/*` scripts and `kebab-case` used for `tests/*` 🤔

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -8,24 +8,28 @@
 #  - Install the Windows 10 SDK if it detected a `.windows-10-sdk-version` file(2)
 #
 # (1) The certificate it installs is stored in our AWS SecretsManager storage (`windows-code-signing-certificate` secret ID)
-# (2) You can skip the Win10 install even if `.windows-10-sdk-version` file is present by using the `SKIP_WINDOWS_10_SDK_INSTALL=1` env var before calling this script
+# (2) You can skip the Windows 10 SDK installation regardless of whether `.windows-10-sdk-version` is present by calling the script with `-SkipWindows10SDKInstallation`.
```
AliSoftware, line 11: 👍

AliSoftware (approved):

.buildkite/pipeline.yml
mokagio: That would be neat and great dogfooding.

.buildkite/pipeline.yml
mokagio: I noticed that too, but didn't take the time to write it down or address it. Sounds like an easy follow up PR for me to tackle.

PR #154, opened 2025-02-19
Introduce basic .cursorrules for this repo

AliSoftware:
Since many of us are using Cursor IDE nowadays, I figured I'd share my initial `.cursorrules` files I created while working on this repo.

This is a very early draft / version of the rules I asked Cursor itself to come up with after analyzing our codebase of this repo + giving it some additional suggestions for rules about documentation and bash 3.2.57. But I'd love for you to add more ideas to that file so that we can consolidate those rules with our conventions and guidances we want for this repo.

---

- [ ] ~I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary~.

### DIFF
--- a/.cursorrules
+++ b/.cursorrules
@@ -2,7 +2,7 @@
 
 ## Documentation Requirements
 
-When editing or creating files in /bin:
+When editing or creating files in `/bin`:
 - Every shell script must have comprehensive header documentation
 - Documentation must include: description, usage, parameters, examples
 - For complex logic, provide detailed examples of different use cases
@@ -36,7 +36,7 @@ When editing or creating files in /bin:
 #
 # Returns:
 #   [Description of exit codes and their meanings]
-
+#
 # Requirements:
 #   [List of required external commands/programs]
 #   [Any platform-specific requirements]
@@ -45,8 +45,8 @@ When editing or creating files in /bin:
 ## Shell Script Standards
 
 For all bash shell scripts:
-- Ensure compatibility with bash 3.2.57
-- Follow shellcheck recommendations
+- Ensure compatibility with `bash 3.2.57`
+- Follow `shellcheck` recommendations
 - Use proper error handling and input validation
 - Document cross-platform considerations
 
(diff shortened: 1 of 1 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.cursorrules
```diff
@@ -0,0 +1,65 @@
...
+#   [Provide practical examples demonstrating different use cases]
+#
+# Notes:
+#   [Any important notes about script behavior, dependencies, or limitations]
+#
+# Returns:
+#   [Description of exit codes and their meanings]
+```
+
+## Bash 3.2 Compatibility
```
iangmaia: Perhaps it is worth it to explicitly mention a cross platform requirement, as very often I've seen mistakes on that front in generated code.

.cursorrules
```diff
@@ -0,0 +1,65 @@
...
+Avoid using:
+- Associative arrays (declare -A)
+- mapfile/readarray command
```
iangmaia:
```
- mapfile/readarray command
```
Very often it resorts to that, I wonder why 😅 
I very often write in prompts to avoid using external tools unless really necessary.

Maybe it makes sense, on top of an explicit "avoid list" of commands, add a "allow list" of common tools that are fine to use?

.cursorrules
AliSoftware: Sure, though I'd expect mentioning compatibility with Bash 3.2 would be enough to make it cross-platform? 🤔

.cursorrules
AliSoftware: > I very often write in prompts to avoid using external tools unless really necessary.

`mapfile` and `readarray` are not _external tools_ (nor _commands_) though. They are bash built-ins. It's just that those built-ins were only introduced in bash 4.0+. Just like support for e.g. `declare -A` for associative arrays was only added in bash 4.0+ too.

And I'm not surprised that it often resorts to those 2 `mapfile` and `readarray` built-ins, because they are indeed really useful to read data into arrays in bash 4.0, while in bash 3.x we had to resort to more convoluted manual constructs like `while read` loops to achieve what `mapfile` and `readarray` can now achieve more concisely… as long as you run bash 4.0+ and not an earlier version like `3.2` that doesn't have those build-ins.

> Maybe it makes sense, on top of an explicit "avoid list" of commands, add a "allow list" of common commands that are fine to use?

Are you suggesting commands or bash build-ins there?

I don't see a point of us listing all the allowed bash 3.2 build-ins manually, as it'd be a very very long list, and I'd expect just mentioning bash 3.2 would be enough for Cursor to know what builtins are supported in that version of `bash`.

What I could see useful though, is indeed a list of _external commands_ we could allow it to use, like `jq` and the like, that are not built-ins but indeed external tools we have to install for them to be available, but we know are installed on most of our agents that will use those scripts.

.cursorrules
iangmaia: Though still it might bring in Linux-only / macOS-only **external tools** even if you mention Bash 3.2, no?

.cursorrules
iangmaia: > What I could see useful though, is indeed a list of external commands we could allow it to use, like jq and the like

🎯 Yep, that's what I meant.

.cursorrules
AliSoftware: Ah, so you're talking about _external commands_ we might _invoke_ from those scripts, not bash builtins/functions/constructs/syntax, I see. Yes, good point, would be worth adding indeed.

There's already a "Document cross-platform considerations" item (L16) and "Document all dependencies and requirements" (L65), but we could definitively be more precise about that point, possibly with a dedicated `##` rule section for that in particular 👍

.cursorrules
AliSoftware: @iangmaia I've pushed a new commit to add additional rules about cross-platform, as well as a validation section, and asked Cursor to clean up the formatting. It should be more complete now 🙂

iangmaia (approved):
Thank you for this 👍 it looks like a great starting point!

PR #155, opened 2025-02-20
Add Cursor rules for PowerShell scripts

AliSoftware:
## What

As per my suggestion in paaHJt-7WG-p2#comment-10141, this is the result of me asking Cursor to add `.cursorrules` for PowerShell in addition to the ones I recently added for `bash`

## How

I used the following prompts to ask Cursor to generate those rules itself:

> You are an expert in PowerShell and Bash scripting.
> Looking at the PowerShell `.ps1` scripts we have in @bin , and accounting for the current @.cursorrules we have for bash scripts, suggest additional rules to @.cursorrules that would be useful to you (Cursor and the LLM) to help apply similar good standards and guidelines for `.ps1` scripts.

_It then generated some rules, but including mentions of ensuring the `.ps1` scripts would be Cross-Platform for Windows, Linux and macOS—based on us having a similar set of Compatibilty rules for `bash`, I assume… So I asked it to refine its suggestion:_

> We will only write `.ps1` PowerShell scripts for Windows systems, so you can remove all the considerations and instructions about those being Cross-Platform

_It then updated its suggestions to the ones I've pushed in cbf4991._

I then added some manual tweaks to the `.cursorrules` file in 7766406 and e9c04a2.

[EDIT] After all that I then moved the markdown section to separate `.cursor/rules/*.mdc` files (instead of having them all in a single `.cursorrules`), since that seems to be the new suggested format to use for those in latest Cursor versions.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware:
Note: we should probably [switch to use `.cursor/rules/NN-name.mdc` files instead of a unique `.cursorrules` file](https://docs.cursor.com/context/rules-for-ai) at that point.

Though:
 - I'm unclear what the expected format for the `.mdc` files, in particular if there's some standard in how to provide the contexts and file patterns that are [mentioned in the docs](https://docs.cursor.com/context/rules-for-ai). I've seen some people mention the YAML front-matter to provide those… but I don't see the format being properly documented officially
 - It [seems there are currently some bugs to iron out on editing those `.mdc` files](https://github.com/getcursor/cursor/issues/2683)…?

[EDIT] Conversion to `.mdc` files done in aeffa77.
I've indeed encountered the bug mentioned in https://github.com/getcursor/cursor/issues/2683 when asking Cursor to generate those `.mdc` files for me from the `.cursorrules` file, so I had to create them manually instead, and quite and restart Cursor. But then after prompting it with `Can you help me improve @cache_cocoapods_specs_repos based on your @rules ?` its answer suggested it correctly accounted for the rules written in the new `.mdc` format (suggested to use `[[ … ]]`, added documentation, …), which seems to confirm the converted rules would still work as expected 👍

AliSoftware:
I've created a draft PR #157 on top of this to check/demo how well those rules could help us use Cursor to improve our scripts and their documentation.

.cursor/rules/powershell_scripts.mdc
```diff
@@ -0,0 +1,141 @@
...
+        - Required Windows Roles: [list roles]
+        - Administrator privileges: [Yes/No]
+
+.OUTPUTS
+    Description of the script's output.
+
+.RETURNVALUES
+    Description of possible return values and their meanings.
+#>
+```
```
mokagio, line 87: This doc format is _huge_

mokagio (approved):
Thanks @AliSoftware ! 

Cursor rules are new to me, so I don't know how effective these would be. But I like how they are tracked in the repo so we can iterate 😄 

Happy to merge and try them out.

.cursor/rules/powershell_scripts.mdc
AliSoftware: Yes indeed.

But I'd encorage you to test it by asking Cursor to `Add documentation header in @bin/prepare_windows_host_for_app_distribution.ps1` as a prompt—which should make it detect that this file has the `.ps1` extension so that it should pick the `@rules` for it automatically—and see what it generates.

In particular I believe it'd be smart enough to only add the sections of this documentation template that are relevant and used (e.g. not include `.EXAMPLE` if it doesn't have any to provide, same for `.PARAMETER` entries…

AliSoftware: PS: You might have to quit Cursor before switching to the git branch that have this `.mdc` files and restarting Cursor, as based on [this bug report](https://github.com/getcursor/cursor/issues/2683) it seems that the files are parsed at Cursor launch (and the cursor executable has an open file handle on them from that point on…), so changes in the `.mdc` files might not be taken into account until you quit and restart the IDE…?

mokagio: Got Cursor to generate docs in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/159

PR #159, opened 2025-02-21
Experiment with Cursor-generated docs for PowerShell scripts

mokagio:
See discussion at https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/155#discussion_r1964277124

This is on top #153 which has all the up to date PS1 scripts.

---

For the record, here's the prompts I gave.

- update the documentation for the ps1 commands based on the rules
- The PS1 scripts should not have a Bash shebang! Also, do it only for the scripts in bin/ the ones in tests/ are fine the way they are
- mmm... why did you skip `@path_aware_refreshenv.ps1`  ?
- thanks. revisit the requirements for @prepare_windows_host_for_app_distribution.ps1 are you sure about them?
  - _at this point it added more requirements..._
- but... doesn't the script itself install Chocolately and jq? 

I also asked it to generate the list of prompts I gave it, which it did but left out quite a few

---

At this point, I don't like this setup. The documentation it generates is very verbose. I suppose we could update the rules to make it more concise, removing redundant entries (like having requirements in notes and in the "Requirements" section). But I felt it took me as much time to get it to generate the docs and iterate on them, plus my final read, as it would have taken me to write a few lines explaining what the script does.

I'm not going to push against having this kind of documentation if @Automattic/apps-infrastructure feels it's useful. I acknowledge I have a low threshold for how much commented text I can read before looking at code. And after all, we did discuss doing a better job at documentation to help everyone in the team be productive across our tooling, particularly in the areas where one has less experience. But... I wonder if verbose headers docs are the way here vs smaller scripts or ad hoc comments to explain what each chunk does, for example.

I'm also okay to keep using LLMs to generate the docs. If we do this, I'd love to have a way to make it clear that's the primary author of the documentation. I feel as a reader it would help me understand why the docs read the way they do. And as a dev, it would make it clear that I should ask an LLM to generate the docs instead of — big air quotes — "waisting" time writing them myself from scartch.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/install_windows_10_sdk.ps1
```diff
@@ -1,3 +1,39 @@
+# Script: install_windows_10_sdk.ps1
+#
+# Description:
+#   Installs the Windows 10 SDK and Visual Studio Build Tools based on a version
+#   specified in a .windows-10-sdk-version file.
```
mokagio, line 5:
```suggestion
#   Installs the Windows 10 SDK and Visual Studio Build Tools based on the
#   version specified in a .windows-10-sdk-version file.
```

bin/path_aware_refreshenv.ps1
```diff
@@ -1,17 +1,35 @@
...
-# To avoid the issue, we save the PATH pre-refreshenv and then manually add all
-# the components that were removed.
+# Examples:
+#   # After installing a package that modifies PATH
+#   choco install some-package
+#   path_aware_refreshenv.ps1
+#
+# Notes:
+#   - Saves PATH before refreshenv and restores unique entries after
+#   - Use after installing packages that modify PATH at runtime
```
mokagio, line 23: This seems redundant

```suggestion
```

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -1,3 +1,46 @@
...
+#
+# Examples:
+#   # Full setup including SDK if version file exists
+#   prepare_windows_host_for_app_distribution.ps1
+#
+#   # Skip SDK installation
+#   prepare_windows_host_for_app_distribution.ps1 -SkipWindows10SDKInstallation
+#
+# Notes:
+#   - Requires administrator privileges
```
mokagio, line 29: Redundant because already in the requirements section

```suggestion
```

bin/install_windows_10_sdk.ps1
```diff
@@ -1,3 +1,39 @@
...
+#   # Install SDK with version from .windows-10-sdk-version
+#   install_windows_10_sdk.ps1
+#
+#   # Validate version file without installing
+#   install_windows_10_sdk.ps1 -DryRun
+#
+# Notes:
+#   - Requires .windows-10-sdk-version file in current directory
+#   - Version must be one of the allowed SDK versions from Visual Studio
+#   - Installation requires admin privileges
```
mokagio, line 26: On a second pass of the diff, I noticed that none of the docs used the full spec from `.cursor/rules/powershell_scripts.mdc`

```
.NOTES
    Author: [Author name]
    Last Edit: [Date]
    Version 1.0 - Initial release
    Requirements: Windows PowerShell 5.1
    Windows Requirements:
        - Windows Server 2019 or later
        - Required Windows Features: [list features]
        - Required Windows Roles: [list roles]
        - Administrator privileges: [Yes/No]
```

Maybe I didn't prompt it well enough?

AliSoftware:
@mokagio I did the same experiment on my end asking my Cursor Composer to generate those docs too, you can see the result in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/160. I left the generated content as-is in the draft PR to highlight what it generated without any manual tweaking.

PS: Did you close the folder/window and quit Cursor before switching to the commit that contained the `.mdc` files then reopening Cursor only then, before trying to ask it for the prompt?

mokagio:
> Did you close the folder/window and quit Cursor before switching to the commit that contained the .mdc files then reopening Cursor only then, before trying to ask it for the prompt?

I didn't record myself going through the process so I might be mistaken, but I'm pretty sure I did because I remember reading about that gotcha either in the other docs PR or on Slack.

mokagio:
> I left the generated content as-is in the draft PR to highlight what it generated without any manual tweaking.

This comment actually makes me think of something interesting / noteworthy / that should be kept in mind when thinking of LLM-generated docs: The result is non-deterministic! 

I don't think that's a show stopper, and I'd like to think that with a set of rules well configured, the result should be similar across different prompt styles. But that's something I don't think I ever thought about and that makes me a bit uncomfortable, in a sense. In a world where we go through a lot of trouble to get deterministic result (VMs, lock-files, inject app state to decouple unit tests from it, etc.) getting different results every time is unsettling.

Again, I don't think it's a blocker, especially when it comes to documentation written in natural language and with the expectation that the better we get at configuring the `.cursor/rules` and at prompting, the better and more reliable the results will be.

AliSoftware:
That is indeed good to keep in mind. Though, to be clear, the use of LLM here was never intended to generate (deterministic) documentation.

After all, the goal is not to have the code not documented and have a process that would generate documentation dynamically on the fly each time we need it, but rather to have a robot intern help us write _some_ documentation for code so we can review and commit it once it's written. In that sense, the LLM is just like a robot intern. We might each have a slightly different intern each with its slightly different voice and tone and style of writing, but that doesn't really matter if the writing style is different between each attempt at asking the LLM the same question from different machines or at different times, what matters is that we end up with some sort of documentation that makes sense and is helpful.

Then the more documentation we'll write as headers of our shell scripts then commit to git, the more likely it will be, when the next person asking their LLM to write new documentation for another script, to pick us a similar tone to the documentation that was already written, as it'd likely use it as inspiration. But even if it didn't what matters is that the LLM is only here as an intern to help speed up the process and gather the information to write an initial take. It doesn't have to be deterministic.

mokagio:
Closing. Maybe some folks still use Cursor, but even then, there's been so much improvement and change since opening this, that it would be better to open a new PR instead.

PR #161, opened 2025-02-21
Prepare release `3.12.0`

iangmaia:
This PR prepares the release of version 3.12.0.

Changes in this version:
- Enhance `pr_changed_files` command to support both exit codes (default) and "true"/"false" string output (via `--stdout` flag) [#151]

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Enhance `pr_changed_files` command to support both exit codes (default) and "true"/"false" string output (via `--stdout` flag) [#151]
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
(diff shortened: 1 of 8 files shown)


AliSoftware (commented):
Technically speaking this should be a breaking change and thus lead to a `4.0.0` major version bump 🤔 

That being said I'm not sure anyone has started using the `pr_changed_files` yet since it was introduced, so this is probably fine…

But also, major version bumps are cheap to do, especialluy in buildkite-plugin land, as it's just a matter of how you name the version and tag, so… couldn't hurt to still do it?🤷‍♂️

mokagio:
For what is worth, I have a breaking change queue in #144, too.

But, as @AliSoftware said, tags are super cheap here, so we don't necessarily need to wait for it to merge and batch the two breaking changes.

iangmaia:
> major version bumps are cheap to do, especialluy in buildkite-plugin land, as it's just a matter of how you name the version and tag, so… couldn't hurt to still do it?

> tags are super cheap here, so we don't necessarily need to wait for it to merge and batch the two breaking changes.

Agreed! And there was an issue with this PR: I created it from the last PR branch instead of trunk 🙈. So I've deleted this release branch and I'm gonna start with `4.0.0`.

PR #162, opened 2025-02-24
Fail early if AWS CLI is missing on not configured

mokagio:
Earlier today, @geekygecko reported (ref p1740374258002719/1740354379.359239-slack-CC7L49W13) an [odd failure in Buildkite](https://buildkite.com/automattic/pocket-casts-android/builds/10679#019535dd-890f-413d-9c9c-2e0494ee8fe5):

```
[2025-02-24T02:52:22Z] Using a8c-ci-cache as cache bucket
[2025-02-24T02:52:23Z] Restoring cache entry pocket-casts-android-Linux-x86_64-ruby3.2.2-7540d16bf427a0fd78072941f12b6d55fedfdd343dd59353ad9ed1f26c4dc8f9
[2025-02-24T02:52:23Z] 	Downloading
[2025-02-24T02:52:25Z]
[2025-02-24T02:52:25Z] Unable to locate credentials. You can configure credentials by running "aws configure".
🔐 Installing Secrets
[2025-02-24T02:52:26Z] bundler: command not found: fastlane
[2025-02-24T02:52:26Z] Install missing gem executables with `bundle install`
```

Looking through the `restore_cache` implementation, I'd say this was due to the bucket acceleration check failing in a `|| true` combo.

https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/df99bdfb1590bdb55ae69af2d3098f24afdcb9ec/bin/restore_cache#L29

While the `|| true` makes sense, it comes at the price of swallowing an occasional error such as this one.

I don't know how the error occurred and it looks like just a hiccup because upon retry the same step [worked fine](https://buildkite.com/automattic/pocket-casts-android/builds/10679#019536da-8011-4e10-921a-7748ebfd0664).

Still, I think it doesn't hurt to be proactive and to fail early if we have issue with the AWS CLI setup. If anything, It would make it clearer for the user why the build failed.

This PR introduces as command to check that the AWS CLI is available and the credentials are set up. 

---

### DIFF
--- a/bin/verify_aws_cli_setup
+++ b/bin/verify_aws_cli_setup
@@ -1,0 +1,19 @@
+#!/bin/bash -eu
+
+echo "~~~ Ensuring AWS CLI is installed and configured."
+
+# Check if aws command is available
+if ! command -v aws &> /dev/null; then
+  echo "[!] AWS CLI not found."
+  exit 1
+else
+  echo "AWS CLI found."
+fi
+
+# Check if AWS credentials are configured
+if ! aws sts get-caller-identity &> /dev/null; then
+  echo "[!] AWS credentials not found."
+  exit 1
+else
+  echo "AWS credentials found."
+fi
--- a/bin/verify_aws_cli_setup
+++ b/bin/verify_aws_cli_setup
@@ -1,19 +1,52 @@
 #!/bin/bash -eu
 
-echo "~~~ Ensuring AWS CLI is installed and configured."
+echo "--- :aws: Ensuring AWS CLI is installed and configured..."
 
-# Check if aws command is available
+echo "~~~ Checking if AWS CLI is installed..."
 if ! command -v aws &> /dev/null; then
   echo "[!] AWS CLI not found."
   exit 1
 else
-  echo "AWS CLI found."
+  echo "AWS CLI found at $(which aws)."
+  aws --version
 fi
 
-# Check if AWS credentials are configured
+echo "~~~ Checking if AWS credentials are configured in env vars or config files..."
+
+if [ -n "${AWS_ACCESS_KEY_ID:-}" ]; then
+    echo "AWS_ACCESS_KEY_ID found in environment variables."
+else
+    echo "No AWS_ACCESS_KEY_ID in environment variables."
+fi
+
+if [ -n "${AWS_SECRET_ACCESS_KEY:-}" ]; then
+    echo "AWS_SECRET_ACCESS_KEY found in environment variables."
+else
+    echo "No AWS_SECRET_ACCESS_KEY in environment variables."
+fi
+
+AWS_CONFIG_PATH="$HOME/.aws/config"
+if [ -f "$AWS_CONFIG_PATH" ]; then
+    echo "AWS config file exists at $AWS_CONFIG_PATH."
+    echo "Printing config via aws configure list:"
+    aws configure list
+else
+    echo "No AWS config file found at $AWS_CONFIG_PATH."
+fi
+
+AWS_CREDENTIALS_PATH="$HOME/.aws/credentials"
+if [ -f "$AWS_CREDENTIALS_PATH" ]; then
+    echo "AWS credentials file exists at $AWS_CREDENTIALS_PATH."
+    echo "Printing profiles via aws configure list:"
+    aws configure list-profiles
+else
+    echo "No AWS credentials file found at $AWS_CREDENTIALS_PATH."
+fi
+
+echo "~~~ Checking AWS credentials via CLI call..."
 if ! aws sts get-caller-identity &> /dev/null; then
-  echo "[!] AWS credentials not found."
+  echo "[!] AWS credentials check failed."
   exit 1
 else
-  echo "AWS credentials found."
+  echo "AWS credentials verified successfully."
 fi
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -158,7 +158,9 @@ steps:
         command: |
           echo "~~~ :hammer: Remove AWS credentials"
           set -x
-          # Removing ~/.aws does not work and neither does clearing the AWS CLI env vars
+          rm -rf ~/.aws
+          unset AWS_ACCESS_KEY_ID
+          unset AWS_SECRET_ACCESS_KEY
           aws configure set aws_access_key_id ""
           aws configure set aws_secret_access_key ""
           set +x
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -129,3 +129,23 @@ steps:
     notify:
       - github_commit_status:
           context: "verify_aws_cli_setup Test succeeds when AWS CLI setup"
+
+  - label: ":aws: verify_aws_cli_setup Test fails when missing AWS"
+    agents:
+      queue: default
+    plugins:
+      # Alpine shouldn't have AWS CLI installed, so this should fail
+      - docker#v5.7.0:
+          image: "alpine:latest"
+          propagate-exit-status: false
+    command: |
+      echo "--- :microscope: Run verify_aws_cli_setup"
+      if bin/verify_aws_cli_setup; then
+        echo "Error: Script succeeded but should have failed (AWS CLI not present in container)"
+        exit 1
+      else
+        echo "Success: Script failed as expected when AWS CLI not present"
+      fi
+    notify:
+      - github_commit_status:
+          context: "verify_aws_cli_setup Test fails when missing AWS"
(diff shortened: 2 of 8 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware:
> due to the bucket acceleration check failing

For the record, a similar thing happened to @wojtekn today in p1740390251662719-slack-CC7L49W13 / https://buildkite.com/automattic/studio/builds/4463#0195375a-7b96-4fbc-9737-c8dd54f8b2a2 . Retries didn't seem to help :/

ParaskP7:
Ah @mokagio so, based on [this conversation](https://github.com/Automattic/pocket-casts-android/pull/3653#issuecomment-2684366123), you think this is related to the `ACCELERATION_STATUS` and say the `aws s3api` command, not the `aws s3 cp` command that follows right after, right? 🤔

What I want to add to this investigation is:
1. If that is indeed because of the `ACCELERATION_STATUS` and say the `aws s3api`, then we should add this `verify_aws_cli_setup` to more scripts that are using that, like:
    - download_artifact -> [ACCELERATION_STATUS](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/download_artifact#L31)
    - upload_artifact -> [ACCELERATION_STATUS](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/upload_artifact#L32C1-L32C20)
2. If that is related to `aws s3 cp` instead, then we should add this `verify_aws_cli_setup` to even more scripts that are using it, like:
    - cache_repo -> [aws s3 cp](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/cache_repo#L19)
    - download_artifact -> [aws s3 cp](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/download_artifact#L33-L38)
    - upload_artifact -> [aws s3 cp](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/upload_artifact#L34-L39)

Isn't that right? 🤔

mokagio:
thanks @ParaskP7 !

> you think this is related to the ACCELERATION_STATUS and say the aws s3api command, not the aws s3 cp command that follows right after, right? 🤔

That was my first guess because of the error message we got in the logs. So [I tried removing the check in a build](https://buildkite.com/automattic/studio/builds/4488#01953ad7-e316-40f5-84e2-036440128996) and the next `aws` call failed. Log:

```
[2025-02-25T02:09:12Z] upload failed: ./studio-npm-Linux-x86_64-node-v20.9.0-419ddaa1b263c7004b4aa666c4262040324412d711925710245fc7a8eb6fb056-87e6f9ba2e212b4f4c85bb4d1c92b69a4d31cfa04cd60da4e34672e1a7359389 to s3://a8c-ci-cache/studio-npm-Linux-x86_64-node-v20.9.0-419ddaa1b263c7004b4aa666c4262040324412d711925710245fc7a8eb6fb056-87e6f9ba2e212b4f4c85bb4d1c92b69a4d31cfa04cd60da4e34672e1a7359389 An error occurred (AuthorizationHeaderMalformed) when calling the CreateMultipartUpload operation: The authorization header is malformed; a non-empty Access Key (AKID) must be provided in the credential.
```

See p1740449278048889/1740390251.662719-slack-CC7L49W13

Thank you for double checking where `aws` is used. I thought I did it but I must have just went with what I thought was happening instead of running a text search. 

I'll update!

ParaskP7:
Thanks for the reply @mokagio ! 🥇 

> That was my first guess because of the error message we got in the logs. So [I tried removing the check in a build](https://buildkite.com/automattic/studio/builds/4488#01953ad7-e316-40f5-84e2-036440128996) and the next aws call failed. Log:

Got it, thanks for sharing, I must admit I didn't follow the full Slack conversation. 😊 

> Thank you for double checking where aws is used. I thought I did it but I must have just went with what I thought was happening instead of running a text search.
> 
> I'll update!

Awesome, thanks for the update! 🌟

AliSoftware:
While I agree that it's better to avoid having builds fail randomly just due to a cache issue with AWS, I wonder if adopting this won't make us totally miss if/when this happens again in the future, shoving the issue under the rug instead.

Like, if we had this in place already, I highly doubt that we'd have realized at any point that the cache was not working on EC2 instances and that there would have been something to investigate.
Likewise, if we start adopting this in all the repos, how will we know if the issue was only transient and a glitch on AWS's side that solved itself after a couple of days[^1], or if the issue is still here and we need to make sure we find time to investigate deeper at some point?

So if the issue was just a glitch on AWS's side and has since disappeared / been fixed on their end, maybe it's not worth adopting this after all? 🤔 

[^1]: I'm actually curious if the issue is still present a couple of days later; I haven't seen anyone else complain in our Slack channel about those issues, and given the random nature of them—including the AWS credentials [working at one part of a job](https://buildkite.com/automattic/pocket-casts-android/builds/10679#019535dd-890f-413d-9c9c-2e0494ee8fe5/181-187) but [failing only a couple of lines later in the same job](https://buildkite.com/automattic/pocket-casts-android/builds/10679#019535dd-890f-413d-9c9c-2e0494ee8fe5/194-199) inside the same machine—I'm really hoping this was a glitch on AWS's side one day and that they might have resolved it since?

bin/verify_aws_cli_setup
```diff
@@ -0,0 +1,52 @@
...
+if [ -f "$AWS_CREDENTIALS_PATH" ]; then
+    echo "AWS credentials file exists at $AWS_CREDENTIALS_PATH."
+    echo "Printing profiles via aws configure list:"
+    aws configure list-profiles
+else
+    echo "No AWS credentials file found at $AWS_CREDENTIALS_PATH."
+fi
+
+echo "~~~ Checking AWS credentials via CLI call..."
+if ! aws sts get-caller-identity &> /dev/null; then
```
AliSoftware, line 47: My first reaction to this was wondering what would happen if the EC2 instance [has the proper role to access the S3 bucket](https://github.com/Automattic/buildkite-ci/blob/trunk/src/aws/access/policies/can_use_ci_cache.tf) we need it to access (`s3://a8c-ci-cache`), but did not have the role/policy to perform that STS call.

But [looking at the docs](https://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html) it appears  that there's no permission needed to call this one 👍  So we should be good on that front at least.

bin/verify_aws_cli_setup
```diff
@@ -0,0 +1,52 @@
...
+fi
+
+AWS_CREDENTIALS_PATH="$HOME/.aws/credentials"
+if [ -f "$AWS_CREDENTIALS_PATH" ]; then
+    echo "AWS credentials file exists at $AWS_CREDENTIALS_PATH."
+    echo "Printing profiles via aws configure list:"
+    aws configure list-profiles
+else
+    echo "No AWS credentials file found at $AWS_CREDENTIALS_PATH."
+fi
```
AliSoftware, line 44: Worth noting that it is totally valid for an EC2 instances to not have `~/.aws/config`, `~/.aws/credentials`, nor env vars to provide AWS access, because those EC2 instances living in AWS infra are provided their credentials via roles and policies instead. I guess that's why you didn't `exit 1` in the `else` of those cases 👍  but just wanted to highlight that fact for other reviewers too.

AliSoftware (approved):

mokagio:
> While I agree that it's better to avoid having builds fail randomly just due to a cache issue with AWS, I wonder if adopting this won't make us totally miss if/when this happens again in the future, shoving the issue under the rug instead.

I was chatting about this with @twstokes in our 1:1 actually and came to the conclusion that before considering this work "done" we should at least have a way to log when the CLI authentication fails, so that we can monitor and decide whether the check is still required.

mokagio:
Closing without merging as discussed in paaHJt-82Y-p2.

In short: The issue this proactive check guards against was likely an hiccup with AWS that we never saw in the years we've used this plugin and haven't seen after the couple of days where we noticed it in our CI. 

We're being optimistic and assuming the hiccup won't come back in a while and therefore deciding against the extra complexity and additional operation in CI.

PR #156, opened 2025-02-20
Prioritise TLDR part over the tree diff

wzieba:
## Description

This PR changes logic of priorities when preparing the "dependency diff commit".

Imagine a PR with huge project dependencies diff and huge build dependencies diff:
- tldr project deps: 1000 chars
- tree project deps: 40 000 chars
- tldr build deps: 500 chars
- tree build deps: 50 000 chars

Limit is 60 000 chars (for simplicity, the real one is 65 400)

### Previously:

Priorities were the same as order of printing. So we would print comment like:

- ✅ TLDR list of project deps (1000 chars)
- ✅ tree of project deps (40 000 chars)
- ❌ TLDR list of build deps
- ❌ tree of build deps (because they both had 41 000 chars, which combining with build dependencies: 41 000 + 50 000 + 500 = 91 500 chars is above limit of the comment PR)

### Now:

TLDR; lists are prioritised. Trees are only an addition "if space allows"

- ✅ TLDR list of project deps (1000 chars)
- ✅ tree of project deps (40 000 chars)
- ✅ TLDR list of build deps (500 chars)
-  ❌ tree of build deps

## Testing

Please run `bats tests/test_craft_comment_with_dependency_diff.bats` - it contains two tests proving that the logic works as intended. 

Also see these PRs:
- https://github.com/Automattic/Matticspace-Mobile/pull/119 (all fits, small diff)
- https://github.com/Automattic/pocket-casts-android/pull/3637 (huge project deps diff, as pocket casts has many modules)

- [ ] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #164, opened 2025-02-27
[AndroidManifest] Merged Manifest Diff

ParaskP7:
Project Thread: paaHJt-7Tl-p2
Required By: [WCAndroid#13637](https://github.com/woocommerce/woocommerce-android/pull/13637)
Related: #152 (`previous workflow`)
(`PR Changes`) Tested By: [WCAndroid#13662](https://github.com/woocommerce/woocommerce-android/pull/13662)

---

## Description

This PR introduces `comment_with_manifest_diff` script that is aiming to help Android product teams to see the merged manifest changes a PR might introduce.

The main problem this script is trying to address is the fact that:
1. Although explicit `AndroidManifest.xml` changes are visible on a PR review, just by looking at the `Files changed` tab for that specific change, this doesn't necessarily mean that the actual, auto-generated, merged `AndroidManifest.xml` file should have the same amount of changes.
2. Sometimes, implicit `AndroidManifest.xml` changes can occur (think new/updated dependency). Those changes are actually invisible for both, to the author and reviewer of that specific PR.

Usage:

```sh
./comment_with_manifest_diff <MODULE> <BUILD_VARIANT>
# Example
./comment_with_manifest_diff app release
```

For more info on the technical/workflow side of things see: p1740739725150069/1740493471.859349-slack-C030U03RC8Y

---

## Test Instructions

You could check this [WCAndroid#13637](https://github.com/woocommerce/woocommerce-android/pull/13637) PR, along with it's test PR ([WCAndroid#13644](https://github.com/woocommerce/woocommerce-android/pull/13644)) and see when and how such comments might appear ([here](https://github.com/woocommerce/woocommerce-android/pull/13644#issuecomment-2690635509) and [here](https://github.com/woocommerce/woocommerce-android/pull/13644#issuecomment-2690636426)):

image

FYI: The above test comment appeared due to an implicit `AndroidManifest.xml` change, adding the [singular](https://support.singular.net/hc/en-us/articles/360037581952-Android-SDK-Integration-Guide) dependency to the project. It would have been impossible for an engineer to know how the merged `AndroidManifest.xml` file got updated without doing such a check manually.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. f6c50f1cd7e2e77c09fac7901df5b0cbee9a8b37


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

bin/comment_with_manifest_diff
```diff
@@ -0,0 +1,140 @@
...
+generate_manifest_comment_body() {
+  cat <<EOF
+
+The following changes in the \`$MODULE\`'s merged \`AndroidManifest.xml\` file were detected (build variant: \`$BUILD_VARIANT\`):
+
+\`\`\`diff
+$(cat "$DIFF_MANIFEST_FILE")
+\`\`\`
+
+Go to \`$MODULE/build/intermediates/merged_manifest/$BUILD_VARIANT/process${BUILD_VARIANT^}MainManifest/AndroidManifest.xml\` and audit the file.
```
wzieba: Maybe we should put both manifests in Artifacts and navigate user to Buildkite instead of local environment?

bin/comment_with_manifest_diff
```diff
@@ -0,0 +1,140 @@
...
+    local header="$1"
+    local content="$2"
+    local github_comment_max_size=65400 # GitHub constraint is 65536, but we decrease this value slightly to reserve for hidden comment_on_pr content and line breaks
+    local warning_message="Content exceeds $github_comment_max_size characters. [Navigate to Buildkite build artifacts](${BUILDKITE_BUILD_URL}#${BUILDKITE_JOB_ID}) to see details."
+
+    local current_size
+    current_size=$(wc -c < "$COMMENT_FILE")
+
+    # Calculate if there's enough space for the content and header
+    if (( current_size + ${#header} + ${#content} > github_comment_max_size - ${#header} - ${#warning_message})); then
```
wzieba: I'm not sure if this condition is okay.

#### Left side

- we have both `current_size` which is `$COMMENT_FILE` (comment body)
- we also have `content` which is result of `generate_manifest_comment_body`

Why do we add here both? If I'm not mistaken, `$COMMENT_FILE` won't ever have content at moment of this condition: it's populated only at the end of `append_content` - before it's always empty, so it's empty at line 124. Or am I missing something?

#### Right side

Why do we substract "header" and "warning message"? I don't think we should: if `content + header` are simply higher than `github_comment_max_size`, we should write warning instead of diff (although I don't think it'll ever happen anyway).

bin/comment_with_manifest_diff
```diff
@@ -0,0 +1,140 @@
...
+      printf "\n%s\n%s\n" "$header" "$warning_message" > "$COMMENT_FILE"
+      return
+    fi
+
+    printf "\n%s\n" "$header" >> "$COMMENT_FILE"
+    printf "%s\n" "$content" >> "$COMMENT_FILE"
```
wzieba: I'd suggest to merge these two: put `header` into `generate_manifest_comment_body` template and pass there an argument. I think it'll simplify the script.

bin/comment_with_manifest_diff
ParaskP7: Yea, true @wzieba, will do, this is indeed a better idea, I was kind of feeling a bit off asking for devs to go to and look at the `release` merged manifest locally, which meant that if they really want to do that they would need to build the `release` variant, else, they would just see the `debug` manifest. 👍

bin/comment_with_manifest_diff
ParaskP7: Short answer @wzieba; I just copy-pasted this logic as is from the (previous) version of the [comment_with_dependency_diff](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/comment_with_dependency_diff) script (before this #156 change), assuming that it works as expected, plus battle tested.

I'll try to dig deeper into the questions you raised and (hopefully) come back with a useful response. 👍

wzieba: I see! The previous version of `comment_with_dependency_diff` worked in a way that it was _appending_ new content (e.g. build env diff) to the existing content (e.g. project diff) _if_ there was space.

As this script doesn't _append_ content (there's simply a single content for a single comment), we don't need this additional behavior.

ParaskP7: True, this script got upgraded with the extra TL;DR content later on, I'll make the change and simplify (remove the additional behavior). But, before I do so, what if we keep this additional behavior should we decide to change the script later on, like after getting some feedback to merge multiple comments (per module) into one. Wdyt? 🤔

bin/comment_with_manifest_diff
ParaskP7: True @wzieba, this can be further simplified, will do! 👍

bin/comment_with_manifest_diff
wzieba: I don't think it's a good idea to have overcomplex code for a reason that one day we might use it. If we'll ever had feedback to merge multiple comments into one, we'll still have to modify this script either way.

ParaskP7: Noted, can't disagree much! 👍 😅

bin/comment_with_manifest_diff
ParaskP7: This is now done: 86c11bf61e3b90d9c9fe436c9bcca6f919fffec7

bin/comment_with_manifest_diff
ParaskP7: This is now done: 8ccd60f89992a2a415975504fb144413bbcdf83d

bin/comment_with_manifest_diff
ParaskP7: This is now done: ddb08a9e5d88b0e352e3678ffce3d4aba3509ad8

PS: I hope this is what you meant. 😊

ParaskP7:
👋 @wzieba I just applied your suggestions (thanks!) and tested them via this [WCAndroid#13662](https://github.com/woocommerce/woocommerce-android/pull/13662) test PR ([comment](https://github.com/woocommerce/woocommerce-android/pull/13662#issuecomment-2700685832) + [comment](https://github.com/woocommerce/woocommerce-android/pull/13662#issuecomment-2700686813)). Could you please take another look and let me know if there is anything else. 🙏 

PS: This [Lint](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1148/canvas?sid=01956620-47fa-4695-b590-708b5e2ba58c) job is failing but for an unrelated reason (p1741176031552239/1741144567.645319-slack-C020Q0Z579V).

bin/comment_with_manifest_diff
wzieba: I thought rather about putting `header` into `generate_manifest_comment_body` and passing it as an argument.

But actually, I see that you could even completely remove `header` variable, simplifying the script

```diff
generate_manifest_comment_body() {

  cat <<EOF
+ ## Project manifest changes for \`MODULE\`

The following changes in the \`$MODULE\`'s merged \`AndroidManifest.xml\` file were detected (build variant: \`$BUILD_VARIANT\`):
\`\`\`diff
$(cat "$DIFF_MANIFEST_FILE")
\`\`\`
Go to https://buildkite.com/$BUILDKITE_PROJECT_SLUG/builds/$BUILDKITE_BUILD_NUMBER/canvas?sid=$BUILDKITE_STEP_ID, click on the \`Artifacts\` tab and audit the files.
EOF
}
```

and then remove all mentions of `header` variable

ParaskP7: Thanks for the clarification on that @wzieba !

FYI: I started doing that, but then I realized that `$header` is used to calculate if there's enough space for the content and header (code snippet below), then print a warning message using that same header. Thus, I then stepped back from doing that and just merged those 2 `printf` lines into one.

```sh
    # Calculate if there's enough space for the content and header
    if (( ${#header} + ${#content} > github_comment_max_size )); then
      printf "\n%s\n%s\n" "$header" "$warning_message" > "$COMMENT_FILE"
      return
    fi
```

Wdyt? 🤔

wzieba: But `content` is result of `generate_manifest_comment_body` - if you'll move `header` to `generate_manifest_comment_body`, then you will count header size in these calculations

wzieba: And for reusing header, you can extract this to other template var and use them directly in comment templates instead of passing them around: I think it's simpler this way

ParaskP7: > But content is result of generate_manifest_comment_body - if you'll move header to generate_manifest_comment_body, then you will count header size in these calculations

Yea, I didn't mention this because of the `content`, but rather this:

```sh
printf "\n%s\n%s\n" "$header" "$warning_message" > "$COMMENT_FILE"
```

> And for reusing header, you can extract this to other template var and use them directly in comment templates instead of passing them around: I think it's simpler this way

After thinking about it, I have the feeling that this is already simple enough, if not simpler than what's suggested (to extract this to other template var, etc).

PS: I also like having this `header` and `content` local variables, feels quite structured and self-explanatory. 🤔

---

If that's not too much of an ask @wzieba, is it okay if we leave that be and more forward, I really think this is quite minor to keep trying simplifying the script? 🙏

PS: Then, when and if we'll need to touch this script again, we can further try making it that much better.

wzieba (approved):

bin/comment_with_manifest_diff
wzieba: Sure 👍

PR #165, opened 2025-02-28
Use snake case in `tests/` file names to be consistent with `bin/`

mokagio:
Follow up to https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/153#discussion_r1967687588

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary. — N.A.


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

Makefile
```diff
@@ -26,4 +26,4 @@ buildkite-plugin-test:
 rspec:
 	@echo ~~~ 🔬 Rspec
 	$(docker_run) ruby:3.2.2 /bin/bash -c \
-	  "gem install --silent rspec -v 3.13.0 && rspec tests/test-that-all-files-are-executable.rb"
+	  "gem install --silent rspec -v 3.13.0 && rspec tests/test_that_all_files_are_executable.rb"
```
mokagio, line 29: I was going to DRY this as a tidy up before renaming, but I wondered why we constrained both RuboCop and RSpec to a single file? 

I'd say we'd be better off removing the path and letting both run on all the files they can run, right?

AliSoftware: Good question. My guess (not verified) is that other Ruby files were not adhering to our Rubocop rules when we introduced them so maybe we decided to postpone the adoption to all the files in the repo but instead only to a specific file? Though it that were the case I'd have preferred for us to use a `.rubocop-todo` file.

As for why we constrained the test to a single file, I think there's no specific reason to it. It's probably because that was the only test written in Ruby at the time we introduced it, while all other tests were written in `.bats` format, which are run by the `make buildkite-plugin-test` task itself (rather than run by us explicitly separately). So at the time this was the only test that we had to run manually/separately. But even then, we should probably have gone with `tests/*.rb` rather than a specific file.

> I'd say we'd be better off removing the path and letting both run on all the files they can run, right?

💯

AliSoftware (approved):

Makefile
mokagio: > As for why we constrained the test to a single file, I think there's no specific reason to it.

I tried to verify this and discovered that, out of the box, `rspec` will not run any file in `tests/`. 

image

I suspect that's due:

1. There not being any `spec` folder
2. None of the files in `tests/` being suffixed with `_spec`

This can be verified by:

1. Running `rspec` in the root of the project (screenshot above)
2. Renaming `tests/` into `spec/` — Still no tests run
3. Renaming the only Ruby spec to something suffixed with `_spec` — It finally runs

If we don't want to rename the folder or files, we can use:

```
rspec tests/test*.rb
```

I think having the `test` prefix in the pattern will help us in the future if we decide to have some Ruby helper.

Anyway, I'm working on a PR that adds a bit more Ruby support and we can talk about it in there.

mokagio: > > I'd say we'd be better off removing the path and letting both run on all the files they can run, right?
>
> 💯

For RuboCop, see https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/167

PR #167, opened 2025-03-02
Set up Ruby version and RuboCop

mokagio:
_This all started with me wanting to run RuboCop in the `Makefile` on the whole project, without specifying each Ruby file manually..._

Adds a `.ruby-version` files, setting the Ruby version for the project to ~~3.3.4, which is the latest coming pre installed in our macOS VMs~~ 3.2.2 which is the same the `linter` queue uses.

Also adds a bare bone `.rubocop.yml`, so we can run `rubocop` on the whole project, without constraining it to a single file, as noticed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/165#discussion_r1975146249

Finally, updates Make to run RuboCop in Docker via Bundler — this I think make sense even though it's not the way it was setup, so maybe I'm missing something... ?

---


mokagio:
Reverted back to draft because the `linter` queue does not have Ruby 3.3.4 and does not install it 🤔 

https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1134/canvas?jid=019558b1-f339-4b45-b2d4-c5f1c57da276

image

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

.ruby-version
```diff
@@ -0,0 +1 @@
+3.2.2
```
mokagio, line 1: I wanted to use 3.3.4 (latest pre-installed in macOS VMs) but it failed to run on the `linter` queue, see https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1134/canvas?jid=019558b1-f339-4b45-b2d4-c5f1c57da276

Decided to stick with the supported 3.2.2 and look into the version issue at a  later moment.

Makefile
```diff
@@ -1,10 +1,12 @@
 .DEFAULT_GOAL := default
 
+RUBY_VERSION := 3.3.4
+
 default: lint test
 lint: buildkite-plugin-lint rubocop shellcheck
 test: buildkite-plugin-test rspec
 
-docker_run := docker run -t --rm -v "${PWD}"/:/plugin:ro -w /plugin
+docker_run := docker run -t --rm -v "${PWD}"/:/plugin -w /plugin
```
mokagio, line 9: The `ro` there made the mounted file system read-only which prevent using `bundle exec`. 

I don't know if there was a particular requirement for the file system to be read-only. It was like that from the very first definition, see 0b77225bae7ee6b2a6f15c50d7a942b45ea33c13

As CI shows, using a read-write version does not result in failures, so I think we can continue with it?

Makefile
```diff
@@ -1,10 +1,12 @@
 .DEFAULT_GOAL := default
 
+RUBY_VERSION := 3.2.2
```
mokagio: Happy to read from `.ruby-version` to have a single source of truth. 

```suggestion
RUBY_VERSION := $(shell cat .ruby-version)
```

I recall some conversation somewhere about wanting to retain control in make vs local settings, but I couldn't find it — and I bet it was in the context of a particular setup because personally I think the more homogeneity we can have, the better. 

Although, one could argue that since we use Docker via Make here one shouldn't even need to set `.ruby-version` because all the Ruby tools should run via `make` instead 🤔 

Still, I think having an explicitly set version like that is handy and idiomatic. What do you think?

AliSoftware: I think the conversation you were remembering might be from when we adopted a similar "single source of truth for Ruby version" in our `buildkite-ci` monorepo, and the rationale / debate for if that was the right move in that context was because this was a monorepo, while each different sub-folders and tools in that repo might still consider using different Ruby versions, or have their ruby version defined at the sub-folder level rather than the top-level of the monorepo…?

Anyway, I think in the context of this `a8c-ci-toolkit` repo, this makes sense to enforce the single source of truth and enforce the local runs to use the same version as CI will in this case 👍  So personally I'd apply your suggestion to have the `Makefile` read it from the `.ruby-version` file 👍

.bundle/config
```diff
@@ -2,3 +2,4 @@
 BUNDLE_PATH: "vendor/bundle"
 BUNDLE_JOBS: "16"
 BUNDLE_SPECIFIC_PLATFORM: "false"
+BUNDLE_FORCE_RUBY_PLATFORM: "true"
```
AliSoftware, line 5: 👍

.ruby-version
AliSoftware: cc @iangmaia wrt to considering to install Ruby `3.3.4` on the `linter` agent to make it available there when requested? Or should we make the linter agent auto-install the requested Ruby version (via `rbenv install`, potentially preceded by an update of `ruby-build`) automatically? 🤔

AliSoftware (approved):
LGTM + I'd agree the `Makefile` should read `RUBY_VERSION` from the `.ruby-version` file as you suggested in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/167#discussion_r1977313850 👍

.ruby-version
iangmaia: Hmm interesting thought. I guess we could theoretically make it install versions on demand... but in reality it's a transient machine, so it would spend some time downloading / installing versions (which isn't a super light-weight process) and that would be soon discarded, and that would have an impact on each agent. It feels more like a system setup component, "heavier" than just simple dependencies.

I'll add a task for me to add a few more recent Ruby versions (there's also `3.3.7`, `3.4.1`...).
Having the Ruby versions we support aligned across our agents somehow would be a nice improvement, though 🤔

AliSoftware: > but in reality it's a transient machine

wait hold on a minute I thought the whole point of the `linter` agent was for it to be a long-living instance so that it could pick up jobs super fast? (And in fact it being a long-running agent is another reason why we worked on locking this agent to only be able to run a specific allowlist of commands, for security)?

PR #177, opened 2025-03-18
[Diff] Enhance Comment with Dependency/Manifest Diff with Git Fetch/Pull

ParaskP7:
Project Thread: paaHJt-7Tl-p2
Tested By:
- Manifest Diff:
    - GHE: [TBAndroid#24349](https://github.tumblr.net/Tumblr/android/pull/24349) + [TBAndroid#24350](https://github.tumblr.net/Tumblr/android/pull/24350)
    - GH: [WCAndroid#13791](https://github.com/woocommerce/woocommerce-android/pull/13791) + [WCAndroid#13792](https://github.com/woocommerce/woocommerce-android/pull/13792)
- Dependency Diff:
    - GHE: [TBAndroid#24353](https://github.tumblr.net/Tumblr/android/pull/24353)
    - GH: [WCAndroid#13791](https://github.com/woocommerce/woocommerce-android/pull/13791) + [WCAndroid#13792](https://github.com/woocommerce/woocommerce-android/pull/13792)

---

## Description

This PR is about enhancing [comment_with_manifest_diff](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/comment_with_manifest_diff) (but also [comment_with_dependency_diff](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/bin/comment_with_dependency_diff)) so that it works as expected with project like [TBAndroid](https://github.tumblr.net/Tumblr/android) where the base branch is **NOT** always up-to-date and at all times, just like it happens on all other projects (see p1742298963526779/1742288408.868439-slack-C0211NP7Y5U).

---

## Test Instructions

- Manifest Diff:

You could check this [TBAndroid#24349](https://github.tumblr.net/Tumblr/android/pull/24349) PR, along with it's test PR [TBAndroid#24350](https://github.tumblr.net/Tumblr/android/pull/24350) and see when and how such [comments](https://github.tumblr.net/Tumblr/android/pull/24350#issuecomment-696383) might appear:

image

- Dependency Diff:

You could check [TBAndroid#24353](https://github.tumblr.net/Tumblr/android/pull/24353) test PR and see when and how such [comment](https://github.tumblr.net/Tumblr/android/pull/24353#issuecomment-696389) might appear: 

image

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #178, opened 2025-03-18
Release 5.3.0

ParaskP7:
Release process PR following [releasing instructions](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin?tab=readme-ov-file#releasing).

Related PRs:

- #177 

FYI: Will admin-merge once CI allows (unless already approved) and proceed with creating a GitHub release from it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,6 +46,12 @@ _None._
 
 ### Internal Changes
 
+_None._
+
+## 5.3.0
+
+### Internal Changes
+
 - Update `comment_with_manifest_diff` and `comment_with_dependency_diff` commands with extra `git fetch/pull` calls to make both of them work as expected on all CI environments. [#177]
 
 ## 5.2.0
--- a/README.md
+++ b/README.md
@@ -26,11 +26,11 @@ steps:
       restore_cache $(hash_file package-lock.json)
 
     plugins:
-      - automattic/a8c-ci-toolkit#5.2.0:
+      - automattic/a8c-ci-toolkit#5.3.0:
           bucket: a8c-ci-cache # optional
 ```
 
-Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `5.2.0`.
+Don't forget to verify what [the latest release](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/releases/latest) is and use that value instead of `5.3.0`.
 
 ## Configuration
 


CHANGELOG.md
```diff
@@ -46,6 +46,12 @@ _None._
 
 ### Internal Changes
 
+_None._
+
+## 5.3.0
+
+### Internal Changes
```
wzieba, line 53: I'm not sure if it's internal change as it fixes something for toolkit consumers. I interpret any "internal" change as something that consumer can't notice. But it's a minor

wzieba (approved):

CHANGELOG.md
ParaskP7: Yea, I know, I was confused about which too use as well, bug fixes or internal changes, as on most projects it is not a bug, but making this change enabled these scripts on that one specific project... 🤷

AliSoftware: Agreed with Wojtek.
I'd qualify this fix as Minor change (aka `patch` version bump, as per SemVer conventions, because you're fixing a bug, not introducing a new feature or a breaking change)

Internal is more for if we fixed, say, `.buildkite/*.yml` pipelines used to run the CI checks on this repo, or add more `.cursor/rules` to this repo, or fixed the tests… anything not visible to toolkit consumers

Here it's clearly a Minor / BugFix. So technically this also means the version bump should be `5.2.1` too, not `5.3.0`

ParaskP7: Gotcha, I'll bear that in mind for next time, although again, I am not sure about it being only a bug fix, or a minor release on `5.2.0` because this `5.2.0` wasn't about that. I was thinking of it as an internal improvement since we are now enhancing these diff scripts to be able to be used on a different CI environment, thus improving our tooling, not fixing it, but again, I get how this can be interpreted differently... 🤷

AliSoftware: > I was thinking of it as an internal improvement since we are now enhancing these diff scripts to be able to be used on a different CI environment

It was still a bug (and thus patch version bump): it's just that you never encountered that bug until you tested the utility in more CI environments.

> thus improving our tooling, not fixing it

And even if you don't want to consider it a bug but rather an improvement our our tooling, then that's an improvement / new feature, so that'd be a minor version bump under SemVer instead of a patch version bump, but still not an internal improvements.

> or a minor release on 5.2.0 because this 5.2.0 wasn't about that.

Wether `5.2.0` was included the thing that you're fixing or if that thing you're fixing was introduced in an earlier version or not has nothing to do with that. SemVer rules about how to bump versions is just if what you're introducing is a bugfix (patch version bump), or a new feature (minor version bump) or a breaking change on an existing feature (major version bump). It could be a bugfix on a feature that we introduced in `1.2.0` ages ago but only discovered the bug today and that wouldn't make a difference.

The key information that SemVer provides is for the consumer to know if when they update to this new version will they have to make changes to their own codebase to adopt the new version (necessary changes in case of major version bump = breaking changes, or optional changes in case of minor version bump that might maybe introduce a new feature that could allow simplifying the codebase and logic that the consumer was using with the new feature making it simpler) or not (it just fixes previous bugs, no change needed on the consumer's codebase, but still better update to benefit from the bugfix in).

ParaskP7: > It was still a bug (and thus patch version bump): it's just that you never encountered that bug until you tested the utility in more CI environments.

👍 

> And even if you don't want to consider it a bug but rather an improvement our our tooling, then that's an improvement / new feature, so that'd be a minor version bump under SemVer instead of a patch version bump, but still not an internal improvements.

True, I could have used it as an improvement via `New Features`, it is just that we don't have an `Improvements` category and I felt it be closed to `Internal Changes` versus `New Features`... 🤷 

Question (❓): Maybe we should add an (external) `Improvements` category? 🤔

> Wether 5.2.0 was included the thing that you're fixing or if that thing you're fixing was introduced in an earlier version or not has nothing to do with that. SemVer rules about how to bump versions is just if what you're introducing is a bugfix (patch version bump), or a new feature (minor version bump) or a breaking change on an existing feature (major version bump). It could be a bugfix on a feature that we introduced in 1.2.0 ages ago but only discovered the bug today and that wouldn't make a difference.

👍 

> The key information that SemVer provides is for the consumer to know if when they update to this new version will they have to make changes to their own codebase to adopt the new version (necessary changes in case of major version bump = breaking changes, or optional changes in case of minor version bump that might maybe introduce a new feature that could allow simplifying the codebase and logic that the consumer was using with the new feature making it simpler) or not (it just fixes previous bugs, no change needed on the consumer's codebase, but still better update to benefit from the bugfix in).

👍

AliSoftware: If that helps we could rename the "New Features" section from the template to "New Features & Improvements"?

Though tbh those section titles are what we use in other repos for which we follow SemVer relatively strictly ([like `release-toolkit`](https://github.com/wordpress-mobile/release-toolkit/blob/trunk/CHANGELOG.md) amongst others).
So basically I think what makes the most sense is to think of those sections as the equivalent types of bumps per official SemVer rules. Also because that's exactly what the logic of our release scripts for doing new versions is also using to determine which type of version bump it should do

 - `## Breaking Changes` => Major version bump under SemVer rules
 - `## New Features` => Minor version bump under SemVer rules
 - `## Bug Fixes` => Patch version bump under SemVer rules
 - `## Internal Changes` => does not impact version bump SemVer rules because no impact at all to consumers. This is only for changes like updating the CI pipelines, or GitHub repo settings, or fixing our documentation in the repo about on how to do a new release, stuff like that

ParaskP7: > If that helps we could rename the "New Features" section from the template to "New Features & Improvements"?

I would be in favor of that, yes. 🤷

> `## Breaking Changes` => Major version bump under SemVer rules
> `## New Features` => Minor version bump under SemVer rules
> `## Bug Fixes` => Patch version bump under SemVer rules
> `## Internal Changes` => does not impact version bump SemVer rules because no impact at all to consumers. This is only for changes like updating the CI pipelines, or GitHub repo settings, or fixing our documentation in the repo about on how to do a new release, stuff like that

I like this, thanks! 👍

PR #179, opened 2025-04-23
`pr_changed_files` improvements to run properly on all agents

iangmaia:
I've noticed a few failures when running `pr_changed_files` on `default`/`android`, `APPIUM`, `tumblr-metal` and `tumblr-android` and updated a few things along the way:

- `tumblr-metal` used an old `git` version and it was failing completely when running the `git diff` command. I've fixed it in the machine itself, so there are no changes for it here
- Update `git diff` call to `merge-base` using `origin/`
- I've updated `pr_changed_files` tests to run on other agents instead of only `mac`. Unfortunately, I couldn't get them running on `tumblr-android` as it is a unclustered agent and it doesn't seem to work on clustered pipelines (the upload immediately fails when running using the agent).
- I've implemented a quick experiment breaking down the main pipeline into smaller focused ones given we had a lot of tests in a single file. So now we have `.buildkite/pipeline-bash-tests.yml` and `.buildkite/pipeline-windows-tests.yml` besides the regular `.buildkite/pipeline.yml` file.

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

AliSoftware:
> tumblr-metal used an old git version and it was failing completely when running the git diff command. I've fixed it in the machine itself, so there are no changes for it here

The fact that `tumblr-metal`'s system is becoming pretty old, and using a quite uncommon and special Linux distribution ("Scientific") which makes it tricky to update to latest tooling, is another argument for tackling AINFRA-132 btw.

bin/pr_changed_files
```diff
@@ -109,11 +109,12 @@ while [[ "$#" -gt 0 ]]; do
     esac
 done
 
+declare -a changed_files=()
```
AliSoftware: Just to note that I double checked with ChatGPT that this syntax was compatible with all the versions of `bash` we use in our various agents.
 
 - In particular, while `declare -A VAR` (declaring an associative array) was only introduced in bash 4.0, `declare -a VAR` (declaring an indexed array) has apparently existed since `bash 2.0`, so we should be good in terms of compatibility on that front
 - `declare -a VAR=()`, i.e. both declaring and assigning the array in a single line, is apparently not always supported by all variants and subversions of `bash 2.0`

I _think_ that all of our CI agents are using at least `bash 3.0` and none are still on `bash 2.0`, so we should be good with declaring + assigning in a single expression with `declare -a changed_files=()` like here. But if you want to be extra safe in terms of compatibility, you could also consider breaking this down in 2 statements, `declare -a changed_files` and `changed_files=()` 🤷 

image

AliSoftware (approved):

bin/pr_changed_files
iangmaia: > declare -a VAR=(), i.e. both declaring and assigning the array in a single line, is apparently not always supported by all variants and subversions of bash 2.0

Good point. Using `declare` was a suggestion from Claude to deal with an issue in one of the agents, and I liked the extra clarity it added. Breaking it down to declaration and initialization for added compatibility sounds good to me 👍

PR #186, opened 2025-07-23
Install Python during windows setup and add new Visual Studio Build tools components 

iangmaia:
To fix the error seen in https://github.com/Automattic/studio/pull/1594 ([build](https://github.com/Automattic/studio/pull/1594)), this PR implemented:
- Python installation during the Windows setup so that it can be used by npm install scripts
- Addition of native compilation components to Visual Studio Build Tools

Being tested via https://github.com/Automattic/studio/pull/1608

---

### DIFF
--- a/bin/install_windows_10_sdk.ps1
+++ b/bin/install_windows_10_sdk.ps1
@@ -4,8 +4,8 @@
 # The list of valid component ids can be found at
 # https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
 #
-# This script installs components required for Node.js native module compilation on Windows:
-# - Windows 10 SDK (configurable version)
+# This script installs the Windows 10 SDK by default.
+# Optionally, it can also install native compilation tools for building native modules:
 # - MSVC v143 compiler toolset (x64/x86)
 # - CMake tools for Visual Studio
 #
@@ -14,7 +14,8 @@
 #   20348
 
 param (
-  [switch]$DryRun = $false
+  [switch]$DryRun = $false,
+  [switch]$InstallNativeCompilationTools = $false
 )
 
 # Stop script execution when a non-terminating error occurs
@@ -60,6 +61,12 @@ if ($allowedVersions -notcontains $windows10SDKVersion) {
 
 Write-Output "Will attempt to set up Windows 10 ($windows10SDKVersion) SDK and Visual Studio Build Tools..."
 
+if ($InstallNativeCompilationTools) {
+  Write-Output "Native compilation tools requested. Will also install MSVC compiler toolset and CMake tools."
+} else {
+  Write-Output "Installing Windows 10 SDK only. Use -InstallNativeCompilationTools to include MSVC compiler and CMake tools."
+}
+
 if ($DryRun) {
   Write-Output "Running in dry run mode, finishing here."
   exit 0
@@ -81,15 +88,25 @@ If (-not (Test-Path $buildToolsPath)) {
   Write-Output "Successfully downloaded Visual Studio Build Tools at $buildToolsPath."
 }
 
-# Install the Windows SDK and other required components for Node.js native module compilation
-Write-Output "~~~ Installing Visual Studio Build Tools with Node.js native module support..."
+# Install the Windows SDK and (optionally) native compilation tools
+if ($InstallNativeCompilationTools) {
+  Write-Output "~~~ Installing Visual Studio Build Tools with native compilation tools..."
+} else {
+  Write-Output "~~~ Installing Visual Studio Build Tools with Windows 10 SDK..."
+}
 
-# Define components needed for Node.js native module compilation
+# Define base components (always installed)
 $components = @(
-  "Microsoft.VisualStudio.Component.Windows10SDK.$windows10SDKVersion",
-  "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",  # MSVC v143 compiler toolset
-  "Microsoft.VisualStudio.Component.VC.CMake.Project"   # CMake tools for native modules
+  "Microsoft.VisualStudio.Component.Windows10SDK.$windows10SDKVersion"
 )
+
+# Add native compilation tools if requested
+if ($InstallNativeCompilationTools) {
+  $components += @(
+    "Microsoft.VisualStudio.Component.VC.Tools.x86.x64",  # MSVC v143 compiler toolset
+    "Microsoft.VisualStudio.Component.VC.CMake.Project"   # CMake tools for native modules
+  )
+}
 
 $argumentList = "--quiet --wait"
 foreach ($component in $components) {
--- a/bin/install_windows_10_sdk.ps1
+++ b/bin/install_windows_10_sdk.ps1
@@ -4,6 +4,11 @@
 # The list of valid component ids can be found at
 # https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
 #
+# This script installs components required for Node.js native module compilation on Windows:
+# - Windows 10 SDK (configurable version)
+# - MSVC v143 compiler toolset (x64/x86)
+# - CMake tools for Visual Studio
+#
 # Example:
 #
 #   20348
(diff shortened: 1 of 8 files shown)


bin/install_windows_10_sdk.ps1
```diff
@@ -55,6 +61,12 @@ if ($allowedVersions -notcontains $windows10SDKVersion) {
 
 Write-Output "Will attempt to set up Windows 10 ($windows10SDKVersion) SDK and Visual Studio Build Tools..."
 
+if ($InstallNativeCompilationTools) {
+  Write-Output "Native compilation tools requested. Will also install MSVC compiler toolset and CMake tools."
```
```suggestion
  Write-Output "Native compilation tools requested. Will also install MSVC compiler toolset and CMake tools in addition to the Windows 10 SDK."
```

bin/install_windows_10_sdk.ps1
```diff
@@ -4,12 +4,18 @@
 # The list of valid component ids can be found at
 # https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
 #
+# This script installs the Windows 10 SDK by default.
+# Optionally, it can also install native compilation tools for building native modules:
+# - MSVC v143 compiler toolset (x64/x86)
+# - CMake tools for Visual Studio
+#
```
AliSoftware: You should move the "example" part of the header closer to the paragraphs that are speaking about the value expected in the `.windows-10-sdk-version` file, otherwise it makes this all quite confusing.

The solution to avoid the confusion might actually just be to move those new lines you added to line 3, before the "The expected .windows-10-sdk-version format …" paragraph instead of after.

bin/install_windows_10_sdk.ps1
```diff
@@ -76,11 +88,36 @@ If (-not (Test-Path $buildToolsPath)) {
   Write-Output "Successfully downloaded Visual Studio Build Tools at $buildToolsPath."
 }
 
-# Install the Windows SDK and other required components
-Write-Output "~~~ Installing Visual Studio Build Tools..."
+# Install the Windows SDK and (optionally) native compilation tools
+if ($InstallNativeCompilationTools) {
+  Write-Output "~~~ Installing Visual Studio Build Tools with native compilation tools..."
```
```suggestion
  Write-Output "~~~ Installing Visual Studio Build Tools and native compilation tools with Windows 10 sdk..."
```

Or at least a wording that still includes Windows 10 SDK probably because that one is always installed regardless of the value of the flag, so having it mentioned in one case but not the other is confusing and might make it look like we don't install the Win10 SDK with the flag set…

bin/install_windows_10_sdk.ps1
AliSoftware: You should also probably mention the name of the `-InstallNativeCompilationTools` flag/param in that message, like the `-SkipWindows10SDKInstallation` one is mentioned in the docs for the `prepare_….ps1` script

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -3,18 +3,24 @@
...
+#  - Download Code Signing Certificates(2)
+#  - Install the Windows 10 SDK if it detected a `.windows-10-sdk-version` file(3)
+#  - Optionally install native compilation tools (MSVC compiler and CMake)(4)
 #
-# (1) The certificate it installs is stored in our AWS SecretsManager storage (`windows-code-signing-certificate` secret ID)
-# (2) You can skip the Windows 10 SDK installation regardless of whether `.windows-10-sdk-version` is present by calling the script with `-SkipWindows10SDKInstallation`.
+# (1) Python is NOT installed by default. You can install Python by calling the script with `-InstallPython`.
+# (2) The certificate it installs is stored in our AWS SecretsManager storage (`windows-code-signing-certificate` secret ID)
+# (3) You can skip the Windows 10 SDK installation regardless of whether `.windows-10-sdk-version` is present by calling the script with `-SkipWindows10SDKInstallation`.
+# (4) Native compilation tools are NOT installed by default. You can install them by calling the script with `-InstallNativeCompilationTools`.
```
AliSoftware:
```suggestion
# (4) Native compilation tools are NOT installed by default. You can install them by calling the script with `-InstallNativeCompilationTools`. This requires the Windows 10 SDK installation to NOT be skipped, as those are installed in tandem.
```

Either that, or handle the case via some test in the code that warns if someone tries to call with `-SkipWindows10SDKInstallation -InstallNativeCompilationTools` so that they're not surprised that the compilation tools are not installed if Win10SDK is skipped.

Or, alternatively update the script to not exit early if `SkipWindows10SDKInstallation` is set (or if `.windows-10-sdk-version` is not set which would also skip it), but instead still call `vs_buildtools.exe` to install just the NativeCompilationTools without the SDK in such a case where that's what the caller demanded with `-InstallNativeCompilationTools -SkipWindows10SDKInstallation`?

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -93,15 +107,6 @@ if ($developerMode.State -eq 'Enabled') {
...
 
-Write-Output "--- :lock_with_ink_pen: Download Code Signing Certificate"
-$certificateBinPath = "certificate.bin"
-$EncodedText = aws secretsmanager get-secret-value --secret-id windows-code-signing-certificate `
-  | jq -r '.SecretString' `
-  | Out-File $certificateBinPath
-$certificatePfxPath = "certificate.pfx"
-certutil -decode $certificateBinPath $certificatePfxPath
-Write-Output "Code signing certificate downloaded at: $((Get-Item $certificatePfxPath).FullName)"
-
```
AliSoftware: I'm guessing there's a reason why you moved this part of the code to the end after the Win10SDK installation instead of before, though I'm not sure what it is.

But did you account for the fact that if `-SkipWindows10SDKInstallation` is set then the script will exit early (line L113/R118), which means that the code signing certificate will also not be installed if that flag is set? Or was it actually the whole point of moving the certificate installation code after that early exit after all?

iangmaia: Yeah, I'm gonna revert that. I'll push a few changes in a bit.

bin/install_windows_10_sdk.ps1
iangmaia: 👍 Good point. Done in 3baed42b74adf9b6bfde3b1897db2cfa5e166765.

iangmaia:
@AliSoftware I think I addressed all the points. I struggled a bit with the unit tests later but everything seems to be finally working now 😌

bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -93,14 +115,23 @@ if ($developerMode.State -eq 'Enabled') {
...
 
-Write-Output "--- :lock_with_ink_pen: Download Code Signing Certificate"
-$certificateBinPath = "certificate.bin"
-$EncodedText = aws secretsmanager get-secret-value --secret-id windows-code-signing-certificate `
-  | jq -r '.SecretString' `
-  | Out-File $certificateBinPath
-$certificatePfxPath = "certificate.pfx"
-certutil -decode $certificateBinPath $certificatePfxPath
-Write-Output "Code signing certificate downloaded at: $((Get-Item $certificatePfxPath).FullName)"
+if ($env:SKIP_CERTIFICATE_DOWNLOAD -eq "true") {
```
AliSoftware, line 118: I'm guessing the reason why this is controlled by an env var and not a flag (aka `param { [switch] … }`) like the others is because the skipping is only used for the tests so far?

But that made me wonder if the studio jobs running the Unit Tests might also not need the cert and want to skip it 🤔 💭

AliSoftware (approved):

bin/prepare_windows_host_for_app_distribution.ps1
iangmaia: >  is only used for the tests so far?

Yes, that's the case so far. Also, this seemed more like a configuration rather than an API argument so the env var felt more appropriate.

> if the studio jobs running the Unit Tests might also not need the cert

Good question. Imagining it still compiles the tested app in some way, perhaps they're still needed? But we can definitely try 👍

PR #188, opened 2025-08-06
Only use Windows SDK 19041

twstokes:
Closes AINFRA-1096

Windows 10 SDK 20348 doesn't appear available for download via the VS code bootstrapping tool, but 
19041 still works.

### DIFF
--- a/bin/install_windows_10_sdk.ps1
+++ b/bin/install_windows_10_sdk.ps1
@@ -11,7 +11,7 @@
 #
 # Example:
 #
-#   20348
+#   19041
 
 param (
   [switch]$DryRun = $false,
@@ -25,14 +25,7 @@ Write-Output "--- :windows: Installing Windows 10 SDK and Visual Studio Build To
 
 # See list at https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022
 $allowedVersions = @(
-  "20348",
-  "19041",
-  "18362",
-  "17763",
-  "17134",
-  "16299",
-  "15063",
-  "14393"
+  "19041"
 )
 
 $windowsSDKVersionFile = ".windows-10-sdk-version"
--- a/.buildkite/pipeline-windows-tests.yml
+++ b/.buildkite/pipeline-windows-tests.yml
@@ -16,7 +16,7 @@ steps:
     steps:
       - label: ":windows: install_windows_10_sdk Tests - Version file with valid format and version"
         command: |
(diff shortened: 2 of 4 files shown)


bin/install_windows_10_sdk.ps1
```diff
@@ -25,14 +25,7 @@ Write-Output "--- :windows: Installing Windows 10 SDK and Visual Studio Build To
...
 $allowedVersions = @(
-  "20348",
-  "19041",
-  "18362",
-  "17763",
-  "17134",
-  "16299",
-  "15063",
-  "14393"
+  "19041"
```
AliSoftware, line 28: ❓  Are we sure all those other versions should be removed from the allowlist, not just 20348?

In other words, have you tested each single one of those and removed them all because you validated that `19041` was indeed the only one remaining valid? Or did you remove them all just because you considered it was not worth the effort for us to allow/support those older versions and didn't want to bother?

(I'm fine with either, just curious)

AliSoftware (approved):
Approving to unblock, but left a question.

I guess the plan is to soon support Win11 SDK anyway (in a separate PR with some refactor etc), so it's probably ok if we only support `19041` until then, just curious if that was intentional or not.

bin/install_windows_10_sdk.ps1
iangmaia: Maybe we could go for something like:

```
$allowedVersions = @(
    # Windows 11 SDK
    "22621",
    "22000", 
    # Windows 10 SDK
    "19041",
    "18362"
)
```

As these are the Windows SDK versions that are [included](https://learn.microsoft.com/en-us/visualstudio/install/workload-component-id-vs-build-tools?view=vs-2022) in the workload "WinUI application development build tools".

iangmaia: (I know Windows 11 choices would make it fail as is but it's something to iterate on later)

twstokes: This was mainly because I don't think they're available either judging by only `19041` showing when searching with the GUI. I also don't think it's worth the effort to support them, although going forward I can see where we'd keep `19041` and then start adding Windows 11 SDKs.

Screenshot 2025-08-05 at 20 40 50

iangmaia (approved):
👍

bin/install_windows_10_sdk.ps1
iangmaia: Ahh got it 👍 makes sense.

PR #189, opened 2025-08-22
Update `run_swiftlint` to allow for custom .swiftlint config files

iangmaia:
Related to https://github.com/Automattic/buildkite-ci/pull/707, this PR updates the `run_swiftlint` script to allow the `--config` parameter specifying a SwiftLint config file.

---

### DIFF
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -34,7 +34,6 @@ fi
 # If no config files specified, use default
 if [[ ${#CONFIG_FILES[@]} -eq 0 ]]; then
   CONFIG_FILES=(".swiftlint.yml")
-  SWIFTLINT_ARGUMENTS+=(--config ".swiftlint.yml")
 fi
 
 # Validate all config files exist
--- a/bin/run_swiftlint
+++ b/bin/run_swiftlint
@@ -3,6 +3,7 @@
 echo "--- :swift: Running SwiftLint"
 
 SWIFTLINT_ARGUMENTS=(--quiet --reporter relative-path)
+CONFIG_FILE=".swiftlint.yml"
 
 while [[ "$#" -gt 0 ]]; do
   case $1 in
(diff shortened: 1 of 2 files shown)


bin/run_swiftlint
```diff
@@ -3,32 +3,47 @@
...
       SWIFTLINT_ARGUMENTS+=("$1")
       shift
       ;;
+    --config)
+      if [[ -z "$2" || "$2" =~ ^- ]]; then
+        echo "Error: --config requires a config file path"
+        echo "Usage: $0 [--strict | --lenient] [--config CONFIG_FILE]"
+        exit 1
+      fi
+      CONFIG_FILE="$2"
```
AliSoftware: Should we allow multiple occurrences of the `--config` flag like the `SwiftLint` binary does?

I.e. make `CONFIG_FILE=()` be initialized as an array and do `CONFIG_FILE+=` here?

Otherwise invoking `run_SwiftLint` with multiple `--config A --config B`, while valid with `swiftlint`, will make `run_SwiftLint` only consider the _last_ one and ignore the other ones, which might be a confusing discrepancy with the official `swiftlint` call…

AliSoftware (approved):
Approving to unblock, with an open question

bin/run_swiftlint
iangmaia: I wasn't considering SwiftLint could deal with multiple `--config` -- nice one, added support for it on 0b5fdc8a25600e39b2fc754797c94a00489f7472

bin/run_swiftlint
```diff
@@ -3,26 +3,53 @@
...
-  echo "Usage: $0 [--strict | --lenient]"
+  echo "Usage: $0 [--strict | --lenient] [--config CONFIG_FILE]..."
   exit 1
 fi
 
-SWIFTLINT_VERSION="$(<.swiftlint.yml awk '/^swiftlint_version: / {print $2}')"
+# If no config files specified, use default
+if [[ ${#CONFIG_FILES[@]} -eq 0 ]]; then
+  CONFIG_FILES=(".swiftlint.yml")
+  SWIFTLINT_ARGUMENTS+=(--config ".swiftlint.yml")
```
AliSoftware: I wonder if that `SWIFTLINT_ARGUMENTS+=(--config ".swiftlint.yml")` is strictly necessary, since that's already handled by `swiftlint` itself that if no `--config …` is passed it uses `.swiftlint.yml` by default… so might as well not pass that `--config` argument in those cases?

iangmaia: Hmm not necessary indeed 🤔 updated on aaa2860d3881e0ffac873995ee4c7e44e3047ce5

AliSoftware (approved):

PR #195, opened 2025-11-15
Add `git-conceal-unlock` wrapper

AliSoftware:
Closes [AINFRA-1606](https://linear.app/a8c/issue/AINFRA-1606/finalize-implementation-of-git-conceal)

## What?

This PR adds `bin/git-conceal-unlock` helper which takes care of downloading `git-conceal` (using the `install.sh` script from https://github.com/Automattic/git-conceal/pull/3) if it's not installed on the CI machine first, then executing `git-conceal unlock env:${1:-GIT_CONCEAL_SECRET_KEY}` to unlock the repo.

## Testing

This has been tested as part of https://github.com/woocommerce/woocommerce-android/pull/14979

## Changelog

### DIFF
--- a/bin/git-conceal-unlock
+++ b/bin/git-conceal-unlock
@@ -16,15 +16,13 @@ env_var_name=${1:-GIT_CONCEAL_SECRET_KEY}
 
 # If installed in the $PATH, execute it (replacing this current process)
 if command -v git-conceal &> /dev/null; then
-    git-conceal unlock "env:${env_var_name}"
-    exit $?
+    exec git-conceal unlock "env:${env_var_name}"
 fi
 
 INSTALL_DIR="${PWD}"
 # If already installed in INSTALL_DIR, execute it (replacing this current process)
 if [[ -x "${INSTALL_DIR}/git-conceal" ]]; then
-    "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
-    exit $?
+    exec "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
 fi
 
 # Otherwise, install it locally and execute it
--- a/bin/git-conceal-unlock
+++ b/bin/git-conceal-unlock
@@ -1,0 +1,35 @@
+#!/bin/bash
+
+# Description:
+#   A wrapper script to run `git-conceal unlock` on CI
+#   See https://github.com/Automattic/git-conceal
+#
+#   The wrapper automatically runs the `install.sh` script of `git-conceal`
+#   if it's not already installed locally before running `git-conceal unlock env:…`.
+#
+# By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
+# You can override it by passing the environment variable name as the first argument.
+#
+
+set -euo pipefail
+
+env_var_name=${1:-GIT_CONCEAL_SECRET_KEY}
+
+# If installed in the $PATH, execute it (replacing this current process)
+if command -v git-conceal &> /dev/null; then
+    git-conceal unlock "env:${env_var_name}"
+    exit $?
+fi
+
+INSTALL_DIR="${PWD}"
+# If was already previously installed in INSTALL_DIR, execute it (replacing this current process)
+if [[ -x "${INSTALL_DIR}/git-conceal" ]]; then
+    "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
+    exit $?
+fi
+
+# Otherwise, install it locally and execute it
+echo "git-conceal binary not found. Installing it..."
+mkdir -p "${INSTALL_DIR}"
+curl -fsSL https://raw.githubusercontent.com/Automattic/git-conceal/trunk/install.sh | bash -s -- --prefix "${INSTALL_DIR}"
+"${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
--- a/bin/git-conceal-unlock
+++ b/bin/git-conceal-unlock
@@ -12,6 +12,10 @@
 
 set -euo pipefail
 
+# Note: this default env var name has been purposely chosen with a `_SECRET_KEY` suffix
+# so that Buildkite will automatically redact any accidental leak of its value in CI logs.
+# See https://buildkite.com/docs/agent/v3/cli-pipeline#redacted-vars
+# See https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/cb27042/s3secrets-helper/secrets/secrets.go#L31-L38
 env_var_name=${1:-GIT_CONCEAL_SECRET_KEY}
 
 # If installed in the $PATH, execute it (replacing this current process)
--- a/bin/git-conceal-unlock
+++ b/bin/git-conceal-unlock
@@ -21,7 +21,7 @@ if command -v git-conceal &> /dev/null; then
 fi
 
 INSTALL_DIR="${PWD}"
-# If was already previously installed in INSTALL_DIR, execute it (replacing this current process)
+# If already installed in INSTALL_DIR, execute it (replacing this current process)
 if [[ -x "${INSTALL_DIR}/git-conceal" ]]; then
     "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
     exit $?
--- a/bin/git-conceal-unlock
+++ b/bin/git-conceal-unlock
@@ -9,7 +9,6 @@
 #
 # By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
 # You can override it by passing the environment variable name as the first argument.
-#
 
 set -euo pipefail
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -40,7 +40,7 @@ _None._
 
 ### New Features
 
-_None._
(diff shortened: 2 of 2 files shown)


git-crypt/README.md
```diff
@@ -0,0 +1,49 @@
+# git-crypt Docker image
+
+## Why?
+
+The `Dockerfile` in this folder is used to be able to run [`git-crypt`](https://github.com/AGWA/git-crypt) on any platform.
+
+Indeed, while `git-crypt` is easy to install on macOS platforms via a simple `brew install git-crypt`, it is not always as easy to install on all other platforms:
+ - While some Linux distributions may have versions of `git-crypt` available via `yum install`/`dnf install`/`apk add`/…, they might not always have the latest version.
+ - Support for `git-crypt` on Windows platforms is at its infancy, less tested and harder to compile on that platform to begin with.
```
wzieba: Thanks for the explanation 👍

AliSoftware:
I believe the test failure on CI is unrelated to this PR and might instead be related to our recent changes to those scripts, or how the test tries to source those scripts for testing (especially as `./bin` is not in the `$PATH` when we run the function under test in a testing context).

@mokagio Care to take a look to fix that unrelated part (in a separate PR)?

bin/git-conceal-unlock
```diff
@@ -0,0 +1,35 @@
...
+# Description:
+#   A wrapper script to run `git-conceal unlock` on CI
+#   See https://github.com/Automattic/git-conceal
+#
+#   The wrapper automatically runs the `install.sh` script of `git-conceal`
+#   if it's not already installed locally before running `git-conceal unlock env:…`.
+#
+# By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
+# You can override it by passing the environment variable name as the first argument.
+#
```
mokagio: nitpick: Unless this is a style convention that I'm not familiar with, I'd either go with 

```suggestion
#
# Description:
#   A wrapper script to run `git-conceal unlock` on CI
#   See https://github.com/Automattic/git-conceal
#
#   The wrapper automatically runs the `install.sh` script of `git-conceal`
#   if it's not already installed locally before running `git-conceal unlock env:…`.
#
# By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
# You can override it by passing the environment variable name as the first argument.
#
```

or

```suggestion
# Description:
#   A wrapper script to run `git-conceal unlock` on CI
#   See https://github.com/Automattic/git-conceal
#
#   The wrapper automatically runs the `install.sh` script of `git-conceal`
#   if it's not already installed locally before running `git-conceal unlock env:…`.
#
# By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
# You can override it by passing the environment variable name as the first argument.
```

bin/git-conceal-unlock
```diff
@@ -0,0 +1,35 @@
...
+
+set -euo pipefail
+
+env_var_name=${1:-GIT_CONCEAL_SECRET_KEY}
+
+# If installed in the $PATH, execute it (replacing this current process)
+if command -v git-conceal &> /dev/null; then
+    git-conceal unlock "env:${env_var_name}"
+    exit $?
```
mokagio: Given the `set -e`, `exit $?` is redundant for all cases other than the successful one.

The comment above says, "replacing this current process". In such case, have you considered `exec`?

```suggestion
    exec git-conceal unlock "env:${env_var_name}"
```

bin/git-conceal-unlock
```diff
@@ -0,0 +1,35 @@
...
+#   The wrapper automatically runs the `install.sh` script of `git-conceal`
+#   if it's not already installed locally before running `git-conceal unlock env:…`.
+#
+# By default, it uses the `GIT_CONCEAL_SECRET_KEY` environment variable as the key source.
+# You can override it by passing the environment variable name as the first argument.
+#
+
+set -euo pipefail
+
+env_var_name=${1:-GIT_CONCEAL_SECRET_KEY}
```
mokagio, line 19: If `GIT_CONCEAL_SECRET_KEY` is not set in the environment, the script won't fail at this stage or later when reading `env_var_name` because then we're not accessing `$GIT_CONCEAL_SECRET` but only printing `GIT_CONCEAL_SECRET`(as far as I understand).

image

I suppose `git-conceal` will fail gracefully if the value given to `env:` is unset, but I wonder if we should fail early here? Or, given it's such an edge case, should we keep as is?

bin/git-conceal-unlock
```diff
@@ -0,0 +1,35 @@
...
+
+# If installed in the $PATH, execute it (replacing this current process)
+if command -v git-conceal &> /dev/null; then
+    git-conceal unlock "env:${env_var_name}"
+    exit $?
+fi
+
+INSTALL_DIR="${PWD}"
+# If was already previously installed in INSTALL_DIR, execute it (replacing this current process)
```
mokagio:
```suggestion
# If already installed in INSTALL_DIR, execute it (replacing this current process)
```

bin/git-conceal-unlock
```diff
@@ -0,0 +1,35 @@
...
+if [[ -x "${INSTALL_DIR}/git-conceal" ]]; then
+    "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
+    exit $?
```
mokagio: In case the suggestion above is accepted:

```suggestion
    exec "${INSTALL_DIR}/git-conceal" unlock "env:${env_var_name}"
```

bin/git-conceal-unlock
AliSoftware: Ah actually the comment comes from the fact that I was using `exec` in the first iterations of that script, so I commented accordingly that it was _replacing_ the process.

I then later switched to call the command as a subprocess instead of `exec`, because I feel like `exec` in bash was not necessarily well known by many devs for how it works / what it does (replacing the current process while keeping the PID and stdin/stdout etc) while calling a command was more common pattern in our shell scripts, so I felt it was less obscure to use the latter for readability for others… but then I didn't update the comment 🙃 

I'm up for switching back to `exec` if that's not a pattern you find confusing 👍

bin/git-conceal-unlock
AliSoftware: > we're not accessing $GIT_CONCEAL_SECRET but only printing GIT_CONCEAL_SECRET(as far as I understand).

This is intentional and a security feature:

The `unlock` command of `git-conceal` intentionally takes the _name_ of the env var and not its value to avoid accidental leaks in CI logs (e.g. if you do `echo "${GIT_CONCEAL_SECRET_KEY}"`). That's why the command takes `env:GIT_CONCEAL_SECRET_KEY` and not `"$GIT_CONCEAL_SECRET_KEY"` for its parameter

> I suppose git-conceal will fail gracefully if the value given to env: is unset

The `git-conceal unlock env:DOES_NOT_EXIST` command is the one that will **internally** read the value of the env var (instead of letting the shell read it and risk exposing it), and [errors with a clear message if it's empty or it doesn't exist](https://github.com/Automattic/git-conceal/blob/ae65f81e059a0998827f204426cb752243e34202/src/key.rs#L88-L96) indeed.

> I wonder if we should fail early here? Or, given it's such an edge case, should we keep as is?
I think we should keep it as is. Because doing otherwise would mean have the shell read the env var to check if it exists or is nil, which if we write the code slightly incorrectly or fail to account for edge cases could increase the risk of leaking the key.

AliSoftware: By the way, the name of the env var we use in this script (and thus expect to use in `.mobile-secrets/CI/secrets/<pipeline>/env` by default) was also chosen carefully so that its name matches [the patterns that are automatically redacted in the Buildkite logs by default](https://buildkite.com/docs/agent/v3/cli-pipeline#redacted-vars)—including when the `s3-secrets-helper` hook [we use in our agents](https://github.com/Automattic/buildkite-ci/blob/trunk/src/agents/macos-vms/playbook/roles/buildkite/templates/buildkite-hooks/environment#L57-L61) to inject those `env` secrets from our S3 bucket [injects those](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/main/s3secrets-helper/secrets/secrets.go#L31-L38) and add the ones with such [suffixes](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/pull/146) to [the `buildkite-agent redactor add`](https://github.com/buildkite/elastic-ci-stack-s3-secrets-hooks/blob/cb2704255976d448a825fc67d3b8008d489800a1/s3secrets-helper/secrets/secrets.go#L620)

🎗️ I should probably add a note in the comments of this script mentioning that the default env var name having the `_SECRET_KEY` suffix has the benefit of making Buildkite automatically redact the secret in logs in case its were accidentally printed.

AliSoftware: Note added about env var name suffix chosen purposely so it's auto-redacted in logs if leaked: 4f8cad67accd2e5b4286c3def4babbc3c3f0b269

mokagio: TIL about the `_SECRET_KEY`. Lovely Buildkite DX.

Neat to have the comment there. Thanks!

mokagio (approved):
[I pointed the guinea pig PR to the latest commit on this branch](https://github.com/woocommerce/woocommerce-android/pull/14979/commits/d8b121fb68a524b6cbba75b50aaf225204d6f840) to confirm the latest worked as expected. All green other than the Dangermattic check, which is unrelated

image

:shipit:

PR #196, opened 2025-11-19
Fix Windows tests

iangmaia:
As reported in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/195, Windows unit tests started failing, presumably after all the changes in our Windows AMI.
This PR should fix the Windows unit tests.


bin/prepare_windows_host_for_app_distribution.ps1
```diff
@@ -27,21 +27,29 @@ param (
...
+    $amiVersion = [Environment]::GetEnvironmentVariable('AMI_VERSION', 'Machine')
+  }
+  Write-Output "AMI_VERSION is: $amiVersion"
+  if (![string]::IsNullOrWhiteSpace($amiVersion) -and [version]$amiVersion -ge [version]'0.2') {
+    Write-Output "Tooling is already pre-installed in AMI versions >= 0.2. Only installing code signing certificate."
+    & "$PSScriptRoot\setup_windows_code_signing.ps1"
+    Exit 0
+  }
+}
+
```
AliSoftware: Those changes don't seem right nor necessary.

The whole point of testing the `AMI_VERSION` as early as possible in this script is that we can exit early and make this script mostly a no-op if it was called by a repo and pipeline that hasn't been updated yet (to stop calling this script altogether) after we deployed the custom AMI for the Windows queue that contains all the pre-installed tooling. In that case, we don't even care if the parameter combinations or the `SkipWindows10SDKInstallation` parameters were set or not.

In the medium term this script will completely disappear and we don't expect any repo to call it anymore now that the custom AMI has been deployed anyway. This early exit on `AMI_VERSION>=2` was thus to help with the transition, so that while we're working on this and have already deployed the custom AMI on our `windows` queue in CI, any repo that is still calling this `prepare_windows_host_for_app_distribution.ps1` (because we haven't had the time to update the `.buildkite/commands/*.sh` scripts of said repo just yet) would result in those being an almost no-op (modulo the installation of the signing certificate, which for good security reasons is not baked into the custom AMI). In the long run we plan to update all those `.buildkite/commands/*.sh` of all the repos that were calling this to just call `setup_windows_code_signing.ps1` directly instead, and once it's done, we'll delete the `prepare_windows_host_for_app_distribution.ps1` from the `release-toolkit` entirely.

---

In fact, we could even consider deleting that `prepare_windows_host_for_app_distribution.ps1` script from the `release-toolkit` repo right away right now, given that any repo that is still using it (because we haven't had the time to update their `.buildkite/commands/*.sh` just yet) is still pointing to `release-toolkit` version `5.7.0` which still has it (and the whole point of this conditional was to give us the necessary time to update the repos to latest release-toolkit version + `setup_windows_code_signing.ps1` calls all in one go)

AliSoftware (commented):
Overall my suggestion for the fix is to just completely delete the `prepare_windows_host_for_app_distribution.ps1` and its test from the repo.

Once we'll find the time to update all the repos that were still calling `prepare_windows_host_for_app_distribution.ps1` in their `.buildkite/commands/*.sh` scripts and relied on the assumption that they might still be running on the base AMI and not our custom one, we'll have those repos just call `setup_windows_code_signing.ps1` directly instead and bump their `release-toolkit` version to `6.0.0`, at which point no repo using `release-tookit >5.7.0` will call `prepare_windows_host_for_app_distribution.ps1` and it won't be needed anymore (which was the whole point of creating our custom AMI in the first place after all.

bin/prepare_windows_host_for_app_distribution.ps1
iangmaia: > In the medium term this script will completely disappear 

> we could even consider deleting that `prepare_windows_host_for_app_distribution.ps1` script [...] right away right now, given that any repo that is still using it [...] is still pointing to release-toolkit version 5.7.0

Ohhh yes indeed. I think I confused myself a bit with what's in the AMI (and if we'd get something keeping the script) -- I agree with removing it straight away to avoid confusions / failing tests given apps will stop using it. Updated on ce15769753ac54dde3b23f63d7c6ad4a9bd209b3.

CHANGELOG.md
```diff
@@ -46,7 +46,7 @@ _None._
 
 ### Internal Changes
 
-_None._
+- Fixed Windows unit tests after Windows AMI changes. [#196]
```
```suggestion
_None._
```

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- We've removed the `prepare_windows_host_for_app_distribution.ps1` script as this functionality is now implemented in the Windows AMI itself. [#196]
```
AliSoftware: I don't think we ever used the "We've …" form/wording in any of our other changelog entries in the past?

```suggestion
- Removed the `prepare_windows_host_for_app_distribution.ps1` script as what it was installing is now already pre-provisioned in our latest custom Windows AMI already. [#196]
```

AliSoftware:
Also feel free to make this PR target `trunk` instead so that it doesn't depend on mine.

Once we merge this PR I can rebase my `git-crypt` PR onto the new `trunk` as I continue working on it (as I want to test it on all our agents before I merge it, so it might take some time until it lands)

CHANGELOG.md
```diff
@@ -34,7 +34,7 @@ _None._
 
 ### Breaking Changes
 
-_None._
+- Removed the `prepare_windows_host_for_app_distribution.ps1` script as what it was installing is now already pre-provisioned in our latest custom Windows AMI already. [#196]
```
AliSoftware: Whoops repetition of `already`

```suggestion
- Removed the `prepare_windows_host_for_app_distribution.ps1` script, given what it was installing is now already pre-provisioned in our custom Windows AMI. [#196]
```

AliSoftware (approved):

PR #210, opened 2026-06-08
Add macOS sign_and_notarize command

mokagio:
Adds a `sign_and_notarize` command. The toolkit has a Windows code-signing command but no macOS counterpart, so consuming pipelines ([platform-imessage](https://github.com/beeper/platform-imessage) today; [bbctl](https://github.com/beeper/bridge-manager) and [Cortext](https://github.com/Automattic/cortext) next) each hand-roll the same `codesign` + `notarytool` script. This centralizes it.

The command is build-agnostic — point it at whatever the build produced (bare binary, `.app`, or `.dmg`).

### Gotcha

Bare binaries are intentionally **not** stapled — only bundles/installers can carry a stapled ticket; a bare binary relies on online notarization.

### How to test

- `make test` runs the bats arg/env contract cases.
- Real end-to-end signing runs on macOS CI via the [platform-imessage refactor](https://github.com/beeper/platform-imessage/pull/100), which repoints its toolkit plugin pin at this branch.

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-- Add `sign_and_notarize`: code-sign (Developer ID, hardened runtime) and notarize macOS artifacts — bare binaries, `.app`, `.dmg`, or `.pkg`.
+- Add `sign_and_notarize`: code-sign (Developer ID, hardened runtime) and notarize macOS artifacts — bare binaries, `.app`, or `.dmg`.
 
 ### Bug Fixes
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Add `sign_and_notarize`: code-sign (Developer ID, hardened runtime) and notarize macOS artifacts — bare binaries, `.app`, `.dmg`, or `.pkg`.
 
 ### Bug Fixes
 
--- a/bin/sign_and_notarize
+++ b/bin/sign_and_notarize
@@ -110,6 +110,9 @@ sign() {
 	echo "==> codesigning $target"
 	codesign "${opts[@]}" "$target"
 	codesign --verify --strict --verbose=2 "$target"
+	# Surface which cert and team identifier actually applied — invaluable when a
+	# Gatekeeper rejection lands in production and you need to confirm the signer.
+	codesign --display --verbose=2 "$target" 2>&1 | grep -E "Authority|TeamIdentifier|Signature|flags|Hash" || true
 }
 
 notarize() {
--- a/bin/sign_and_notarize
+++ b/bin/sign_and_notarize
@@ -1,0 +1,177 @@
+#!/bin/bash -eu
+
+# Usage
+# sign_and_notarize [--team-id TEAM_ID] [--entitlements FILE] <artifact> [<artifact>...]
+#
+# Code-signs (Developer ID, hardened runtime) and notarizes one or more macOS artifacts: bare
+# executables, `.app` bundles, `.dmg`, or `.pkg`. Build-agnostic — point it at whatever your build
+# produced.
+#
+# The Developer ID Application certificate must already be in the keychain (e.g. fetched via
+# `fastlane match` / a `set_up_signing` lane before calling this). The identity is resolved from the
+# keychain by team id; set `IDENTITY` to force a specific one.
+#
+# Required environment (App Store Connect API key, used by notarytool):
+#   APP_STORE_CONNECT_API_KEY_KEY_ID
+#   APP_STORE_CONNECT_API_KEY_ISSUER_ID
+#   APP_STORE_CONNECT_API_KEY_KEY      (the .p8 PEM; `\n` escapes are decoded)
+#
+# Options:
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -38,7 +38,7 @@ _None._
 
 ### New Features
 
-_None._
+- Add `sign_and_notarize`: code-sign (Developer ID, hardened runtime) and notarize macOS artifacts — bare binaries, `.app`, `.dmg`, or `.pkg`.
```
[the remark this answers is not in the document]
mokagio: Good catch — you're right that `codesign` + Developer ID Application doesn't produce a valid `.pkg` signature (that needs `productsign` + Developer ID Installer). Rather than implement installer signing now, I've narrowed the advertised scope to bare binaries, `.app`, and `.dmg`, and removed the `.pkg` branches from the code so docs and behaviour match. Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/commit/6848c743aa50e332ec95809cf2c50eaf00cfcf1e.

*Posted by Claude (Opus 4.8) on behalf of @mokagio with approval.*

bin/sign_and_notarize
```diff
@@ -0,0 +1,180 @@
+#!/bin/bash -eu
+
+# Usage
+# sign_and_notarize [--team-id TEAM_ID] [--entitlements FILE] <artifact> [<artifact>...]
```
mokagio, line 4: Up for debate: The fastlane implementation for notarization may be more reliable than this hand-rolled script. As such, maybe we should only sign via this script. Or, delegate to fastlane altogether. After all, we _need_ fastlane to get the certificate to sign in Developer ID mode.

On the other hand, the process is simple enough that having this dedicated script might be DRYer than using fastlane.

We could also explore moving the fastlane implementation here (via direct calls of `gym`, `match` and `notarize`) to (maybe) get the best of both worlds.

Regardless, I think this is a decent step to at least have consistency across the few apps and CLIs in our roster when it comes to code signing and as such I'd like to merge now and iterate later.

iangmaia: > fastlane implementation for notarization may be more reliable than this hand-rolled script. As such, maybe we should only sign via this script.

I'm not sure I follow.
And in any case, what is the biggest advantage of having a custom script? Or a script combined with Fastlane?
What's the "best of both worlds"?

The main advantage I can think of would be to eventually avoid Ruby and its dependencies... but most of our projects already rely on such tooling anyway.
Maybe if the functionality isn't directly available via Fastlane and requires multiple calls and some logic, maybe you could add it to `release-toolkit` instead (so a new action that orchestrates calls to Fastlane + adds logic on top)? 🤔

bin/sign_and_notarize
```diff
@@ -0,0 +1,180 @@
...
+	xcrun notarytool submit "$payload" \
+		--key "$p8" \
+		--key-id "$key_id" \
+		--issuer "$issuer_id" \
+		--wait \
+		--output-format json > "$out"
+
+	local status submission_id
+	status=$(python3 -c 'import json,sys; print(json.load(open(sys.argv[1]))["status"])' "$out")
+	submission_id=$(python3 -c 'import json,sys; print(json.load(open(sys.argv[1]))["id"])' "$out")
```
iangmaia, line 146: This introduces `python3`, but I guess we could do it with `jq` assuming it is available, given this is a mac?

mokagio: Good point! However, I think our VMs must already have `python3` available because there is no additional code to set it up in the script or in the consumer repo.

bin/sign_and_notarize
mokagio: Thanks Ian. Some context first. This implementation evolved from the AI agent being given much autonomy (and, likely, not enough context on previous approaches). When operating on the relatively green field (at least as far as Ruby automation goes) of the recent standalone CLIs that we wanted to code sign for macOS, it defaulted to Bash. I then prompted it to consolidate the implementation and the result was to extract this script at the CI toolkit level.

As per my comment above, in reviewing the code at the time I was a bit skeptical of the value of having a script for the "x-platform CLI that needs macOS code signing" use case. I'm not surprised you found it debatable, too.

Now that I'm more removed from the work, I feel even more strongly that we should accept that fastlane is our go to tool for this kind of work, even if it may look odd to have Ruby code in JS/Go/Rust repos. This is really the crucial point for me:

> After all, we _need_ fastlane to get the certificate to sign in Developer ID mode.

mokagio: I created [Only use fastlane to build, sign, and notarize CLIs](https://linear.app/a8c/project/only-use-fastlane-to-build-sign-and-notarize-clis-55666b3ccec0/overview) to follow up on this and move back to fastlane. 

I'm going to tag this PR as do not merge, but leave it open so the commit remain on GitHub and the projects that use it won't be affected.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
      <th width="100%" data-danger-table="true" data-kind="Error">
          1 Error
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:no_entry_sign:</td>
      <td data-sticky="false">This PR is tagged with <code>do not merge</code> label(s).</td>
    </tr>
  </tbody>
</table>

</p>

PR #216, opened 2026-07-16
Restore cache from Synology NAS before S3 on macOS

wzieba:
### Description

On the CyberLynk Mac fleet we serve the cache (SPM, gems, etc.) to builders over **billed S3 egress**, even though a Synology NAS on the same LAN already mirrors the `a8c-ci-cache` bucket via CloudSync.

### Change

`restore_cache now` tries the NAS first on macOS hosts (HTTP `HEAD` probe with a short timeout, then `GET`), and falls back to S3 on any miss or unreachable NAS. The chosen source is logged. Connecting by IP with a Host header mirrors `hostmgr`'s `CacheServer`.

### Context and research

paaHJt-ael-p2

#### Gems
image

https://buildkite.com/automattic/day-one-apple/builds/7132/list?tab=output&jid=019f6a2e-b2cf-44b0-845c-662f779d7436#L354

#### SPM
image

https://buildkite.com/automattic/day-one-apple/builds/7132/list?tab=output&jid=019f6a2e-b2cf-44b0-845c-662f779d7436#L411

---

### DIFF
--- a/bin/restore_cache
+++ b/bin/restore_cache
@@ -20,22 +20,38 @@ fi
 
 echo "Using $S3_BUCKET_NAME as cache bucket"
 
-if aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/null 2>&1; then
-	SECONDS=0
-	echo "Restoring cache entry $CACHE_KEY"
+# On the macOS (CyberLynk) fleet, a Synology NAS mirrors the cache bucket on the
+# LAN and serves it over HTTP, so we fetch from it first to avoid billed S3
+# egress and fall back to S3. Mirrors hostmgr: connect by IP, set the Host header
+# (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
+NAS_CACHE_HOST="cache.ci.internal"
+NAS_CACHE_ADDRESS="192.168.0.250"
 
-	echo "	Downloading"
+restore_from_nas() {
+	[[ "$(uname)" == "Darwin" ]] || return 1
+	local url="http://$NAS_CACHE_ADDRESS/cache/$CACHE_KEY"
+	curl -fsS --head --max-time 5 -H "Host: $NAS_CACHE_HOST" "$url" > /dev/null 2>&1 || return 1
+	echo "	Fetching from NAS (Synology)"
+	curl -fsS -H "Host: $NAS_CACHE_HOST" "$url" -o "$CACHE_KEY"
+}
+
+restore_from_s3() {
+	aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/null 2>&1 || return 1
+	echo "	Fetching from AWS S3"
 	# If the bucket has transfer acceleration enabled, use it!
 	ACCELERATION_STATUS=$(aws s3api get-bucket-accelerate-configuration --bucket "$S3_BUCKET_NAME" | jq '.Status' -r || true)
-
 	if [ "$ACCELERATION_STATUS" = "Enabled" ]; then
-		echo "Downloading with transfer acceleration"
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet --endpoint-url https://s3-accelerate.amazonaws.com
 	else
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet
 	fi
+}
 
+SECONDS=0
+if restore_from_nas || restore_from_s3; then
+	echo "Restoring cache entry $CACHE_KEY"
 	CACHE_SIZE=$(wc -c < "$CACHE_KEY")
+
 	echo "	Decompressing"
 	tar -xf "$CACHE_KEY"
 
--- a/bin/restore_cache
+++ b/bin/restore_cache
@@ -43,6 +43,7 @@ restore_from_s3() {
 	# If the bucket has transfer acceleration enabled, use it!
 	ACCELERATION_STATUS=$(aws s3api get-bucket-accelerate-configuration --bucket "$S3_BUCKET_NAME" | jq '.Status' -r || true)
 	if [ "$ACCELERATION_STATUS" = "Enabled" ]; then
+		echo "Downloading with transfer acceleration"
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet --endpoint-url https://s3-accelerate.amazonaws.com
 	else
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet
--- a/bin/restore_cache
+++ b/bin/restore_cache
@@ -53,7 +53,7 @@ restore_from_s3() {
 annotate() {
 	[[ "$(uname)" == "Darwin" ]] || return 0
 	command -v buildkite-agent > /dev/null 2>&1 || return 0
-	printf -- "- %s\n" "$1" | buildkite-agent annotate --context cache-restore --scope job --append || echo "⚠️ cache-restore annotation failed"
+	printf -- "- %s\n" "$1" | buildkite-agent annotate --context cache-restore --scope job --style info --append || echo "⚠️ cache-restore annotation failed"
 }
 
 SECONDS=0
--- a/bin/restore_cache
+++ b/bin/restore_cache
@@ -20,17 +20,17 @@ fi
 
 echo "Using $S3_BUCKET_NAME as cache bucket"
 
-# On the macOS (CyberLynk) fleet, a Synology NAS mirrors the cache bucket on the
-# LAN and serves it over HTTP, so we fetch from it first to avoid billed S3
-# egress and fall back to S3. Mirrors hostmgr: connect by IP, set the Host header
-# (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
-NAS_CACHE_HOST="cache.ci.internal"
-NAS_CACHE_ADDRESS="192.168.0.250"
+# A Synology NAS on the macOS (CyberLynk) fleet's LAN mirrors the cache bucket
+# and serves it over HTTP, so we fetch from it first to avoid billed S3 egress
+# and fall back to S3. BUILDKITE_NAS_CACHE_HOST / _ADDRESS are exported only on
+# those agents (via buildkite-ci's macOS environment hooks); the BUILDKITE_ prefix
+# is what lets hostmgr forward them into the build VM, so their presence enables
+# the NAS path. Mirrors hostmgr: connect by IP, set the Host header (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
+NAS_CACHE_HOST=${BUILDKITE_NAS_CACHE_HOST:-}
+NAS_CACHE_ADDRESS=${BUILDKITE_NAS_CACHE_ADDRESS:-}
 
-# The NAS is only reachable from the macOS (CyberLynk) fleet, so only those
-# hosts try it; everything else goes straight to S3.
 SHOULD_TRY_NAS=false
-[[ "$(uname)" == "Darwin" ]] && SHOULD_TRY_NAS=true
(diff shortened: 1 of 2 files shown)


bin/restore_cache
```diff
@@ -20,32 +20,60 @@ fi
...
 	else
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet
 	fi
+}
 
+annotate() {
+	[[ "$(uname)" == "Darwin" ]] || return 0
+	command -v buildkite-agent > /dev/null 2>&1 || return 0
+	printf -- "- %s\n" "$1" | buildkite-agent annotate --context cache-restore --scope job --append || echo "⚠️ cache-restore annotation failed"
+}
```
wzieba, line 65: This issue is partially caused by reduced visibility. I don't think we should fail a job in case of a fallback to S3 (sometimes even it's by design if CloudSync didn't finish yet), but I thought this annotation could help provide us more info faster (and it doesn't bother anyone I think).

Image

AliSoftware: I like the idea but I think having the annotation systematically (including on cache hit) on every job is a bit more noisy than we need (given we already have the info in the logs anyway)

Wdyt about only printing the annotation if there was a cache miss (i.e. Darwin machine and missed on NAS and falling back to S3)

bin/restore_cache
```diff
@@ -20,32 +20,60 @@ fi
...
-	echo "Restoring cache entry $CACHE_KEY"
+# On the macOS (CyberLynk) fleet, a Synology NAS mirrors the cache bucket on the
+# LAN and serves it over HTTP, so we fetch from it first to avoid billed S3
+# egress and fall back to S3. Mirrors hostmgr: connect by IP, set the Host header
+# (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
+NAS_CACHE_HOST="cache.ci.internal"
+NAS_CACHE_ADDRESS="192.168.0.250"
+
+restore_from_nas() {
+	[[ "$(uname)" == "Darwin" ]] || return 1
```
AliSoftware: I was wondering if testing `"$(uname)" == "Darwin"` was the right approach there 🤔
Sure that should work but that means we're testing a coincidence (of the Macs in our fleet being the only ones being able to access the NAS because they're the only one on the same network) instead of testing the root logic (is the NAS on the same network as the current machine) 🤔

At the very least I'd suggest to move all the tests against `Darwin` into a helper function like `is_nas_reachable()`, or a global variable running the test only once at the start of the script and giving the value a name, so that even if for the time being we keep the test being based on `uname`, it'd be clearer what the test is about, and make it easier to change later.

As for finding another condition for said test, I wonder if we can't rather rely on something from the `aws` command line to tell us which network group name or whatever the machine lives in 🤔

bin/restore_cache
```diff
@@ -20,32 +20,60 @@ fi
...
 
-if aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/null 2>&1; then
-	SECONDS=0
-	echo "Restoring cache entry $CACHE_KEY"
+# On the macOS (CyberLynk) fleet, a Synology NAS mirrors the cache bucket on the
+# LAN and serves it over HTTP, so we fetch from it first to avoid billed S3
+# egress and fall back to S3. Mirrors hostmgr: connect by IP, set the Host header
+# (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
+NAS_CACHE_HOST="cache.ci.internal"
+NAS_CACHE_ADDRESS="192.168.0.250"
```
AliSoftware: I think it could be nice if those were configurable via the buildkite plugin (even if those values were still the defaults), just [like we already do for the name of the S3 bucket](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin#bucket-optional-string)

bin/restore_cache
wzieba: Sounds fair, applied in 925c17ee579644dd9edeb09dd7b95ee4036006c2

bin/restore_cache
wzieba: `"$(uname)" == "Darwin"` is a cheap way to exclude all AWS EC2s: I don't want all of our EC2s to now start pinging a (for them) random private ip of NAS.

> At the very least I'd suggest to move all the tests against `Darwin` into a helper function like `is_nas_reachable()`, or a global variable running the test only once at the start of the script and giving the value a name, so that even if for the time being we keep the test being based on uname, it'd be clearer what the test is about, and make it easier to change later.

Sounds good to me, added a variable with explanation: 341aaf12496497e76791b71e3be3b2be782ce88e

> As for finding another condition for said test, I wonder if we can't rather rely on something from the aws command line to tell us which network group name or whatever the machine lives in 🤔

I'm sceptical as it's more complex than a simple, static check of "am I on macOS?". I think it's a good proxy check for our scenario. I understand it's not bulletproof, but I highly doubt we'll e.g. add a Linux machine to the CyberLynk fleet. WDYT?

bin/restore_cache
wzieba: I see, but would it work to move this just to ENV variables? While I can understand that someone wants to use a specific `S3 bucket`, I don't think any project will need to provide a "custom NAS Synology IP and hostname" - it's a very specific, implementation detail IMO.

Adding this as parameters for the plugin sounds like unnecessary complication for consumers (even when they only discover it through documentation).

bin/restore_cache
AliSoftware: Yeah that's why I suggested to centralize the check but still keep it relying on Darwin or not for now, as a compromise between the two.

I could see us adding Linux machines in our MacMiniVault fleet in the far future — after all, we considered it not that long ago when we were working on the bare-metal agent to run Android emulators for UI tests, before we ended up settling on using a Mac for that mostly because it was cheaper than a Linux box after all.
We are not there yet, so maybe it's ok for now. But if that ends up happening at one point in the future I can totally see us missing this and forgetting to update `restore_cache` to account for it

AliSoftware: Actually I think there's an easier solution that will cover both my comments and solve this gracefully:

 - Make the `NAS_CACHE_HOST` / `NAS_CACHE_ADDRESS` env vars not be hardcoded in the script, but provided as external env vars / plugin parameters (like the S3 bucket is)
 - In our CI infra, only set the values for `NAS_CACHE_HOST` and `NAS_CACHE_ADDRESS` in the `environment` hook of the machines that live in our MacMiniVault infra (i.e., currently just our `mac` and `mac-metal` instances)
 - That way your script can just skip testing the NAS if those variables are not defined (e.g. if we're on an EC2 instance, or a Tumblr or Kubernetes instance…) and only test it if the variables are defined.

wzieba: > In our CI infra, only set the values for `NAS_CACHE_HOST` and `NAS_CACHE_ADDRESS` in the environment hook of the machines that live in our MacMiniVault infra (i.e., currently just our `mac` and `mac-metal` instances)

Can you share a direction on how it'd look? I'm aware of `env` in our secrets, but this is shared between all agents.

AliSoftware: Sure. This would simply consist of adding those to https://github.com/Automattic/buildkite-ci/blob/trunk/src/agents/macos-hosts/resources/buildkite-hooks/environment and https://github.com/Automattic/buildkite-ci/blob/trunk/src/agents/mac-metal/resources/buildkite-hooks/environment

bin/restore_cache
AliSoftware: Yeah maybe making them as plugin parameters is overkill and just moving them to env variables outside of the script (and instead set on the `environment` hook of the machines themselves) would be enough. I'm ok with that compromise 👍 

(FYI, adding a parameter to a Buildkite plugin is as easy as adding an entry for it in [this `plugin.yaml`](https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/trunk/plugin.yml#L14-L15) then reading the value from `$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_{paramname}` env var (e.g. `$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_NAS_CACHE_HOST` & `$BUILDKITE_PLUGIN_A8C_CI_TOOLKIT_NAS_CACHE_ADDRESS`) in your scripts).

bin/restore_cache
wzieba: Ok thanks. We still could forget setting these variables when introducing a new e.g. Linux machine, but I guess it's less likely now:
- https://github.com/Automattic/buildkite-ci/pull/902/
- 30538aa1c967d1c7b4277b644063f8f128e6cc66

WDYT?

wzieba: plus 41682c7 , I missed that the AI-added comment was too detailed

bin/restore_cache
```diff
@@ -20,22 +20,48 @@ fi
...
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet --endpoint-url https://s3-accelerate.amazonaws.com
 	else
 		aws s3 cp "s3://$S3_BUCKET_NAME/$CACHE_KEY" "$CACHE_KEY" --quiet
-	fi
+	fi || return 1
+	annotate "S3 fallback \`$CACHE_KEY\`"
+}
+
+annotate() {
+	[[ "$SHOULD_TRY_NAS" == true ]] || return 0
```
AliSoftware: I feel like having this condition here makes it hard to reason about. When I read this that make me wondering "wait, why should we make the annotate function not work if we have access to the NAS? I don't see the logical reasoning"

I think it would make it more readable if you kept the `annotate()` function focused on doing the Buildkite annotation, and move that condition at the call site on line 51 instead, like this:
```diff
-     annotate "S3 fallback \`$CACHE_KEY\`"
+     # Only annotate about S3 fallback if we had to try NAS first and we tried S3 because NAS failed
+     [[ "$SHOULD_TRY_NAS" == true ]] && annotate "Cache miss on NAS but hit on S3 fallback: \`$CACHE_KEY\`"
```

Wdyt?

bin/restore_cache
```diff
@@ -20,22 +20,48 @@ fi
 
 echo "Using $S3_BUCKET_NAME as cache bucket"
 
-if aws s3api head-object --bucket "$S3_BUCKET_NAME" --key "$CACHE_KEY" > /dev/null 2>&1; then
-	SECONDS=0
-	echo "Restoring cache entry $CACHE_KEY"
+# A Synology NAS on the macOS (CyberLynk) fleet's LAN mirrors the cache bucket
+# and serves it over HTTP, so we fetch from it first to avoid billed S3 egress
+# and fall back to S3. Mirrors hostmgr: connect by IP, set the Host header (see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
```
AliSoftware: ⛏️   💄  Nitpick: adding links to our docs about the Synology NAS cache synchronized via CloudSync
```suggestion
# Our Synology NAS maintains a mirror of the S3 cache bucket (via Synology CloudSync) and serves it over HTTP.
# See https://github.com/Automattic/buildkite-ci/blob/trunk/docs/synology/synology-nas.md
# So if we are on the same LAN as this NAS (i.e. macOS fleet in MacMiniVault), we fetch from it first
# to avoid billed S3 egress, and only fall back to S3 on cache miss on the NAS.
# Mirrors hostmgr behavior (connect by IP, set the Host header, see https://github.com/Automattic/hostmgr/blob/fa4ed2b10e3163aec6bcffe4d6b94fc94f3466ec/Sources/libhostmgr/Servers/CacheServer.swift#L46-L72).
```

bin/restore_cache
```diff
@@ -20,22 +20,48 @@ fi
...
+NAS_CACHE_HOST=${BUILDKITE_NAS_CACHE_HOST:-}
+NAS_CACHE_ADDRESS=${BUILDKITE_NAS_CACHE_ADDRESS:-}
```
AliSoftware, line 32: ⛏️ 💄  Adding small comment to help understand where those values come from

```suggestion

# Those `BUILDKITE_NAS_CACHE_*` vars are defined in the `environment` Buildkite hook of our CI machines living in MacMiniVault in the same LAN as the NAS
# See https://github.com/Automattic/buildkite-ci/pull/902
NAS_CACHE_HOST=${BUILDKITE_NAS_CACHE_HOST:-}
NAS_CACHE_ADDRESS=${BUILDKITE_NAS_CACHE_ADDRESS:-}
```

AliSoftware (approved):
Looking good! Only left a couple of nitpicks (comments, moving logic condition at call site) to make it more readable but those are not hard blockers.

bin/restore_cache
wzieba: > I think it would make it more readable if you kept the annotate() function focused on doing the Buildkite annotation, and move that condition at the call site on line 51 instead, like this:

Heh I was suggesting exactly the same, and for the same reason, but I was convinced by AI that the convention in this project is "guards inside methods" rather than checks before 🙂. Both approaches are fine to me

PR #218, opened 2026-07-18
Add checksum-verified a8c-secrets installer

mokagio:
## Why

Consuming repos bootstrap `a8c-secrets` with `curl -fsSL .../main/install.sh | bash`, which runs an unreviewed, mutable script from `main` and then installs whatever the `latest` release is, unverified. [Copilot](https://github.com/Automattic/pocket-casts-ios/pull/4731#discussion_r3565275428) and [Olivier](https://github.com/Automattic/pocket-casts-ios/pull/4731#discussion_r3589565698) flagged this as a supply-chain risk for a tool that handles secrets, and asked for a version pin + checksum verification. This DRYs that into a single toolkit command. Experiment per [AINFRA-2710](https://linear.app/a8c/issue/AINFRA-2710).

## What

`install_a8c_secrets` downloads a version-pinned release binary directly from the GitHub releases CDN and refuses to install it unless its SHA-256 matches a checksum vendored in the script. No `curl | bash`, no GitHub API/token needed.

## Tradeoffs

- Upstream publishes no `.sha256` assets, so expected hashes are recorded inline per target triple and must be bumped together with `A8C_SECRETS_VERSION`. The bump procedure is documented in the script header.
- Reimplements the platform→triple mapping from upstream's `install.sh`. Delegating to that script can't pin a version or verify the binary, which is the whole point.

## How to test

- `make lint && make test` (Docker). Locally I ran `shellcheck` and `bats tests/test_install_a8c_secrets.bats` (8 cases, hermetic via `--print-plan`).
- End-to-end on this Mac: `install_a8c_secrets --prefix /tmp/x` installs and runs `a8c-secrets 1.0.0`; a tampered expected checksum aborts with exit 1 and installs nothing.

Follow-up (separate PR, gated on a toolkit release): swap `pocket-casts-ios`'s `.buildkite/commands/install-secrets.sh` curl line for a call to `install_a8c_secrets`.

---

### DIFF
--- a/bin/install_a8c-secrets_binary
+++ b/bin/install_a8c-secrets_binary
@@ -16,9 +16,9 @@ set -euo pipefail
 #   the value recorded below. Everything else — platform detection, install location,
 #   the macOS quarantine strip — mirrors upstream's `install.sh`.
 #
-#   Bumping the version: update `A8C_SECRETS_VERSION` and every entry in
-#   `expected_checksum` together — the checksums are only valid for the pinned version.
-#   Get a new checksum with `shasum -a 256` on the downloaded release asset.
+#   Bumping the version: update `A8C_SECRETS_VERSION` and every `A8C_SECRETS_CHECKSUM_*`
+#   constant together — the checksums are only valid for the pinned version. Get a new
+#   checksum with `shasum -a 256` on the downloaded release asset.
 #
 # Usage:
 #   install_a8c-secrets_binary
@@ -43,8 +43,12 @@ set -euo pipefail
 # Requirements:
 #   `curl`, `install`, `uname`, and the toolkit's own `hash_file` helper.
 
-# The pinned release. Keep in lockstep with `expected_checksum` below.
+# The pinned release and the SHA-256 of each `a8c-secrets-<triple>-$A8C_SECRETS_VERSION`
+# asset we publish and have vetted. Bump them together.
 A8C_SECRETS_VERSION="1.0.0"
+A8C_SECRETS_CHECKSUM_AARCH64_APPLE_DARWIN="2b59604261053d2b57a805c53e3b727c43b750e821b587c0492dee7f717bab24"
+A8C_SECRETS_CHECKSUM_X86_64_UNKNOWN_LINUX_GNU="b8fd670c430843ff5af954a85903dc1726d5ca8e3294f725e5e521ac86683bf7"
+A8C_SECRETS_CHECKSUM_X86_64_PC_WINDOWS_GNU="fd58cc1f7330de1078042ad1cb7afcb127ac66e8de840f202b03b70fd829b086"
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 
@@ -92,14 +96,13 @@ resolve_target_triple() {
 	esac
 }
 
-# SHA-256 of `a8c-secrets-<triple>-$A8C_SECRETS_VERSION` for each triple we publish and
-# have vetted. Only triples with an entry here are installable; a resolvable triple with
-# no entry (e.g. a target upstream didn't publish for this version) is refused.
+# Only triples with a pinned checksum are installable; a resolvable triple with none
+# (e.g. a target upstream didn't publish for this version) is refused.
 expected_checksum() {
 	case "$1" in
-		aarch64-apple-darwin) echo "2b59604261053d2b57a805c53e3b727c43b750e821b587c0492dee7f717bab24" ;;
-		x86_64-unknown-linux-gnu) echo "b8fd670c430843ff5af954a85903dc1726d5ca8e3294f725e5e521ac86683bf7" ;;
-		x86_64-pc-windows-gnu) echo "fd58cc1f7330de1078042ad1cb7afcb127ac66e8de840f202b03b70fd829b086" ;;
+		aarch64-apple-darwin) echo "$A8C_SECRETS_CHECKSUM_AARCH64_APPLE_DARWIN" ;;
+		x86_64-unknown-linux-gnu) echo "$A8C_SECRETS_CHECKSUM_X86_64_UNKNOWN_LINUX_GNU" ;;
+		x86_64-pc-windows-gnu) echo "$A8C_SECRETS_CHECKSUM_X86_64_PC_WINDOWS_GNU" ;;
 		*) return 1 ;;
 	esac
 }
--- a/bin/install_a8c-secrets_binary
+++ b/bin/install_a8c-secrets_binary
@@ -21,23 +21,26 @@ set -euo pipefail
 #   checksum with `shasum -a 256` on the downloaded release asset.
 #
 # Usage:
-#   install_a8c-secrets_binary [--os <os>] [--arch <arch>]
+#   install_a8c-secrets_binary [--os <os>] [--arch <arch>] [--install-dir <path>]
 #
 # Options:
-#   --os <os>       Target platform, in `uname -s` form (e.g. `Darwin`). Defaults to
-#                   the host's.
-#   --arch <arch>   Target architecture, in `uname -m` form (e.g. `arm64`). Defaults to
-#                   the host's.
-#   -h, --help      Print usage and exit.
+#   --os <os>              Target platform, in `uname -s` form (e.g. `Darwin`). Defaults
+#                          to the host's.
+#   --arch <arch>          Target architecture, in `uname -m` form (e.g. `arm64`).
+#                          Defaults to the host's.
+#   --install-dir <path>   Directory to install the binary into. Defaults to the
+#                          location described in the notes below.
+#   -h, --help             Print usage and exit.
 #
 # Examples:
 #   install_a8c-secrets_binary
 #   install_a8c-secrets_binary --os Linux --arch x86_64
+#   install_a8c-secrets_binary --install-dir "$HOME/bin"
 #
 # Notes:
-#   Installs to `/usr/local/bin` when writable, otherwise `$HOME/.local/bin` (matching
-#   upstream). The latter is not always on the agent's PATH; the command prints a
-#   reminder to add it when that is the case.
+#   Without `--install-dir`, installs to `/usr/local/bin` when writable, otherwise
+#   `$HOME/.local/bin` (matching upstream). The latter is not always on the agent's
+#   PATH; the command prints a reminder to add it when that is the case.
 #   Windows (MSYS/MinGW bash) is supported on a best-effort basis, mirroring upstream.
 #
 # Returns:
@@ -67,7 +70,7 @@ cleanup() {
 trap cleanup EXIT
 
 usage() {
-	echo "usage: install_a8c-secrets_binary [--os <os>] [--arch <arch>]"
+	echo "usage: install_a8c-secrets_binary [--os <os>] [--arch <arch>] [--install-dir <path>]"
 }
 
 # Map `uname -s` / `uname -m` to the Rust target triple used in the release asset
@@ -111,25 +114,41 @@ expected_checksum() {
 	esac
 }
 
+# Where to put the binary: the `--install-dir` value when given, otherwise the default
+# location upstream's install.sh picks for the target.
+resolve_install_dir() {
+	local triple="$1" override="${2:-}" local_app_data_dir
+	if [[ -n "$override" ]]; then
+		echo "$override"
+	elif [[ "$triple" == *windows* ]]; then
+		local_app_data_dir="${LOCALAPPDATA:+$LOCALAPPDATA/Programs/a8c-secrets}"
+		echo "${local_app_data_dir:-$HOME/.local/bin}"
+	elif [[ -w "/usr/local/bin" ]]; then
+		echo "/usr/local/bin"
+	else
+		echo "$HOME/.local/bin"
+	fi
+}
+
 main() {
-	local os="" arch=""
+	local os="" arch="" install_dir=""
 	while [[ $# -gt 0 ]]; do
 		case "$1" in
 			-h | --help)
 				usage
 				return 0
 				;;
-			--os | --arch)
+			--os | --arch | --install-dir)
 				if [[ $# -lt 2 ]]; then
 					echo "missing value for $1" >&2
 					usage >&2
 					return 1
 				fi
-				if [[ "$1" == "--os" ]]; then
-					os="$2"
-				else
-					arch="$2"
-				fi
+				case "$1" in
+					--os) os="$2" ;;
+					--arch) arch="$2" ;;
+					--install-dir) install_dir="$2" ;;
+				esac
 				shift 2
 				;;
 			*)
@@ -171,16 +190,7 @@ main() {
 	asset="a8c-secrets-${triple}-${A8C_SECRETS_VERSION}${asset_suffix}"
 	url="https://github.com/Automattic/a8c-secrets/releases/download/${A8C_SECRETS_VERSION}/${asset}"
 
-	# Default install location, mirroring upstream install.sh.
-	local install_dir
-	if [[ "$triple" == *windows* ]]; then
-		install_dir="${LOCALAPPDATA:+$LOCALAPPDATA/Programs/a8c-secrets}"
-		install_dir="${install_dir:-$HOME/.local/bin}"
-	elif [[ -w "/usr/local/bin" ]]; then
-		install_dir="/usr/local/bin"
-	else
-		install_dir="$HOME/.local/bin"
-	fi
+	install_dir="$(resolve_install_dir "$triple" "$install_dir")"
 
 	local tmp_bin
 	TMP_DIR="$(mktemp -d)"
--- a/bin/install_a8c-secrets_binary
+++ b/bin/install_a8c-secrets_binary
@@ -21,14 +21,18 @@ set -euo pipefail
 #   checksum with `shasum -a 256` on the downloaded release asset.
 #
 # Usage:
-#   install_a8c-secrets_binary
+#   install_a8c-secrets_binary [--os <os>] [--arch <arch>]
 #
 # Options:
-#   -h, --help   Print usage and exit.
+#   --os <os>       Target platform, in `uname -s` form (e.g. `Darwin`). Defaults to
+#                   the host's.
+#   --arch <arch>   Target architecture, in `uname -m` form (e.g. `arm64`). Defaults to
+#                   the host's.
+#   -h, --help      Print usage and exit.
 #
-# Environment:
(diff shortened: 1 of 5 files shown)


bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,204 @@
...
+#   Installs the `a8c-secrets` CLI on a CI agent, pinned to a known release and
+#   verified against a vendored SHA-256 checksum before it is put on disk.
+#
+#   Upstream's documented install path is `curl -fsSL .../main/install.sh | bash`,
+#   which executes a mutable script from `main` and then downloads whatever the
+#   `latest` release happens to be, unverified. Since `a8c-secrets` handles secrets,
+#   this command instead downloads a specific, version-pinned release binary straight
+#   from the GitHub releases CDN and refuses to install it unless its checksum matches
+#   the value recorded below. Everything else — platform detection, install location,
+#   the macOS quarantine strip — mirrors upstream's `install.sh`.
```
AliSoftware: Not 100% sure this part of the comment is worth keeping 🤷

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+#   which executes a mutable script from `main` and then downloads whatever the
+#   `latest` release happens to be, unverified. Since `a8c-secrets` handles secrets,
+#   this command instead downloads a specific, version-pinned release binary straight
+#   from the GitHub releases CDN and refuses to install it unless its checksum matches
+#   the value recorded below. Everything else — platform detection, install location,
+#   the macOS quarantine strip — mirrors upstream's `install.sh`.
+#
+#   Bumping the version: update `A8C_SECRETS_VERSION` and every entry in
+#   `expected_checksum` together — the checksums are only valid for the pinned version.
+#   Get a new checksum with `shasum -a 256` on the downloaded release asset.
```
AliSoftware: It'd be nice if all the constants that needs bumping would be grouped together instead, i.e. having `A8C_SECRETS_CHECKSUM_WINDOWS`/`A8C_SECRETS_CHECKSUM_MAC`/`A8C_SECRETS_CHECKSUM_LINUX` constants next to `A8C_SECRETS_VERSION` constant all be declared together near the top of the script (like `A8C_SECRETS_VERSION` already is on line 47), not only to make such update easier and more localized, but so that it's also easier to remember that one has to update the checksums alongside the version number.

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+#   0 on success; 1 on usage, download or checksum error; 2 on an unsupported platform
+#   or a target with no pinned checksum.
+#
+# Requirements:
+#   `curl`, `install`, `uname`, and the toolkit's own `hash_file` helper.
+
+set -euo pipefail
+
+# The pinned release. Keep in lockstep with `expected_checksum` below.
+A8C_SECRETS_VERSION="1.0.0"
```
AliSoftware: I wonder if it'd be worth it to allow the caller to provide those values via arguments, i.e. call `install_a8c_secrets 1.0.0 --mac <checksum> --windows <checksum> --linux <checksum>` or similar?

(Optionally allowing to _not_ provide _all_ the checksums on that command line if you're calling that helper from an agent that you already know is a Mac/Windows/Linux machine)

That way we could still have the helper install a fixed version (`1.0.0` today) by default, but if we need to migrate, or want to test the installation and behavior of a new version on just one pipeline (or just one CI agent) for smoke-testing, we would not need to update this `install_a8c_secrets` script for every repo at once but instead could do a gradual rollout if needed (without having to update the `a8c-ci-toolkit` for each update of the version we want `install_a8c_secrets` to install)

wdyt?

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+TMP_DIR=""
+cleanup() {
+	if [[ -n "$TMP_DIR" ]]; then
+		rm -rf "$TMP_DIR"
+	fi
+}
+trap cleanup EXIT
+
+usage() {
+	echo "usage: install_a8c_secrets"
```
AliSoftware: I'm of two minds about this, but the name of this script feels a bit unsatisfying because the binary it installs is named `a8c-secrets`, not `a8c_secrets`.
OTOH, naming the script `install_a8c-secrets` and thus making it named with a mix of `_` and `-` also feels a bit odd…

Maybe a compromise could be `install_a8c-secrets_binary`, which looks more balanced? 🤷

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+}
+
+# SHA-256 of `a8c-secrets-<triple>-$A8C_SECRETS_VERSION` for each triple we publish and
+# have vetted. Only triples with an entry here are installable; a resolvable triple with
+# no entry (e.g. a target upstream didn't publish for this version) is refused.
+expected_checksum() {
+	case "$1" in
+		aarch64-apple-darwin) echo "2b59604261053d2b57a805c53e3b727c43b750e821b587c0492dee7f717bab24" ;;
+		x86_64-unknown-linux-gnu) echo "b8fd670c430843ff5af954a85903dc1726d5ca8e3294f725e5e521ac86683bf7" ;;
+		x86_64-pc-windows-gnu) echo "fd58cc1f7330de1078042ad1cb7afcb127ac66e8de840f202b03b70fd829b086" ;;
```
AliSoftware: As mentioned above, I'd suggest to move those 3 SHA1 as constants at the top of the script, closer to the `A8C_SECRETS_VERSION` constant.

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+		*)
+			echo "unknown option: $1" >&2
+			usage >&2
+			return 1
+			;;
+	esac
+
+	local os arch triple
+	os="${A8C_SECRETS_OS:-$(uname -s)}"
+	arch="${A8C_SECRETS_ARCH:-$(uname -m)}"
```
AliSoftware: Not convinced we need to support overriding the OS and ARCH via env vars?

Or is this only for you to be able to write unit tests for those? But if so, maybe exposing those as parameters `--os` / `--arch` would then allow this script to be more generic about those all while exposing those more officially? 🤷

bin/install_a8c_secrets
```diff
@@ -0,0 +1,204 @@
...
+	# Default install location, mirroring upstream install.sh.
+	local install_dir
+	if [[ "$triple" == *windows* ]]; then
+		install_dir="${LOCALAPPDATA:+$LOCALAPPDATA/Programs/a8c-secrets}"
+		install_dir="${install_dir:-$HOME/.local/bin}"
+	elif [[ -w "/usr/local/bin" ]]; then
+		install_dir="/usr/local/bin"
+	else
+		install_dir="$HOME/.local/bin"
+	fi
```
AliSoftware: I noticed that this removes from the original `install.sh` the ability to provide your own install path (which could be provided via the optional `--prefix` flag in the original `install.sh`, name borrowed from other install tools like brew, but I'm open to call it `--install-dir` instead if that makes it clearer.

The benefit I see from allowing to install in a custom folder is:
1. On CI, we may want to install it inside a local dir because we might not have the admin permissions to install it in `/usr/local/bin`.
    - Sure, this is partially covered by your lines L157-L160 which tests if `/usr/local/bin` is writable and if not installs it in `~/.local/bin`. But on a CI agent where the `$HOME` is shared across jobs, that might create an implicit shared state between jobs
    - Also if we have multiple jobs that end up installing different versions of `a8c-secrets` (because e.g. they use different versions of the `a8c-ci-toolkit` that hardcode a different `A8C_SECRETS_VERSION` for each, or we implement my suggestion of giving the ability to override the version to install via parameters…), installing it in `~/.local/bin` is not the best solution either
    - So for this use case it might actually make sense to have that helper install `a8c-secrets` in the job's working directory / checkout directory or similar?
2. If one wants to install `a8c-secrets` locally on their machine, they might prefer installing it in `~/bin` or `~/tools/` or whatever, depending on how they prefer organizing their own machine.
    - Agreed this 2nd argument is a bit weak because this particular `bin/install_a8c_secrets` is not intended to be called by developers on their local machine, it being intended to be used as a Buildkite Plugin and thus on CI.
    - But I could totally see us copy/pasting the logic of that script somewhere else (maybe even inside the `a8c-secrets` repo as the future replacement for `install.sh`) in the future, especially as this version has the benefit of validating the checksums. While we will still  have CI use this `bin/install_a8c_secrets` helper to avoid the risk of supply-chain attack on our CI of having a blind `curl … | bash`, our devs wanting to install (a new version of) `a8c-secrets` on their local machine could still benefit from downloading and running that `install.sh` script that would now checks the checksums and everything. And at that point having said script support a custom `--prefix`/`--install-dir` would be welcome, so if such future version of `install.sh` ends up coming from a copy/paste of this `install_a8c_secrets`, better make it flexible anyway.

bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,204 @@
...
+	fi
+
+	local install_path="$install_dir/$exe_name"
+	echo "Installing to $install_path..."
+	"$install_cmd" -d "$install_dir"
+	"$install_cmd" -m 755 "$tmp_bin" "$install_path"
+
+	# a8c-secrets is downloaded, not built locally, so macOS quarantines it.
+	if [[ "$os" == "Darwin" ]]; then
+		xattr -d com.apple.quarantine "$install_path" 2>/dev/null || true
```
AliSoftware, line 219: I just realized that we might want to add some logs about this too

```suggestion
		echo "Removing quarantine attribute from $install_path..."
		xattr -d com.apple.quarantine "$install_path" 2>/dev/null || true
```

bin/install_a8c-secrets_binary
mokagio: Doesn't matter how much I tell my agents to not be verbose, they keep adding comments like these all over the place.

bin/install_a8c_secrets
mokagio: `install_a8c-secrets_binary` sounds good to me.

bin/install_a8c_secrets
mokagio: +1 for using a parameter over the env var to allow unit testing.

bin/install_a8c-secrets_binary
mokagio: Definitely need some logs.

bin/install_a8c_secrets
mokagio: I like that adopting or testing out a new `a8c-secrets` version would not require a new CI toolkit version or branch.

At the same time, I understand checksum validation as a defensive mechanism against supply-chain attacks. If we allow callers to pass the checksum, then the hash stops coming from a vetted source, and the check goes from "this is the binary someone with access to CI toolkit approved" to "this is the binary someone with access _to the repo running the command_ said it would be".

Granted, the distinction might not be as clear cut as I'm making it out to be, given Automattic has a hight-trust culture where most people have access to most resources.

Still, I feel that given a8c-secret will be a crucial piece of infra, a bit of friction in updating it in the interest of more security/control might be worth it.

We _could_ make the command accept a `--version` and update it to have per-version triples, so a caller can switch between (previously vetted in CI toolkit) versions. But, I don't know why one would want to install an older version of the tool. I mean, I can think of a new version introducing a bug, and maybe if we change core pieces such as the way the files are encrypted, but if the toolkit always installs the latest version, then addressing those could be done via changing toolkit version... That is to say, I'm okay to follow up with a `--version` parameter, but if my concern on the loss of vetting behind my doubts on `--[platform] <checksum>` is sound, then adding `--version` might not be that useful.

bin/install_a8c_secrets
mokagio: I added `--install-dir` and aliased it to `--prefix`.

bin/install_a8c_secrets
AliSoftware: Yeah you make a good point. Maybe not add `--version` for now at least then.

mokagio:
Thanks for the suggestions @AliSoftware ! I incorporated most of them. Tested this implementation in https://buildkite.com/automattic/pocket-casts-ios/builds/16740

image

bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,231 @@
...
+
+# Script: install_a8c-secrets_binary
+#
+# Description:
+#   Installs the `a8c-secrets` CLI on a CI agent, pinned to a known release and
+#   verified against a vendored SHA-256 checksum before it is put on disk.
+#
+#   Bumping the version: update `A8C_SECRETS_VERSION` and every `A8C_SECRETS_CHECKSUM_*`
+#   constant together, as the checksums are only valid for the pinned version. Get a new
+#   checksum with `shasum -a 256` on the downloaded release asset.
```
AliSoftware: Or more simply, use the ⧉ copy button provided by the GitHub UI to copy the checksums shown on the [GitHub Release page](https://github.com/Automattic/a8c-secrets/releases) of the a8c-secrets repo for each binary of that release? (which avoids having to download all 3 assets on your computer yourself just to be able to run `shasum -a 256` on each of them…)

Image

bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,231 @@
...
+# Examples:
+#   install_a8c-secrets_binary
+#   install_a8c-secrets_binary --os Linux --arch x86_64
+#   install_a8c-secrets_binary --install-dir "$HOME/bin"
+#
+# Notes:
+#   Without `--install-dir`, installs to `/usr/local/bin` when writable,
+#   otherwise `$HOME/.local/bin`. The latter is not always on the agent's PATH
+#   so the command prints a reminder to add it when that is the case. Windows
+#   (MSYS/MinGW bash) is supported on a best-effort basis, mirroring upstream.
```
AliSoftware: The mention of "upstream" in those comments is a bit confusing. I think the AI meant for this to mean "the `install.sh` script this was inspired from and that is in the upstream/remote `a8c-secrets` repo? But this mention makes things more confusing than they need out of context.

bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,231 @@
...
+#   install_a8c-secrets_binary [--os <os>] [--arch <arch>] [--install-dir <path>]
+#
+# Options:
+#   --os <os>              Target platform, in `uname -s` form (e.g. `Darwin`). Defaults
+#                          to the host's.
+#   --arch <arch>          Target architecture, in `uname -m` form (e.g. `arm64`).
+#                          Defaults to the host's.
+#   --install-dir <path>   Directory to install the binary into. Defaults to the
+#                          location described in the notes below. Also accepted as
+#                          `--prefix`, the name upstream's `install.sh` uses.
```
AliSoftware: Same as below, the mention of "upstream" in that comment is unclear / confusing, especially for someone who would read that comment out of context of this PR and not knowing that this `install_a8c-secrets_binary` script was inspired by the `install.sh` from the `a8c-secrets` repo.

I would suggest to get rid of that mention… and tbh since we haven't yet adopted `a8c-secrets` fully in many repos and even the ones we've adopted it already we haven't used `--prefix` anywhere yet, I'd even get rid of the `--prefix` alias support in this script altogether to avoid having 2 different flags meaning the same thing.

Once this lands, I'm really thinking about copy/pasting most of the logic from your work on this script here into the `install.sh` script in `a8c-secrets`, in particular borrowing some logic about supporting the check for checksum[^1] and similar, especially since I'm planning on making changes and improving this `install.sh` script anyway as part of finalizing https://linear.app/a8c/issue/AINFRA-2708/improve-a8c-secrets-install-script-for-rate-limits-and-version-pinning .

Even if I end up keeping `install.sh` different and write the changes/improvements to it from scratch instead ofmostly-copy/pasting parts of this `install_a8c-secrets_binary` of yours, I think I'll still use the occasion to rename the `--prefix` flag in that `install.sh` script to `--install-dir` anyway, as such a flag name is clearer and more explicit IMHO. So let's normalize towards that `--install-dir` and ultimately drop `--prefix` in both scripts.

[^1]: EDIT: I just realized that it wouldn't make sense to implement checksum checks in the `install.sh` script itself after all, since that would not solve the supply-chain-attack risk and in fact provide a false sense of security (indeed, if someone were planning to do a supply-chain-attack by hijacking the `install.sh` script to serve a malicious binary instead, they could just as well modify (or remove) the checksums that would be hardcoded in that `install.sh` 😉. So unless the caller would check the checksum _of the `install.sh` script itself_ between the `curl` call and running it thru `| bash`, this would be useless in protecting against malicious intent.

bin/install_a8c-secrets_binary
```diff
@@ -0,0 +1,231 @@
...
+
+	if ! triple="$(resolve_target_triple "$os" "$arch")"; then
+		echo "Unsupported platform for a8c-secrets: os=$os arch=$arch" >&2
+		return 2
+	fi
+
+	local checksum
+	if ! checksum="$(expected_checksum "$triple")"; then
+		echo "No pinned checksum for target '$triple' at a8c-secrets $A8C_SECRETS_VERSION." >&2
+		echo "Add one to install_a8c-secrets_binary (see the bump procedure in its header)." >&2
```
AliSoftware:
```suggestion
		echo "If a binary actually exists on a8c-secrets' GitHub Releases for this triple and version, add its checksum to install_a8c-secrets_binary (see the bump procedure in its header)." >&2
```

AliSoftware (approved):
Approving to unblock, since my remaining feedback is only nitpicks on wording/doc/flags cleanup

 - Instructions on how to get checksum for new version can be simpler (no need to download and run shasum on each asset)
 - Support for `--prefix` alias flag can be removed IMHO, I was thinking or renaming the flag in `install.sh` too anyway
 - Nits on some AI-generated comments being confusing/not helping (mention of "upstream" referencing the `install.sh` script your AI used as reference)

bin/install_a8c-secrets_binary
mokagio: https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/218/commits/c18e03cdf489031ac7b1075755cf0058a7ad71c9

bin/install_a8c-secrets_binary
mokagio: https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/218/commits/e3c30dffa9e3b79e400fa7a44a1a81cd09bd8584

bin/install_a8c-secrets_binary
mokagio: `--prefix` removed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/218/commits/4c18b8aca1d53711197463d27e1c669d43593723

mokagio:
Thanks for the review and additional suggestion. Proceeding to merge and ship so we can adopt in repos with stable version.

PR #219, opened 2026-07-18
Adopt `#!/usr/bin/env bash` shebang across scripts

mokagio:
## Why

Resolve `bash` via `PATH` instead of hardcoding `/bin/bash`, so scripts pick up a newer `bash` when one is installed ahead of the system copy (the stock macOS `/bin/bash` is 3.2). Updates the repo's own `.cursor` bash guidance to recommend `#!/usr/bin/env bash` so new scripts follow suit.

Stacked on #218 — review/merge that first.

## Tradeoff worth a look

Scripts that had flags in the shebang (`#!/bin/bash -eu`) can't become `#!/usr/bin/env bash -eu`: the kernel hands `bash -eu` to `env` as one argument, so it wouldn't run. Those (`hash_file`, `get_libc_version`) now set the flags in the body via `set -eu` instead. Behavior is unchanged.

Ruby (`#!/usr/bin/env ruby`) and PowerShell scripts are untouched.

## How to test

`make lint` (shellcheck across all scripts) and `make test`. All 45 migrated scripts are shellcheck-clean.

---

### DIFF
--- a/.cursor/rules/bash_scripts.mdc
+++ b/.cursor/rules/bash_scripts.mdc
@@ -15,11 +15,11 @@ For all bash shell scripts:
 ## Best Practices
 
 For all bash shell scripts:
-- Use `#!/bin/bash` as the first line of the script (a.k.a. shebang)
+- Use `#!/usr/bin/env bash` as the first line of the script (a.k.a. shebang), with a single parameter and no flags after `bash` (the kernel passes `env` everything after the interpreter as one argument, so `#!/usr/bin/env bash -eu` is not portable)
 - Use `set -e` to exit on error
 - Use `set -u` to error on undefined variables
 - Use `set -o pipefail` for proper error handling
-- Combine those options in a single `set -euo pipefail` command, or as part of the shebang line, as appropriate
+- Combine those options in a single `set -euo pipefail` command near the top of the script body
 - Always use `$()` instead of backticks to execute commands
 - Quote all variable expansions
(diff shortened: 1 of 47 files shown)


AliSoftware (approved):
🎉

mokagio:
> Stacked on https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/218 — review/merge that first.

I'm sure I told the AI to do this off `trunk`...

.cursor/rules/bash_scripts.mdc
```diff
@@ -15,11 +15,11 @@ For all bash shell scripts:
 ## Best Practices
 
 For all bash shell scripts:
-- Use `#!/bin/bash` as the first line of the script (a.k.a. shebang)
+- Use `#!/usr/bin/env bash` as the first line of the script (a.k.a. shebang), with a single parameter and no flags after `bash` (the kernel passes `env` everything after the interpreter as one argument, so `#!/usr/bin/env bash -eu` is not portable)
 - Use `set -e` to exit on error
 - Use `set -u` to error on undefined variables
 - Use `set -o pipefail` for proper error handling
-- Combine those options in a single `set -euo pipefail` command, or as part of the shebang line, as appropriate
+- Combine those options in a single `set -euo pipefail` command near the top of the script body
```
mokagio:
```suggestion
- Combine those options in a single `set -euo pipefail` command immediately after the shebang, separated with one empty line
```

PR #220, opened 2026-07-18
Add integration tests for `install_a8c-secrets_binary`

mokagio:
Adds a CI step that runs `install_a8c-secrets_binary ` on Linux and macOS. They run in <5s and <50s respectively. Seem like a cheap and easy way to have faster feedback on changes on that command.

At the same time, given we'll soon have a way to update the CI boxes to manage `a8c-secrets` themselves, this might be redundant.

At the same time... _later is never_ and this is cheap, so, why not?

<details>
<summary>AI-Generated Details</summary>

## Why

`install_a8c_secrets`' unit tests cover the platform/checksum contract but stub the network. This adds a real end-to-end run — download, checksum verification, install, then invoking the binary — on Linux and macOS agents, which is the only place the GitHub releases CDN and the agents' own `install`/`xattr`/writability of `/usr/local/bin` actually come into play.

Stacked on #218 — review/merge that first.

## Note on version

a8c-secrets is still at 1.0.0 (no new release; `main` only has un-released commits, which our checksum-pinned setup can't consume anyway). So this exercises the pinned 1.0.0 path. The test asserts the installed binary reports that version, so it will need updating in lockstep whenever `A8C_SECRETS_VERSION` is bumped.

## How to test

CI runs the new `install_a8c_secrets Tests` group (Linux + macOS). Locally: `HOME=$(mktemp -d) tests/install_a8c_secrets/test_install.sh` installs into the temp home and asserts `a8c-secrets 1.0.0` runs.

---

</details>


.buildkite/pipeline-bash-tests.yml
```diff
@@ -132,3 +132,21 @@ steps:
         notify:
           - github_commit_status:
               context: "pr_changed_files Tests: Edge Cases"
+
+  - group: ":closed_lock_with_key: install_a8c-secrets_binary Tests"
```
mokagio:
```suggestion
  - group: ":closed_lock_with_key: install_a8c-secrets_binary Integration Tests"
```

tests/install_a8c_secrets/test_install.sh
```diff
@@ -0,0 +1,31 @@
...
+echo "--- :mag: Verify the installed binary runs"
+# install_a8c_secrets installs to /usr/local/bin or ~/.local/bin but doesn't touch this
+# shell's PATH, so add both candidate locations before invoking the binary.
+export PATH="/usr/local/bin:$HOME/.local/bin:$PATH"
+
+version="$(a8c-secrets --version)"
+echo "$version"
+
+# The command pins a specific release, so the installed binary must report that version.
+expected="1.0.0"
```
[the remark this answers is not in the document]
mokagio: Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/220/changes/44e5cac790c24fdc0464d0e147042956db4e4d38

bin/install_a8c-secrets_binary
```diff
@@ -51,6 +55,9 @@ A8C_SECRETS_CHECKSUM_AARCH64_APPLE_DARWIN="2b59604261053d2b57a805c53e3b727c43b75
 A8C_SECRETS_CHECKSUM_X86_64_UNKNOWN_LINUX_GNU="b8fd670c430843ff5af954a85903dc1726d5ca8e3294f725e5e521ac86683bf7"
 A8C_SECRETS_CHECKSUM_X86_64_PC_WINDOWS_GNU="fd58cc1f7330de1078042ad1cb7afcb127ac66e8de840f202b03b70fd829b086"
 
+DOWNLOAD_ATTEMPTS="${DOWNLOAD_ATTEMPTS:-3}"
```
iangmaia: Locally, Codex alerted me for the possibility of "Bash arithmetic injection". Not sure how realistic that would be, but considering this is open and could be installed in third party repos, perhaps worth considering. Actually it kinda makes sense to at least not make some of the vars environment variables, but anyway.

The output:

```bash
DOWNLOAD_ATTEMPTS is environment-controlled (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/632f98c6bcaee9bfcec047071d367751dbdac983/bin/install_a8c-secrets_binary#L40-L59) and later used directly in a numeric comparison (https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/blob/632f98c6bcaee9bfcec047071d367751dbdac983/bin/install_a8c-secrets_binary#L122-L134). Bash recursively evaluates arithmetic operands.

I reproduced command execution on the required macOS Bash 3.2.57 after a simulated download failure using a value shaped like:

DOWNLOAD_ATTEMPTS='BASH_SOURCE[$(command)]'

Benign malformed values also crash with “unbound variable”; zero still performs one attempt; huge values are unbounded.

Best fix: remove these environment knobs if they only exist for tests. Otherwise namespace and validate them before use:
- attempts: positive integer with a sensible maximum;
- delay: bounded non-negative integer;
- names such as A8C_SECRETS_DOWNLOAD_ATTEMPTS.
```

bin/install_a8c-secrets_binary
```diff
@@ -108,6 +115,25 @@ expected_checksum() {
...
 }
 
+# The releases CDN is flaky often enough to make a single-shot download a source of CI
+# failures. Retried in bash rather than with curl's `--retry-all-errors`, which needs
+# curl 7.71+ and we can't assume that across the fleet; plain `--retry` wouldn't cover
+# the connection and DNS errors we want to survive.
+download() {
+	local url="$1" destination="$2" attempt=1
+	while true; do
+		if curl --location --silent --show-error --fail --output "$destination" "$url"; then
```
iangmaia: Maybe good to set explicit timeouts?

tests/install_a8c-secrets_binary/test_install.sh
```diff
@@ -0,0 +1,39 @@
...
+	echo "$A8C_SECRETS_VERSION"
+)"
+
+"$INSTALLER" --install-dir "$install_dir"
+
+echo "--- :mag: Verify the installed binary runs"
+version="$("$install_dir/a8c-secrets" --version)"
+echo "$version"
+
+case "$version" in
```
iangmaia: Maybe we could use an exact comparison instead? Like:
```bash
[[ "$version" == "a8c-secrets $expected_version" ]]
```

bin/install_a8c-secrets_binary
mokagio: Good catch, and thanks to Codex — the vector is real: Bash arithmetic-evaluates both operands of `-ge` inside `[[ ]]`, recursively, so an environment value shaped like `BASH_SOURCE[$(cmd)]` runs `cmd`.

Went with your "best fix": the knobs only ever existed so the bats tests could skip the sleep and pin the attempt count, and both tests call `download` directly, so they now pass the tuning as arguments instead. No environment surface left to validate.

Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/220/commits/6375c2a

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

bin/install_a8c-secrets_binary
mokagio: Added `--connect-timeout 10 --max-time 120`. Without them a stalled connection hung the job until Buildkite's own timeout instead of falling through to the retry loop, which rather defeated the point of adding retries. The asset is 3 MB, so 120s is generous.

Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/220/commits/09af06d

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

tests/install_a8c-secrets_binary/test_install.sh
mokagio: Done. The substring match would have been satisfied by anything merely containing the pin — an `a8c-secrets 1.0.0-rc` would have passed a `1.0.0` expectation.

Addressed in https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/220/commits/003811e

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

iangmaia (approved):
👍

PR #223, opened 2026-08-06
Add `checkout_release_branch` command

AliSoftware:
Part of [AINFRA-2828](https://linear.app/a8c/issue/AINFRA-2828/dry-checkout-release-branchsh-scripts-into-a8c-ci-toolkit).

Thirteen app repos each carry a byte-identical copy of `.buildkite/commands/checkout-release-branch.sh`, standardized across them in [AINFRA-2725](https://linear.app/a8c/issue/AINFRA-2725/fix-release-selection-and-checkout-release-branchsh-in-release-tooling). This moves it here as `checkout_release_branch`, so there is one copy to fix the next time it needs fixing. The follow-up PRs in those repos, which delete their copy and call this command instead, are blocked on the next release of this plugin.

### Behaviour

A straight port of the script the repos have today:

- The release version comes from the first argument, else the `RELEASE_VERSION` environment variable, else the `release/*` branch the build is running on.
- The branch is reset to `FETCH_HEAD` rather than pulled, so a working copy that a previous job left at a different commit is corrected rather than merged.

Added on top of the original, to match the conventions of the more recent commands in `bin/`: a `--help` flag, and rejection of unknown options and extra arguments.

### Tests

The tests stub `git` on `PATH` and assert the exact command sequence the script issues. What `reset --hard FETCH_HEAD` then does to a working copy is git's behaviour rather than ours — and the `buildkite/plugin-tester` image ships no `git` to exercise it with. Pinning the sequence is what guards the fix this was extracted from: a `git pull` here would merge a diverged branch instead of discarding it.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

PR #224, opened 2026-08-06
Export `$A8C_CI_TOOLKIT_PLUGIN_DIR` for steps running in a container

AliSoftware:
Part of [AINFRA-2828](https://linear.app/a8c/issue/AINFRA-2828/dry-checkout-release-branchsh-scripts-into-a8c-ci-toolkit), alongside #223.

The commands reach a job by having `bin/` prepended to its `$PATH` in `hooks/environment`. A step that runs its command inside a container gets none of that, because the plugin's files sit on the agent host rather than in the image. Studio's Linux release build is one such step: it uses the `docker` plugin only, which is why it still calls a repo-local `checkout-release-branch.sh` by path rather than a toolkit command.

Exporting the plugin's absolute path lets those steps mount it in and use the commands anyway:

```yml
steps:
  - command: |
      export PATH="$$PATH:/a8c-ci-toolkit/bin"

      checkout_release_branch "$RELEASE_VERSION"
    plugins:
      - automattic/a8c-ci-toolkit#6.2.0
      - docker#v5.13.0:
          image: node:22-bookworm
          expand-volume-vars: true
          volumes:
            - "$$A8C_CI_TOOLKIT_PLUGIN_DIR:/a8c-ci-toolkit:ro"
```

Three details make it work, and the README section spells them out:

- Listing this plugin next to `docker`. All plugins' `environment` hooks run before any plugin's `command` hook, so the export lands before the `docker` plugin builds its `docker run` arguments, whatever order they're listed in.
- `expand-volume-vars: true` on the `docker` plugin, without which the `volumes` entry is passed through verbatim.
- `$$` rather than `$`. Per the docker plugin's docs, "`$VARIABLE_NAME` is resolved at pipeline upload time, whereas `$$VARIABLE_NAME` is at run time" — and neither variable has a value at upload time.

`PLUGIN_ROOT` is now resolved to an absolute path. It was only ever used as a `$PATH` entry before, where a relative path happened to work; as the source of a bind mount it would not.

---


README.md
```diff
@@ -38,6 +38,35 @@ Don't forget to verify what [the latest release](https://github.com/Automattic/a
...
+          expand-volume-vars: true
+          volumes:
+            - "$$A8C_CI_TOOLKIT_PLUGIN_DIR:/a8c-ci-toolkit:ro"
+```
+
+Three things make that work:
+
+- Listing this plugin alongside `docker`. Every plugin's `environment` hook runs before any plugin's `command` hook, so the variable is exported before the `docker` plugin assembles its `docker run` arguments — whichever order the two are listed in.
+- `expand-volume-vars: true`, without which the `docker` plugin passes the `volumes` entry through verbatim instead of interpolating the variable.
+- The `$$` in both `$$A8C_CI_TOOLKIT_PLUGIN_DIR` and `$$PATH`. A single `$` is resolved when the pipeline is uploaded, where neither variable has a value yet; `$$` defers it to when the job runs.
```
AliSoftware:
```suggestion
- The `$$` in both `$$A8C_CI_TOOLKIT_PLUGIN_DIR` and `$$PATH`. A single `$` is resolved when the pipeline is uploaded, where neither variable has a value yet; `$$` defers it to when the job runs. Note that this is only needed if you write that line directly in the `.yml`; if instead your `command:` calls a dedicated `.sh` script you can do `export PATH=$PATH:/a8c-ci-toolkit/bin` in that `.sh` file instead of in the `.yml` and avoid the need for the `$$` escape.
```

AliSoftware:
### Validated on CI ✅

Pointed studio's `CI_TOOLKIT_VERSION` at this branch's head (`e425b62`) and added a temporary step to its PR pipeline exercising the README recipe verbatim: Automattic/studio@977fd2bbd7703bc1d05b2bea8484eda5e2daf482 (reverted right after).

[The job](https://buildkite.com/automattic/studio/builds/20546#019fddaf-35e8-41cb-a651-152cf9c600f0) passed, running `node:24.15.0-bookworm` via the `docker` plugin on the `default` queue:

```
--- :file_folder: What got mounted
total 232
-rwxr-xr-x. 1 root nogroup  583 Aug  7 19:24 add_host_to_ssh_known_hosts
-rwxr-xr-x. 1 root nogroup  365 Aug  7 19:24 add_ssh_key_to_agent
-rwxr-xr-x. 1 root nogroup 8511 Aug  7 19:24 annotate_test_failures
-rwxr-xr-x. 1 root nogroup  669 Aug  7 19:24 build_and_test_pod
--- :mag: Resolve a toolkit command from inside the container
/a8c-ci-toolkit/bin/hash_file
--- :abacus: Run it
dc9f87cd7360d7e25772a1e391e36f011f39b4d62d5f685df4a72dbc18e757e5
```

The `docker run` the plugin assembled confirms `$$A8C_CI_TOOLKIT_PLUGIN_DIR` resolved at run time to this plugin's checkout for that exact SHA:

```
--volume /var/lib/buildkite-agent/plugins/ci-default-i-0b3106c025cc50991-1/github-com-automattic-a8c-ci-toolkit-buildkite-plugin-e425b6287bfe8c1635a14f9ec6c276a4dfcc4c10:/a8c-ci-toolkit:ro
```

So all three moving parts hold up: the `environment` hook exports the path before the `docker` plugin resolves its `volumes`, `expand-volume-vars` interpolates it, and `$$` defers it past pipeline upload.

README.md
```diff
@@ -38,6 +38,35 @@ Don't forget to verify what [the latest release](https://github.com/Automattic/a
...
+```yml
+steps:
+  - command: |
+      export PATH="$$PATH:/a8c-ci-toolkit/bin"
+
+      checkout_release_branch "$RELEASE_VERSION"
+    plugins:
+      - automattic/a8c-ci-toolkit#6.2.0
+      - docker#v5.13.0:
+          image: node:22-bookworm
```
mokagio, line 56: Not something we need to solve in this PR, but I wonder about the pros and cons of tracking version numbers in the README.

- Pro: makes for easier copy paste (but who copy pastes steps these days?)
- Con: needs automation to be kept up to date (but AI makes this cheaper and we already have some linting of the version in the instructions because of Buildkite's requirments)

mokagio (approved):
Approving with a note that my pre-research agent worried that the new `cd .. && pwd` would have created issues in Windows, but I verified that's not the case https://github.com/Automattic/pocket-casts-desktop/pull/509

README.md
AliSoftware: That's a good point, I didn't think of that.

iinm we _already_ have mentions of the version elsewhere in other examples in the `README.md`, so that's not a new problem. But indeed that's still an annoying thing having to update the version in the `README.md` every time we bump just to satisfy `buildkite-plugin-linter`… 🤷

PR #225, opened 2026-08-07
Cover checkout_release_branch's untested branches

mokagio:
Stacked on #223 — base is its branch, not `trunk`, so review that one first and this diff stays three test cases.

### Why

`reset --hard FETCH_HEAD` is the line the command was extracted to preserve, and it was the one git step with no coverage; `fetch` and `checkout` both had failure cases. The other two close an asymmetry: an empty `RELEASE_VERSION` was covered but an empty first argument was not, and a `BUILDKITE_BRANCH` of exactly `release/` strips to an empty version that has to reach the usage error.

### Gotcha

The reset case was mutation-checked rather than assumed: swallowing the failure with `git reset --hard FETCH_HEAD || true` turns exactly that case red and leaves the other 38 green. Without that check the assertion would pass whether or not the script propagates the exit code.

### How to test

`make test`, or `docker run -t --rm -v "$PWD"/:/plugin -w /plugin buildkite/plugin-tester` for just the bats suite. 39 tests, 0 failures on this head.

---

- [x] I have considered if this change warrants release notes and have added them to the appropriate section in the `CHANGELOG.md` if necessary.

Tests only — no release notes, and deliberately no `CHANGELOG.md` edit so this does not add a third PR to the conflict #223 and #224 already have on that file.

PR #226, opened 2026-08-07
Homogenize option handling across bin/ commands

mokagio:
Stacked on #225 (which is stacked on #223) — base is that branch, not `trunk`. No reviewers requested; this is up for discussion on the convention before it's worth anyone's time.

### Why

Three of the 46 commands in `bin/` parse options, and each does it differently:

| | structure | unknown-option message |
|---|---|---|
| `pr_changed_files` | top-level `while`/`case` | `Error: unknown option $1` |
| `install_a8c-secrets_binary` | `main()` + `while`/`case` | `unknown option: $1` |
| `checkout_release_branch` | top-level `if` chain | `Error: unknown option '$1'.` |

The convention picked here takes the **structure** from `install_a8c-secrets_binary` and the **message form** from everywhere else. That split is deliberate: `main()` + `case` is the only shape that extends to an option with a value, but on messages `install_a8c-secrets_binary` is the outlier — its `unknown option: $1` and `missing value for $1` are the only two bare stderr messages in `bin/`, against roughly nine `Error: …` ones elsewhere, including `pr_changed_files`' own.

### Intentional tradeoff

**`pr_changed_files` keeps its top-level parser.** Its parsing is already `while`/`case` — only the `main()` wrapper is missing — and wrapping it means relocating the Buildkite env check and the `git fetch` that run *before* parsing, and converting a set of `exit 255` paths whose codes differ from the other two commands. That is a behaviour-risk refactor for no testability gain, so this PR changes only its two message strings. If the convention lands, that wrapper is a follow-up worth doing on its own.

### Gotcha

`checkout_release_branch`'s version resolution now uses a lowercase local. `local RELEASE_VERSION="${1:-${RELEASE_VERSION:-}}"` would have been reading the variable it was in the middle of declaring, so the local is deliberately named differently from the environment variable it falls back to.

### How to test

`make test` — 39 tests, 0 failures, and `make lint` is clean. The seventeen existing `checkout_release_branch` cases cover every path through the rewritten parser and are unchanged by it; the two `install_a8c-secrets_binary` assertions that quote the message text moved in the same commit as the message.

---

No release notes: no behaviour change beyond the text of three error messages. Also no `CHANGELOG.md` edit on purpose, since #223 and #224 already conflict on that file.

### DIFF
--- a/bin/checkout_release_branch
+++ b/bin/checkout_release_branch
@@ -61,37 +61,48 @@ usage() {
 	echo "usage: checkout_release_branch [<release-version>]"
 }
 
-if [[ "${1:-}" == "-h" || "${1:-}" == "--help" ]]; then
-	usage
-	exit 0
-fi
+main() {
+	while [[ $# -gt 0 ]]; do
+		case "$1" in
+			-h | --help)
+				usage
+				return 0
+				;;
+			-*)
+				echo "Error: unknown option '$1'." >&2
+				usage >&2
+				return 1
+				;;
+			*)
+				break
+				;;
+		esac
+	done
 
-if [[ "${1:-}" == -* ]]; then
-	echo "Error: unknown option '$1'." >&2
-	usage >&2
-	exit 1
-fi
+	if [[ $# -gt 1 ]]; then
+		echo "Error: too many arguments." >&2
+		usage >&2
+		return 1
+	fi
 
-if [[ $# -gt 1 ]]; then
-	echo "Error: too many arguments." >&2
-	usage >&2
-	exit 1
-fi
+	echo "--- :git: Checkout Release Branch"
 
-echo "--- :git: Checkout Release Branch"
+	# Lowercase local so the `RELEASE_VERSION` on the right stays the environment one.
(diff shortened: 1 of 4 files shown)


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

mokagio:
Superseded by #227, which carries the same three commits based on #223 directly instead of on #225.

#225 was never a dependency: two of the three commits touch files already on `trunk`, and the third touches `checkout_release_branch`, which comes from #223. Basing on #225 chained two independent changes for no reason — #225 and #227 are siblings and can merge in either order.

*Posted by Claude (Opus 5) on behalf of @mokagio with approval.*

PR #227, opened 2026-08-07
Homogenize option handling across bin/ commands

mokagio:
Spawned from the work in #223, where I noticed that the options management, which I think is useful and can see more scripts benefiting from, was inconsistently implemented.

I farmed this to Opus 5 for homogenization after asking for a few options. This one made sense to me because `main()` is a construct used in many other scripts/wrappers/tools.

<details>
<summary>AI-generated details</summary>

Stacked on #223, alongside #225 rather than behind it — the two are independent and can merge in either order. No reviewers requested; this is up for discussion on the convention first.

### Why

Three of the 46 commands in `bin/` parse options, and each does it differently:

| | structure | unknown-option message |
|---|---|---|
| `pr_changed_files` | top-level `while`/`case` | `Error: unknown option $1` |
| `install_a8c-secrets_binary` | `main()` + `while`/`case` | `unknown option: $1` |
| `checkout_release_branch` | top-level `if` chain | `Error: unknown option '$1'.` |

The convention here takes the **structure** from `install_a8c-secrets_binary` and the **message form** from everywhere else. That split is deliberate: `main()` + `case` is the only shape that extends to an option with a value, but on messages `install_a8c-secrets_binary` is the outlier — its `unknown option: $1` and `missing value for $1` are the only two bare stderr messages in `bin/`, against roughly nine `Error: …` ones elsewhere, including `pr_changed_files`' own.

### Intentional tradeoff

**`pr_changed_files` keeps its top-level parser.** Its parsing is already `while`/`case` — only the `main()` wrapper is missing — and wrapping it means relocating the Buildkite env check and the `git fetch` that run *before* parsing, and converting a set of `exit 255` paths whose codes differ from the other two commands. That is behaviour risk for no testability gain, so this changes only its two message strings. If the convention lands, that wrapper is a follow-up worth doing on its own.

### Gotcha

`checkout_release_branch`'s version resolution now uses a lowercase local. `local RELEASE_VERSION="${1:-${RELEASE_VERSION:-}}"` would have been reading the variable it was in the middle of declaring, so the local is deliberately named differently from the environment variable it falls back to.

### How to test

`make test` — 36 tests, 0 failures, and `make lint` is clean. #223's sixteen `checkout_release_branch` cases cover every path through the rewritten parser and are unchanged by it. #225 adds three more to that file; the two land in either order without conflict, since it touches only the test and this touches only the command.

The two `install_a8c-secrets_binary` assertions that quote the message text moved in the same commit as the message.

---

No release notes: no behaviour change beyond the text of three error messages. Also no `CHANGELOG.md` edit on purpose, since #223 and #224 already conflict on that file.

</details>


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

bin/checkout_release_branch
```diff
@@ -61,37 +61,48 @@ usage() {
 	echo "usage: checkout_release_branch [<release-version>]"
 }
 
-if [[ "${1:-}" == "-h" || "${1:-}" == "--help" ]]; then
-	usage
-	exit 0
-fi
+main() {
```
AliSoftware, line 64: Having that `main()` embedding and thus indenting that **whole** chunk of code felt a bit much at first; but that's probably because that diff as seen on GitHub makes it look like what you did here ends up indenting and embedding _the whole script_ into a `main()` just to call it at the end.
I only later realized looking closer at the diff that this `main()` wrapping only starts at line 64 and there are some non-main code before it, which then makes it more ok compared to what I initially thought this was ("wrap the whole file content inside a `main()` just to call it at the end)

---

Maybe one middle ground could be to extract the parsing of the args into its own function that returns the branch name (or at least the release version needed to compose that branch name), then the `main()` would just consist of calling that function then do the git dance?
I think it would help isolate "parse args" from "the core goal of the script (which is to do the git dance)".
Not sure if, semantically, that dedicated function should be called `parse_args` (to highlight that it's mostly the `while` loop at parsing) or `parse_requested_release_version`

The only issue with that idea is that might not scale well when standardizing such approach across other scripts in that repo (which is the whole point of that PR), eg for scripts that need to set/return multiple parameters as part of their `while` arg-parsing loop, while bash only allows to return a single value from a function (which is fine for this script which only needs the branch name as the end result of parsing the args, but might not be that simple for other scripts we'd generalize this pattern to)… so maybe I'm overthinking this after all …

Anyway, pre-coffee morning thoughts, so maybe a bit rambling 🙃 but food for thoughts

AliSoftware (commented):
Like the idea of standardizing 👍😍

bin/checkout_release_branch
AliSoftware: Maybe the solution is to adopt the convention used by `pr_changed_files` as the standard instead of use `install_a8c-secrets_binary` as the base: always use `while` loop for arg parsing not `if`, but leave that and the scripts core/main logic that goes after it all top-level rather than in a `main()`?

I'd also be ok to leave it up to each script to decide if a dedicated `main()` is worth it vs top-level, and only standardizing on the "use `while` for arg parsing" aspect 🤷

AliSoftware:
Also once we settle on which convention we want to use we need to teach our agent about it in `AGENT.md`

AliSoftware (approved):
Ok I changed my mind, let's not drag this forever: any standardization is fine by me, and I'm ok to have the `main()` in `bin/checkout_release_branch` even if that ends up englobing (and indenting) almost the whole script due to that `main()` standardization.
It's more important be consistent across scripts IMHO 👍

mokagio:
> Ok I changed my mind, let's not drag this forever: any standardization is fine by me, and I'm ok to have the `main()` in `bin/checkout_release_branch` even if that ends up englobing (and indenting) almost the whole script due to that `main()` standardization. It's more important be consistent across scripts IMHO 👍

Thanks @AliSoftware. When I posted [this comment](https://github.com/Automattic/apps-infra-toolbox/pull/354#issuecomment-5247291950) on merging when it works, not when it's right, I hadn't read your comments here yet. 

But the spirit is the same.

This approach works, even though it might not be right. LLMs make it easier for us to iterate on the kind of mechanical changes such as adopting a different coding pattern to implement the same behavior. So, we are better off cashing in the win ASAP, and continue to move forward.

PR #228, opened 2026-08-07
Bump `.ruby-version` to 3.4.9 and `bundler` to 4.0.18

AliSoftware:
Closes [AINFRA-2838](https://linear.app/a8c/issue/AINFRA-2838)

`.ruby-version` was on `3.2.2`, two minor versions behind, and `Gemfile.lock` recorded `bundler` `2.6.8`.

### Scope

Neither file reaches consuming repos. `.ruby-version` is read only by the `rubocop` and `rspec` targets in the `Makefile`, which derive their `ruby:$(RUBY_VERSION)` image tag from it, and `Gemfile.lock` covers this repo's own `danger` and `rubocop` tooling. The Ruby-related commands in `bin/` — `install_gems`, `dump_ruby_env_info`, the CocoaPods helpers — all use the Ruby of the repo they run against, never this one.

So this is safe to ship in either `6.3.0` or a later `6.3.1`, whichever is convenient.

### How

Both bumps came from the real toolchain rather than a text edit, so the results are the resolver's:

- `.ruby-version` → `3.4.9`, then `bundle install` to confirm the existing lockfile still resolves.
- `bundle update --bundler` under Ruby 3.4.9 to record `4.0.18` (the latest `4.0.x`).

The lockfile diff is one line — `BUNDLED WITH` — with no gem version moved. Worth noting given Ruby 3.4 unbundled a few former default gems and bundler 4 is a major release: neither disturbed this dependency tree.

### Verification

Run after both commits, against the Dockerised targets that CI uses:

- `make lint` — plugin linter, `rubocop` and `shellcheck` clean; the `ruby:3.4.9` container self-installs `bundler` `4.0.18` from `BUNDLED WITH` without help.
- `make test` — 36 bats tests and 47 rspec examples, 0 failures.
- `git status` clean afterwards, so nothing re-resolves on a fresh run.

---


CHANGELOG.md
```diff
@@ -47,6 +47,8 @@ _None._
 ### Internal Changes
 
 - Bump the Xcode image used by this repo's own CI from `xcode-15.4` to `xcode-26.6`, and introduce `.buildkite/shared-pipeline-vars` so that version is now defined in a single place. [#222]
+- Bump `.ruby-version` from `3.2.2` to `3.4.9` [#228]
+- Bump `bundler` from `2.6.8` to `4.0.18` in `Gemfile.lock` [#228]
```
```suggestion
- Update Ruby to `3.4.9` in `.ruby-version` and `bundler` to `4.0.18` in `Gemfile.lock` [#228]
```

mokagio (approved):

PR #229, opened 2026-08-10
Clarify when `$` needs the `$$` escape in the README

mokagio:
AI-assisted refinement to the `$` vs `$$` elaboration in the README. Spawned off reviewing #224.

I don't feel strongly about this, though. If you feel the refinement is unnecessary, I'd default to close the PR.

@iangmaia I'm interested in your input in particular. @AliSoftware is the #224 author and I reviewed it and opened this, so we both have context and some baggage. You could be an impartial observer.
 
<details>
<summary>AI-generated details</summary>

Stacked on #224 — please merge that one first; this targets its branch, so the diff here is one line.

## Rationale

#224's last commit added a note that the `$$` escape is "only needed if you write that line directly in the `.yml`". Read as the rule, that contradicts the example a few lines above it: `checkout_release_branch "$RELEASE_VERSION"` is written directly in the same `command:` block and uses a single `$`.

Both are correct, but for a reason the section never states. The discriminator isn't *where* the line lives, it's *when* the variable gets its value — `RELEASE_VERSION` is set on the build and so has one at upload time, while `$A8C_CI_TOOLKIT_PLUGIN_DIR` and `$PATH` only exist once the job runs. This spells that out and names `$RELEASE_VERSION` as the deliberate exception, so a reader copying the snippet can tell a considered single `$` from a missing escape.

The `.sh` tip is kept, reframed as a way to sidestep the question rather than as the rule itself.

## Gotchas

No `CHANGELOG.md` entry: this only clarifies documentation that #224 itself introduces, and #224 already carries the entry for the feature. Danger will warn about the missing entry; that warning is expected here.

## How to test

`make lint` and `make test` don't cover Markdown, so there's nothing to run — this is a prose review. Worth reading the rendered "Using the commands inside a Docker container" section against the YAML snippet above it and checking the rule now accounts for every `$` in that snippet.

---

</details>


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
    </tr>
  </tbody>
</table>

</p>

README.md
```diff
@@ -63,7 +63,7 @@ Three things make that work:
...
+- The `$$` in both `$$A8C_CI_TOOLKIT_PLUGIN_DIR` and `$$PATH`. A single `$` is resolved when the pipeline is uploaded; `$$` defers it to when the job runs. Escape whichever variables only get their intended value at runtime: `$A8C_CI_TOOLKIT_PLUGIN_DIR` is exported by this plugin's `environment` hook, and `$$PATH` refers to the container's `$PATH` (a single `$PATH` would capture the uploader's PATH), so neither should be expanded at upload time. `$RELEASE_VERSION` above keeps a single `$` for the opposite reason — it is set on the build, so it already has a value when the pipeline is uploaded. To avoid weighing this up line by line, have `command:` call a dedicated `.sh` script instead: nothing inside a `.sh` file is interpolated at upload time, so a plain `export PATH=$PATH:/a8c-ci-toolkit/bin` works there.
```
AliSoftware, line 66: Note that delaying the expansion of `$RELEASE_VERSION` to be done at runtime rather than upload time—either via `$$RELEASE_VERSION`, or by putting the whole `checkout_release_branch "$RELEASE_VERSION"` in a dedicated .`sh` script called by the pipeline—would NOT work unless you pass the value of `RELEASE_VERSION` to the docker container (either via [`propagate-environment: true`](https://github.com/buildkite-plugins/docker-buildkite-plugin#propagate-environment-optional-boolean) or via [`environment: [RELEASE_VERSION]`](https://github.com/buildkite-plugins/docker-buildkite-plugin#environment-optional-array) on the `docker#v…:` plugin entry).

And if you wanted to use a dedicated `.sh` script called from the `command:` to avoid the double-escaping of `$$PATH`:
 - It's still important to note that this wouldn't remove the need to use `$$A8C_CI_TOOLKIT_PLUGIN_DIR` escaping in the `volumes:` attribute, which can't be moved outside the pipeline `.yml` like a `command:` can with a dedicated `.sh`.
 - As for passing the right `RELEASE_VERSION` for the dedicated shell script to resolve, instead of having to pass `environment: [RELEASE_VERSION]` to the docker plugin for it to propagate the value inside the container, I'd argue that it might be better to have the dedicated script take the release version as its `$1`, and then use `command: dedicated-script.sh "$RELEASE_VERSION"` in the `pipeline.yml`, so that this `$RELEASE_VERSION` is still resolved at pipeline-upload time and thus by the time the command is actually called at runtime it ends up having the command already resolved as a call like `dedicated-script.sh "12.3"`

All of this might be a lot to explain in a concise-enough way in a README though. Maybe linking to the official Buildkite docs pages about this, i.e. https://buildkite.com/docs/pipelines/configure/environment-variables#runtime-variable-interpolation and https://buildkite.com/docs/agent/cli/reference/pipeline#environment-variable-substitution

PR #233, opened 2026-08-13
Keep SwiftPM binary artifacts on disk after caching

mokagio:
Part of [AINFRA-2909](https://linear.app/a8c/issue/AINFRA-2909/investigate-swiftpm-binary-target-caching).

<details>
<summary>Overly details AI explanation</summary>

## Rationale

`install_swiftpm_dependencies` deletes `~/Library/Caches/org.swift.swiftpm/artifacts` *after* resolving. That saves no download in the pass that just ran — its only effects are to keep artifacts out of the uploaded tarball and to guarantee that any later resolution in the same job starts from an empty artifact cache.

Day One resolves twice per build job, so it downloads all 13 XCFrameworks (~950 MB) from GitHub a second time, every job. That duplicate is the proximate cause of the `downloadError("Could not connect to the server.")` failures in [DOAPPLE-3674](https://linear.app/a8c/issue/DOAPPLE-3674) — in build 8472 the first pass succeeded and the second failed on all 13.

So this leaves the artifacts where they are and has `save_cache` leave them out of the archive instead.

The comment's justification cannot be right as written. That directory did not exist in the shared cache in March 2023: [swiftlang/swift-package-manager#7101](https://github.com/swiftlang/swift-package-manager/pull/7101) created it eight months later, in Swift 5.10, so the line was a no-op until 5.10 shipped and nobody revisited it. The concrete bug it plausibly meant — a failed download cached as if it were an archive — was fixed upstream in [#7663](https://github.com/swiftlang/swift-package-manager/pull/7663).

A warm `artifacts/` **has** broken a build here since, though: [AINFRA-2154](https://linear.app/a8c/issue/AINFRA-2154), the Day One 2026.6 code freeze, build [1688](https://buildkite.com/automattic/day-one-apple/builds/1688#019cd83d-f4cc-4dd9-950c-9c71aff5fa27).

```
failed downloading '.../5.1.1/ObjectBoxGenerator.artifactbundle.zip' which is required by binary target 'ObjectBoxGenerator':
/Users/builder/Library/Caches/org.swift.swiftpm/artifacts/https___..._ObjectBoxGenerator_artifactbundle_zip already exists in file system
```

That is not an argument against this change. The macOS agents run one ephemeral VM per job and the cache tarball has never carried artifacts, so `artifacts/` was *empty* when that resolution began — the collision was created inside pass 1, not inherited from anywhere. The deletion this PR removes runs after resolution and could not have prevented it, and nothing here alters pass 1.

An empty artifact cache is the stated precondition for [#10203](https://github.com/swiftlang/swift-package-manager/pull/10203), a download race upstream describes as "only present when the cache is empty". I have not proved that is the cause. But re-emptying `artifacts/` after every resolution is how the current code recreates that precondition for every later pass in the job — so whatever AINFRA-2154 was, the deletion is not the cure for it.

## Measured on Day One CI

Build [9305](https://buildkite.com/automattic/day-one-apple/builds/9305) ran two jobs on real mac agents: the same script against this branch and against the current 6.3.0 pin, asserting opposite behaviours. Both passed, so the pair is an A/B rather than a single green run. (Throwaway PR, since closed: bloom/DayOne-Apple#29637.)

| | this branch | 6.3.0 |
|---|---|---|
| Pass 1 | 56s, **14** artifacts kept | 76s, **0** artifacts kept |
| Pass 2 | **29s**, manifest byte-identical | **63s**, 14 artifacts re-downloaded |

The name/mtime/size manifest of `artifacts/` is unchanged across pass 2 on this branch: not one entry was rewritten, so the second resolution went to cache and not to the network. On 6.3.0 the same manifest goes from empty to 14 entries — the re-download, measured rather than inferred.

That is **966 MB** per job (699 MB of it the seven Sentry variants), across the seven jobs that call `set-up-for-building.sh`, and ~34s of resolution per job.

Caveat: that build measured the first revision of this branch, which held `artifacts/` aside by renaming it. What it asserts — artifacts on disk after the save, second resolve served from cache — is unchanged, but the mechanism producing it is not the one under review now.

## Intentional tradeoffs

**The exclusion lives in `save_cache`, not here.** Per review: a `--exclude PATTERN` flag any caller can use, rather than this one command renaming a directory around the save and restoring it through an `EXIT` trap. Its flags now parse with the usual `while` + `case` loop, so they work in any position.

**`save_cache` still accepts its pre-flag calling convention.** Callers pass a bare boolean or an empty string where `--force` now goes — beeper's `cache-bundle.sh` does the latter — so both are accepted and ignored rather than becoming a usage error. Say the word if you'd rather break them loudly instead.

**Exclusion patterns are `tar`'s, and the two implementations disagree.** `--exclude ./artifacts` is anchored to the top level on GNU tar; bsdtar — which is what the mac agents run — also matches a nested `artifacts`, confirmed on an agent (see below). Harmless for this cache, which has no nested directory by that name, but a cross-platform helper can only promise top-level exclusion, so that is all the tests assert.

**Artifacts are still held out of the tarball.** Sharing them *across* jobs is the bigger win and belongs in a follow-up, because it needs an atomic cache restore first: a poisoned entry that fails validation is replayed on every later resolve, and the fix ([#10164](https://github.com/swiftlang/swift-package-manager/pull/10164)) is on `main` only, so no shipping Xcode has it. Within a single job the exposure is unchanged — the current code already resolves against a warm `artifacts/` and only deletes it afterwards.

**Safe for repos with no binary targets.** They have no `artifacts/` to exclude, so the payload is byte-identical. Nothing this does escapes the job.

Deleting `checkouts` alongside it was dead code — the shared cache has no such directory.

## How to test

`make test` — 57 tests, 0 failures under the standard `buildkite/plugin-tester` image. `tests/test_save_cache.bats` runs real `tar` with S3 stubbed, so the archive itself is what it asserts on; `tests/test_install_swiftpm_dependencies.bats` stubs every Apple tool and pins what gets handed to `save_cache`.

### Proven on a real agent

The fixture jobs never exercise `save_cache`'s tar-and-upload leg: they restore an existing entry and stop at `This file is already cached – skipping upload`, because a `binaryTarget` is pinned by checksum in `Package.swift` and never reaches `Package.resolved`, so adding one to the fixture left the cache key where it was. That left `--exclude` covered only by GNU tar in the plugin-tester container.

Closed with a throwaway stacked PR, #237 (green, since closed), build [1466](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1466): a full save → upload → restore under a key containing `$BUILDKITE_BUILD_ID`, so the upload could not be skipped. On `bsdtar 3.5.3 / libarchive 3.7.4` under `bash 3.2.57` it confirmed

- `--exclude ./artifacts` keeps `artifacts/` out of the uploaded archive and `manifests/` in it;
- the same round trip *without* `--exclude` brings `artifacts/` back, so the exclusion is what removes it;
- `PATH KEY false --use_relative_path_in_tar` and `PATH KEY ""` still parse on the agent's own bash;
- `PATH KEY surprise` fails through the guard, matched on its message rather than its exit code.

It also settled the anchoring question: bsdtar dropped a nested `repositories/artifacts` decoy too, so `--exclude ./artifacts` is not path-anchored there.

---

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

</details>


mokagio:
Build [#1454](https://buildkite.com/automattic/ci-toolkit-buildkite-plugin/builds/1454) is green, 19/19, and the fixture does what it was added for — the macOS package jobs now take the non-empty branch:

```
~~~ Holding binary artifacts out of the cache
Held 1 binary artifact(s) aside; they stay on disk at /Users/builder/Library/Caches/org.swift.swiftpm/artifacts for the rest of this job.
--- :mag: Verify binary artifacts survived the cache save
Binary artifacts are still in /Users/builder/Library/Caches/org.swift.swiftpm/artifacts.
```

The artifact was downloaded fresh (`Downloading binary artifact … [218776/218776]`) while the source dependency in the same step came from cache, so both paths are visible in one log.

## Known limitation, pre-existing

Both package jobs restored an existing cache entry and hit `This file is already cached – skipping upload`. So the assertion passed against a `save_cache` that returned early: the hold-aside and restore ran for real, but the **tar-and-upload leg is still only covered by the bats suite against stubs**.

That predates this PR — the fixtures' `Package.resolved` rarely changes, so these jobs have always restored and skipped. Making the key vary per run would exercise it, at the cost of a new S3 cache entry on every build. Not obviously worth it; noting it so the gap is on record rather than assumed covered.

## One thing this surfaced that matters beyond the fixture

Remote binary targets are pinned by checksum in `Package.swift` and never appear in `Package.resolved` — adding one leaves the resolved file byte-identical, which is why the fixture's cache key did not move.

The cache key is a hash of `Package.resolved`, so it tracks binary targets **only when they arrive transitively from a pinned dependency**. That holds for Day One (it declares none of its own), but a repo declaring a binary target directly can change that URL or checksum without moving the key. Worth knowing before anyone caches artifacts across jobs.

---

*Posted by Claude Code (Opus 5) on behalf of @mokagio with approval.*

bin/install_swiftpm_dependencies
```diff
@@ -121,15 +141,18 @@ elif [[ "${USE_SPM}" == "true" ]]; then
...
+# keeps an archive that fails validation and replays it on every later resolve
+# (swiftlang/swift-package-manager#10164 fixes this on `main` only, so no shipping Xcode has it).
+echo "~~~ Holding binary artifacts out of the cache"
+if [[ -d "${ARTIFACTS_LOCATION}" ]]; then
+  trap return_artifacts_to_cache EXIT
+  mv "${ARTIFACTS_LOCATION}" "${ARTIFACTS_HELD_ASIDE}"
+  echo "Held $(find "${ARTIFACTS_HELD_ASIDE}" -type f | wc -l | tr -d ' ') binary artifact(s) aside; they stay on disk at ${ARTIFACTS_LOCATION} for the rest of this job."
+else
+  echo "No binary artifacts to hold aside."
+fi
```
AliSoftware: Wouldn't it be way simpler to instead make `save_cache` accept an optional `--exclude [PATTERN]` flag (just like it already supports the `--use_relative_path_in_tar` flag), and forward that to the `tar` invocation that `save_cache` uses internally?

This way not only would we not need to do this "held-aside" dance just for that, but this would also make `save_cache` more versatile/flexible, as it would allow us to use that new `--exclude [PATTERN]` flag in any other place we use `save_cache`, not just for this particular cases in `install_swiftpm_dependencies` (and I'm sure there are other use cases for which excluding a subdir from directory we want to cache would be useful, maybe even for other caching helpers that depend on `save_cache` like `install_gems` or `install_npm_packages` and the like?)

---

As an aside, that would be the perfect occasion to improve the way our `save_cache` implementation parses arguments and flags, as currently it's doing it in a non-conventional way, reading the `SHOULD_FORCE` parameter from `$3` and not as a flag, and reading `--use_relative_path_in_tar` as `$4` (instead of accepting it at any position like POSIX flags usually are). If we migrate `save_cache` to read the flags using the typical `while` + `case` loop that we use for other `a8c-ci-toolkit` helpers this will make the whole implementation more readable, the helper more flexible and consistent with other helpers in this repo, and make it easier to add support for more flags to this command in the future.

dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Warning
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:warning:</td>
      <td data-sticky="false">This PR is larger than 500 lines of changes. Please consider splitting it into smaller PRs for easier and faster reviews.</td>
    </tr>
  </tbody>
</table>
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

CHANGELOG.md
```diff
@@ -38,15 +38,17 @@ _None._
...
 ### Bug Fixes
 
-_None._
+- `install_swiftpm_dependencies` keeps downloaded binary artifacts on disk, so a later resolution in the same job reuses them. They stay out of the uploaded cache. [#233]
 
 ### Internal Changes
 
 - `setup_azure_trusted_signing` documents why `AZURE_TIMESTAMP_SERVER` needs the `http://` scheme. [#232]
+- Add assertion tests for `save_cache` and `install_swiftpm_dependencies` covering what reaches the uploaded cache. [#233]
+- Give the `install_swiftpm_dependencies` integration fixtures a binary target. [#233]
```
mokagio:
```suggestion
```

PR #235, opened 2026-08-27
Remove `git-conceal-unlock` obsolete command

AliSoftware:
As the title says. `git-conceal` has been retired and its repo archived, and we're in the process of migrating to `a8c-secrets` anyway (for which this plugin already expose a `bin/install_a8c-secrets_binary` command, btw).

No repository has ever used this `git-conceal-unlock` action (it was only ever used in PRs that ended up being closed without ever being merged, once we decided not to go with this `git-conceal` solution)

---


CHANGELOG.md
```diff
@@ -14,7 +14,7 @@ When releasing a new version:
 
 ### Breaking Changes
 
-_None._
+- Remove `git-conceal-unlock` helper, given the tool was never used and is now superseded by `a8c-secrets`. [#235]
```
mokagio, line 17: > No repository has ever used this git-conceal-unlock action (it was only ever used in PRs that ended up being closed without ever being merged, once we decided not to go with this git-conceal solution)

Should we do ourselves a favor, then, and not track this as a breaking change?

How does the puzzle go? If a tree falls in the forest but nobody is there, does it really make a sound? If no one ever used this code, is removing it really a breaking change?

Sure, it would be a bit of a deviation from proper SemVer, but, even when AI does it for us, considering whether to adopt a new major version always requires comparing the changelog and migration guide against the repo's setup, so it would be practically useful.

I'm fine either way. The adoption cost of bumping to 7.0.0 is non zero but still little. Just thought I'd ask the question.

mokagio (approved):
+1 on the removal. 

Approving with disabled auto-merge to get your input on possibly not considering this a breaking change. As I said, I'm fine with whichever option.

CHANGELOG.md
AliSoftware: Yeah to be fair I asked myself the exact same question when I wrote that changelog entry here, with the exact same reasoning as yours (I should have mentioned it in the PR description).

And ultimately I decided to go with breaking change because:
1. the cost of doing a major bump is not _that_ big
2. We don't use `~> x.y`-like constraints for a8c-ci-toolkit, so it's not like doing a major version bump instead of a minor one will prevent automatic updates or dependabot from adopting the new version automatically: for the case of `a8c-ci-toolkit`, adopting a new version of it in any repo requires a manual PR from us updating the hardcoded version value in `shared-pipeline-vars` (with us being the ones deciding which version to set it to), not some `Gemfile` or some other dependency-manager-related file that would treat a major version bump as not eligible for some `~> x.y` declaration
3. None of the other sections of our CHANGELOG felt right anyway as an alternative: this is clearly not really a "New Feature" nor a "Bug Fix". Maybe we could consider it an "Internal Change", but even then that felt kinda odd.

So in the end I concluded that "eh, let's list it under Breaking Change as that's the section that makes the most sense for this change anyway, and because even if it means it'll lead to a major version bump it's not that big of a deal after all" 🙃

AliSoftware: PS: We could also always decide to do a minor version bump next time we do a release, _even if_ this entry is listed under "Breaking Change" in the CHANGELOG 🙃
We're the one choosing which version number to tag with after all. That would look a bit inconsistent with our usual process, but that's still an option if we want to.

PR #236, opened 2026-08-31
Delete CocoaPods-related commands from the plugin

mokagio:
Part of [AINFRA-3027](https://linear.app/a8c/issue/AINFRA-3027)

We no longer use CocoaPods, so we can stop carrying along this dead weight.

I noticed these while looking at the code for https://github.com/Automattic/a8c-ci-toolkit-buildkite-plugin/pull/233 and catching a `pod` on my machine...

I think this would be safe to ship. The only repo I can find that still uses CocoaPods is https://github.com/wordpress-mobile/AztecEditor-iOS. The reason it does is because of `gutenberg-mobile`, but that project has been archived so it might be safe to look at removing CocoaPods from it, too. Regardless, this project is versioned so if CP needs to stick around for a few more months in the repo, we can simply not bump the CI toolkit version there.

---


dangermattic:
<table>
  <thead>
    <tr>
      <th width="50"></th>
          1 Message
      </th>
     </tr>
  </thead>
  <tbody>
    <tr>
      <td>:book:</td>
    </tr>
  </tbody>
</table>

</p>

MIGRATION.md
```diff
@@ -1,5 +1,14 @@
 # Migration Instructions for Major Releases
 
+## From 6.0.0 to 7.0.0
+
+* The `git-conceal-unlock` helper has been removed.
+  Use [`a8c-secrets`](https://github.com/Automattic/a8c-secrets) instead, installing it with `install_a8c-secrets_binary`.
+* All the CocoaPods commands have been removed: `build_and_test_pod`, `cache_cocoapods`, `cache_cocoapods_specs_repos`, `install_cocoapods`, `lint_pod`, `patch-cocoapods`, `publish_pod`, `publish_private_pod`, `slack_notify_pod_published`, `validate_podfile_lock`, and `validate_podspec`.
+  This plugin offers no replacement for them, so pipelines that still build or publish with CocoaPods need to stay pinned to a `6.x` version.
```
mokagio:
```suggestion
* CocoaPods is no longer supported. All the CocoaPods commands have been removed:
  * `build_and_test_pod`
  * `cache_cocoapods`
  * `cache_cocoapods_specs_repos`
  * `install_cocoapods`
  * `lint_pod`
  * `patch-cocoapods`
  * `publish_pod`
  * `publish_private_pod`
  * `slack_notify_pod_published`
  * `validate_podfile_lock`
  * `validate_podspec`
```

MIGRATION.md
```diff
@@ -1,5 +1,14 @@
...
+* `lint_localized_strings_format` no longer runs `install_cocoapods` when it finds a `Podfile.lock`.
+  Install the pods before calling it if the strings generation depends on them.
```
mokagio:
```suggestion
* As a result of the above, `lint_localized_strings_format` no longer runs `install_cocoapods` when it finds a `Podfile.lock`.
```

AliSoftware (approved):
Great minds, I thought about it the other day when I opened my PR to remove `git-conceal-unlock` and saw those Cocoapods commands too, but forgot to follow-up 🙃  

So thanks for cleaning those up! 🧹 🧹 🧹 ✨ 

The only other repo I could think of that might still be using CocoaPods is `Matticspace-Mobile` (see https://github.com/Automattic/Matticspace-Mobile/pull/233#discussion_r3598606599 and https://github.com/Automattic/Matticspace-Mobile/pull/234). But that's just an artifact of Kotlin Multiplatform Mobile (KMM) relying on it _internally_ for the Sentry pod (`cocoapods {…}` block in `app/build.gradle.kts` which generates the `Podfile` on the fly, etc), and I thus don't think the repo uses any of those Cocoapods-related `a8c-ci-toolkit` helper directly.

